PR #1, opened 2019-04-30
Set up CI with Azure Pipelines

TimHess:


src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
```diff
@@ -246,10 +246,12 @@ internal ConfigEnvironment DoLoad(bool updateDictionary = true)
                     else
                     {
                         // Single, server make Config Server URI from settings
+#pragma warning disable CS0618 // Type or member is obsolete
```
TimHess: @dtillman could you please take a look at this section? It wasn't immediately obvious to me how to update this section to use non-obsolete methods

TimHess:
@dtillman @hananiel some things to note:

- Several tests are currently marked with `(Skip = "Fails on hosted agent")`
- netcoreapp2.0 has been removed to keep build times under one hour
- test projects that only work on Windows have had folders renamed from `*.Test` to `*.Net4Test`
- The OSX job currently doesn't run tests due to around 2-6 test failures in HystrixBase on every run
- All project files are a bit slimmer now via sharedproject and sharedtest props files
- There is now only one `versions.props` file
- .sln files should all have updated references to the files shared by csproj .files
- Build _warnings_ will break the build
- The OpenCensus projects now comply with Stylecop... which means they have copyright notices

PR #55, opened 2019-06-05
Merge serilog pr

hananiel:


hananiel:
@TimHess almost all of the code we used is dependent on Serilog.Core or DynamicLogger - so ended up renaming to DynamicSerilogCore and adding some tests for coverage. If there is appetite to support for framework I don't see any code moving out of here - since our common abstractions are already either in the serilog or dynamic logger assemblies.

PR #57, opened 2019-06-11
Microsoft HealthChecks: Remove 3rd Party Deps, add tests

hananiel:
Prevents dependency on AspnetCore.HealthChecks.*  packages, while still allowing integration for health endpoint to work with community health checks.

### DIFF
--- a/src/Connectors/src/ConnectorBase/IConnectorInfo.cs
+++ b/src/Connectors/src/ConnectorBase/IConnectorInfo.cs
@@ -1,0 +1,32 @@
+// Copyright 2017 the original author or authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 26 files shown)


src/Connectors/src/ConnectorBase/IConnectorInfo.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Reflection;
+
+namespace Steeltoe.CloudFoundry.Connector
+{
+    public interface IConnectorInfo
```
TimHess: It looks like all the implementations of this interface are named *ConnectionInfo, maybe we should stick with that name

src/Connectors/src/ConnectorCore/PostgresProviderServiceCollectionExtensions.cs
```diff
@@ -84,25 +86,24 @@ public static IServiceCollection AddPostgresConnection(this IServiceCollection s
...
+            if (!services.Any(s => s.ServiceType == typeof(HealthCheckService)) || addSteeltoeHealthChecks)
             {
                 services.Add(new ServiceDescriptor(typeof(IHealthContributor), ctx => new RelationalHealthContributor((IDbConnection)factory.Create(ctx), ctx.GetService<ILogger<RelationalHealthContributor>>()), ServiceLifetime.Singleton));
             }
-            else
-            {
-                healthChecksBuilder.AddNpgSql(factory.CreateConnectionString());
-            }
+
+            var foo = new PostgresProviderConfigurer();
```
TimHess: Are these two lines supposed to be here still?

Issue #70, opened 2019-06-26
Missing `.ConfigureAwait(false)` in all await call


pengweiqhca:
Missing `.ConfigureAwait(false)` could be deadlocks  in asp.net when call async method in sync method.

https://blog.stephencleary.com/2012/07/dont-block-on-async-code.html
https://blogs.msdn.microsoft.com/alazarev/2017/05/20/fun-with-configureawait-and-deadlocks-in-asp-net/
https://medium.com/rubrikkgroup/understanding-async-avoiding-deadlocks-e41f8f2c6f5d
https://medium.com/bynder-tech/c-why-you-should-use-configureawait-false-in-your-library-code-d7837dce3d7f

TimHess:
I think we got all them all, thanks for your contribution!

PR #71, opened 2019-06-27
Add the missing `.ConfigureAwait(false)`

pengweiqhca:
https://github.com/SteeltoeOSS/steeltoe/issues/70


TimHess:
Thanks for opening this PR - I changed the base because we already had some work in progress that would include a few of these changes. Some of the changes (to Hystrix in particular) need to be reviewed relatively in depth, so we won't be straight up accepting it right away and it might take a few days to work through the changes.

PR #80, opened 2019-07-09
Override default path on external actuator context path if it points to /cloudfoundryapplication

### LINKED ISSUE #74: External facing actuators should not accept a setting of "/CloudFoundryApplication"
This is the typical default from previous examples and the one path that will not work. When running in cloudfoundry and endpoint:management:path is set to "/CloudfoundryApplication" should revert to /actuators and print a warning message that this has been changed.

hananiel:
Resolves #74


src/Management/test/EndpointBase.Test/ManagementEndpointOptionsTest.cs
```diff
@@ -0,0 +1,61 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Collections.Generic;
+using Xunit;
+
+#pragma warning disable CS0612 // Type or member is obsolete
```
TimHess: marking this class with `[Obsolete]` should let the tests keep running and prevent the warning from showing

Issue #45, opened 2019-05-30
Oracle DB Connector Support


clklachu:
Is there a support for the health check of the Oracle DB? I see there are some implementation for OracleServiceInfo but there are no OracleProviderConnectorFactory and OracleProviderConfigurer implementations. Any specific reason for it? If not, can we go ahead and contribute on it?

TimHess:
We currently have nothing beyond the Service Info for Oracle DB, largely because we haven't seen much demand for it. If you'd like to add support, we'd love to see it contributed!

kishoreallwynraj:
@TimHess you can assign this issue to my name. @clklachu is a colleague of mine. I will discuss with him and we will contribute the Oracle connector support.

TimHess:
Github won't let me directly assign the issue to you, but let's all consider that to be the arrangement

kishoreallwynraj:
@TimHess We have submited the PR to add Oracle connectors, health checks and associated test cases. Please review and let us know your feedback.
https://github.com/SteeltoeOSS/steeltoe/pull/84

TimHess:
Thanks for your contribution!

PR #84, opened 2019-07-16
Oracle DB Connector

kishoreallwynraj:
Adding connectors, health checks and test cases to support Oracle DB.

Addresses https://github.com/SteeltoeOSS/steeltoe/issues/45

### DIFF
--- a/src/Connectors/src/ConnectorBase/Relational/Oracle/OracleProviderConnectorOptions.cs
+++ b/src/Connectors/src/ConnectorBase/Relational/Oracle/OracleProviderConnectorOptions.cs
@@ -1,4 +1,4 @@
-﻿// Copyright 2019 Infosys Ltd.
+﻿// Copyright 2017 the original author or authors.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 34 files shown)


src/Connectors/src/ConnectorBase/Relational/Oracle/OracleProviderConnectorOptions.cs
```diff
@@ -0,0 +1,90 @@
...
+            var section = config.GetSection(ORACLE_CLIENT_SECTION_PREFIX);
+
+            section.Bind(this);
+
+            if (Uri != null)
+            {
+                Server = Uri.Split(':')[2].Substring(2);
+            }
+
+            Username = Uid;
```
TimHess: Is there a reason for including `Uid` and `Pw` and mapping them over to `Username` and `Password`?

For reference, this behavior is included with the SqlServer connector to account for the way the Sql Server Service Broker provides credentials. Is there a broker for Oracle?

TimHess:
Hi @kishoreallwynraj and @clklachu,

Thanks for your contribution! If you could please sign the CLA and answer about the Uid/Pwd fields and service broker we'll be just about ready to accept this

src/Connectors/src/ConnectorBase/Relational/Oracle/OracleProviderConnectorOptions.cs
kishoreallwynraj: I think those properties were not removed when we reused the code from existing options class from other DB classes. I commented them and still able to connect to the Oracle DB. I will rerun all test cases and push the code shortly.

kishoreallwynraj: Pushed the fixes for the review comment

PR #87, opened 2019-07-23
Add support for DbMigrations endpoint

macsux:


.gitignore
```diff
@@ -1,3 +1,29 @@
+# Common IntelliJ Platform excludes
```
TimHess: Can/should all of these be covered with `**/.idea/*` instead of listing them out? Are there some files in that folder we might want to include?

macsux: Yes, Rider has the ability to create shared build profiles and workspace settings which can be promoted to "team" level. Build profiles are similar to launchsettings but more powerful. I may choose to use them in future contributions which people using rider will benefit from.

TimHess:
Thanks @macsux!

PR #93, opened 2019-08-09
recognize vcap setting p.config-server for scs3.0

### LINKED ISSUE #92: SCS 3.0 config server is not working with Steeltoe libs
Error when binding app to SCS 3.x config server:
`2019-08-02T09:39:31.688-05:00` [CELL/0] [OUT] Starting health monitoring of container

2019-08-02T09:39:33.807-05:00 [APP/PROC/WEB/0] [OUT] info: Steeltoe.Extensions.Configuration.ConfigServer.ConfigServerConfigurationProvider[0]

2019-08-02T09:39:33.807-05:00 [APP/PROC/WEB/0] [OUT] Fetching config from server at: http://localhost:8888

2019-08-02T09:39:36.515-05:00 [APP/PROC/WEB/0] [OUT] fail: Steeltoe.Extensions.Configuration.ConfigServer.ConfigServerConfigurationProvider[0]

2019-08-02T09:39:36.515-05:00 [APP/PROC/WEB/0] [OUT] Config Server exception: System.Net.Http.HttpRequestException: An error occurred while sending the request. ---> System.Net.WebException: Unable to connect to the remote server ---> System.Net.Sockets.SocketException: No connection could be made because the target machine actively refused it 127.0.0.1:8888

2019-08-02T09:39:36.515-05:00 [APP/PROC/WEB/0] [OUT] at System.Net.Sockets.Socket.InternalEndConnect(IAsyncResult asyncResult)

2019-08-02T09:39:36.515-05:00 [APP/PROC/WEB/0] [OUT] at System.Net.Sockets.Socket.EndConnect(IAsyncResult asyncResult)
`

hananiel:
Resolves #92

### DIFF
--- a/src/Configuration/Configuration.sln.DotSettings
+++ b/src/Configuration/Configuration.sln.DotSettings
@@ -1,0 +1,3 @@
+﻿<wpf:ResourceDictionary xml:space="preserve" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:s="clr-namespace:System;assembly=mscorlib" xmlns:ss="urn:shemas-jetbrains-com:settings-storage-xaml" xmlns:wpf="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=CONFIGSERVER/@EntryIndexedValue">True</s:Boolean>
+	<s:Boolean x:Key="/Default/UserDictionary/Words/=VCAP/@EntryIndexedValue">True</s:Boolean></wpf:ResourceDictionary>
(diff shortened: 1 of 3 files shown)


cdelashmutt-pivotal:
Does the test actually cover the SCS 3.0 case?  In the test code at https://github.com/SteeltoeOSS/steeltoe/pull/93/commits/b12e1658ac6b3dfa6cf40684e911d3838794f56e#diff-1088cbd86c30ab8e1719d4de5713d91bR566, I see the section name is `p-config-server`.  Shouldn't that be `p.config-server` for SCS 3.0?

hananiel:
@TimHess I looked into this changing this to stream instead of temp file, but there is no way to a jsonString from memory - we could write a new extension but this `internal class JsonConfigurationFileParser` would need to be replicated. Alternatively, we could translate the json into a Dictionary, but that would not be the same thing ...

PR #94, opened 2019-08-09
Initial netcoreapp3.0 Compatibility

TimHess:
Initial draft for working with ANC3/netcoreapp3.0, starting with dynamic logging


src/Logging/src/DynamicLogger/DynamicILogger.cs
```diff
@@ -13,47 +13,30 @@
...
-using Microsoft.Extensions.Logging.Console;
-using Microsoft.Extensions.Logging.Console.Internal;
 using System;
 using System.Collections.Generic;
-using System.Runtime.CompilerServices;
 
 namespace Steeltoe.Extensions.Logging
 {
-    public class DynamicConsoleLogger : ILogger
+    public class DynamicILogger : ILogger
```
hananiel: Howabout DynamicLogger instead of DynamicILogger?

TimHess: Just putting it back the way it was

TimHess:
@anurse @rynowak @brennanconroy this is a draft PR that updates Steeltoe's Dynamic Logging to work with `Microsoft.Extensions.Logging` 3.0.  I'm particularly interested in your insights around an alternative to self-managing the logging filters as this implementation is doing, but please feel free to point out any other issues you see, thanks!

src/Logging/src/DynamicLogger/DynamicConsoleLogger.cs
```diff
@@ -13,46 +13,30 @@
...
-        {
-            get => _delegate.Filter;
-
-            set => _delegate.Filter = value;
-        }
+        public ILogger Delegate { get; set; }
 
-        [Obsolete("Changing this property has no effect. Use " + nameof(ConsoleLoggerOptions) + "." + nameof(ConsoleLoggerOptions.IncludeScopes) + " instead")]
-        public bool IncludeScopes { get; set; }
+        public Func<string, LogLevel, bool> Filter { get; set; }
```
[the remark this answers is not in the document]
TimHess: `internal` set might work, but `DynamicConsoleLoggerProvider` is currently [directly updating this filter](https://github.com/SteeltoeOSS/steeltoe/blob/86351b78f31dd97b26c376c7e08f76433107e00a/src/Logging/src/DynamicLogger/DynamicConsoleLoggerProvider.cs#L154)

src/Logging/src/DynamicLogger/DynamicConsoleLoggerProvider.cs
```diff
@@ -23,66 +23,43 @@
...
         /// </summary>
         /// <param name="options">Pass-through to ConsoleLoggerProvider constructor</param>
         /// <param name="filterOptions">Logger filters</param>
         /// <param name="messageProcessors">message processors to apply to message</param>
-        public DynamicLoggerProvider(IOptionsMonitor<ConsoleLoggerOptions> options, IOptionsMonitor<LoggerFilterOptions> filterOptions, IEnumerable<IDynamicMessageProcessor> messageProcessors = null)
+        public DynamicConsoleLoggerProvider(IOptionsMonitor<ConsoleLoggerOptions> options, IOptionsMonitor<LoggerFilterOptions> filterOptions, IEnumerable<IDynamicMessageProcessor> messageProcessors = null)
         {
             _filterOptions = filterOptions;
-            SetFiltersFromOptions();
             _delegate = new ConsoleLoggerProvider(options);
```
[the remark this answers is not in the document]
TimHess: They could, and yes we'd probably want to avoid that. Is there a straightforward way to avoid that situation?

src/Logging/src/DynamicLogger/DynamicConsoleLoggerProvider.cs
```diff
@@ -194,23 +171,23 @@ protected virtual void Dispose(bool disposing)
...
-                    _settings = null;
                     _loggers = null;
                 }
 
                 disposed = true;
             }
         }
 
-        ~DynamicLoggerProvider()
+        ~DynamicConsoleLoggerProvider()
```
rynowak: It's a little strange to see a finalizer here. You don't have any unmanaged resources

src/Logging/src/DynamicLogger/DynamicConsoleLoggerProvider.cs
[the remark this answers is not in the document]
rynowak: It might make sense to abandon the idea of wrapping the logger altogether and *just* register a filter. https://github.com/aspnet/Extensions/blob/d656c4f7e22d1c0b84cab1b453c50ce73c89a071/src/Logging/Logging/src/FilterLoggingBuilderExtensions.cs#L20

You could keep something like this class around as a DI service to manage configuration, and then just register a filter around the console logging provider to do your work. Then you can always avoid duplication.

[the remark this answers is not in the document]
TimHess: @rynowak I like that idea, and I was hoping for something simpler than what we've got, but there's a detail that I tend to forget about: in the Dynamic Console Logger, we also provide a hook for providing additional message processing before passing the log entry to the underlying logger. The main (only?) use case we have right now is for enabling log correlation with OpenCensus -- https://github.com/SteeltoeOSS/steeltoe/blob/master/src/Management/src/TracingBase/TracingLogProcessor.cs we'd need an alternative mechanism for attaching metadata for basically all log entries to move in that direction

src/Management/src/EndpointCore/Mappings/MappingsEndpointMiddleware.cs
```diff
@@ -266,6 +270,9 @@ private ApiDescriptionProviderContext GetApiDescriptions(IReadOnlyList<ActionDes
                 return new ApiDescriptionProviderContext(new List<ActionDescriptor>());
             }
 
+#if NETCOREAPP3_0
+            throw new NotImplementedException("ApiDescriptionActionData is `internal` in AspNetCore 3");
```
TimHess: @rynowak this is another spot that breaks for us in 3.0. I'm sure we'll get to something that works, but if you have a suggested alternative it'd be nice to save some research time

rynowak: What's the purpose of this method? What is this trying to do?

rynowak: If you're just trying to access the API Descriptions - use `IApiDescriptionCollectionProvider`.  This looks really scary to me though, it looks like you're duplicating functionality.

TimHess: This method exists to return the MVC and WebAPI Routes and Route templates used by the application, specifically for compatibility with a feature in Apps Manager on PCF

image

rynowak: You should be able to get this data from `IApiDescriptionCollectionProvider`.

TimHess: Good to know, I'll have to try that out to see if there's a reason we have it in the way it is now instead...

PR #100, opened 2019-08-15
Support collecting core dump on Linux

cshung:
This is my attempt to add crash dump collection support on Linux. It appears to be a big code change, but it isn't, because most of the files are just copy-and-paste from the [dotnet/diagnostics](https://github.com/dotnet/diagnostics) repo.

The key file that is original is LinuxHeapDumper, using just one line of code, it calls into the code I copied to send a message to the runtime through the DiagnosticsIPC channel. After receiving the message, the runtime will collect a crash dump at that point in time. The API is blocking so that it does not return before the API returns.

The pull request is a work in progress, I notice there are quite a few more things we need to work on:

**Dependencies related:**
- [x] Should we copy the code in or should we consume the dependencies as binary?
- [ ] The API defined on that binary is not finalized yet, have a plan to address things might change in the future? What kind of changes do we like to see?
- [x] Check out the [licensing ](https://github.com/dotnet/diagnostics/blob/master/LICENSE.TXT)to make sure it is okay to consume the library. 

**Coding related**
- [x] Make sure the code is robust with respect to errors, and
- [ ] Make sure we map the options so that it collects the right type of crash dump.
- [x] Conditioning enable the capability - it is only for Linux and .NET Core 3.0 or above.

**Testing related**
- [x] Try it out in an integrated scenario to make sure it works as expected, for example
- [x] Does the collected dump contains the right stacks, and
- [ ] Does the collected dump contains an analyzable managed heap.


jkonicki:
Thank you @cshung! This is a great contribution.  We will have the team review this PR and get back with any comments.

TimHess:
Hi @cshung,
Thanks again for the contribution, I plan to take a closer look tomorrow with regard to conditionally calling the code and trying it out, but I wanted to confirm that the licensing looks fine. Also, copying the code makes the most sense for now - after the API firms up and ships we will switch to taking a dependency. I don't have any comments with regard to changes at this point

src/Management/src/Diagnostics/EventPipe/IpcClient.cs
```diff
@@ -0,0 +1,115 @@
...
+                string ipcPort = Directory.GetFiles(IpcRootPath) // Try best match.
+                    .Select(namedPipe => (new FileInfo(namedPipe)).Name)
+                    .SingleOrDefault(input => Regex.IsMatch(input, $"^dotnet-diagnostic-{processId}-(\\d+)-socket$"));
+                if (ipcPort == null)
+                {
+                    throw new PlatformNotSupportedException($"Process {processId} not running compatible .NET Core runtime");
+                }
+                string path = Path.Combine(Path.GetTempPath(), ipcPort);
+                // TODO: Enable this for appropriate build
+                // var remoteEP = new UnixDomainSocketEndPoint(path);
```
cshung: With the version bumped up, can we try and see if this compile?
This line is critical to access the pipe.

TimHess: Yes, please go ahead and change anything you think needs changing here

cshung:
My apology for progress has been really slow here. 

I attempted to fix the error related to `UnixDomainSocketEndPoint`. 

It appears the type is only available in [NET Standard 2.1 preview](https://docs.microsoft.com/en-us/dotnet/api/system.net.sockets.unixdomainsocketendpoint?view=netstandard-2.1), but the project `Steeltoe.Management.Diagnostics` references .NET Standard 2.0. I don't know how it worked with the diagnostics repo. Over [there](https://github.com/dotnet/diagnostics/blob/master/src/Microsoft.Diagnostics.Tools.RuntimeClient/Microsoft.Diagnostics.Tools.RuntimeClient.csproj) the library referenced netcoreapp2.1.

To remedy this, I attempted to copy the source code for UnixDomainSocketEndPoint from [CoreFX](https://github.com/dotnet/corefx/blob/master/src/System.Net.Sockets/src/System/Net/Sockets/UnixDomainSocketEndPoint.cs) and see how it goes. If we have a solution for this we could contribute that back to the diagnostics repo for this [issue](https://github.com/dotnet/diagnostics/issues/307). I have got the Windows version compiles except StyleCop warning, for Unix I knew it is incomplete, something must be done with the Interop call.

dtillman:
Thanks again for this contribution, we will take a closer look at this as we focus in on a Steeltoe 3.0 release.

cshung:
> Thanks again for this contribution, we will take a closer look at this as we focus in on a Steeltoe 3.0 release.

Thank you for your attention

dtillman:
@cshung :  Thanks again for this contrib ..
Starting to look at this in detail... question:
 - With Steeltoe 3.0 we are removing support for full framework(Windows) .. and only supporting .NET Core .. it seems that we might be able to eliminate the two ways of getting a dump (WIndowsHeapDumper & LinuxHeapDumper)  ..  and just have one single way (i.e. HeapDumper) that uses the IPC mechanism for both ... would you agree?

Dave

cshung:
> @cshung : Thanks again for this contrib ..
> Starting to look at this in detail... questions:
> 
> * With Steeltoe 3.0 we are removing support for full framework(Windows) .. and only supporting .NET Core .. it seems that we might be able to eliminate the two ways of getting a dump (WIndowsHeapDumper & LinuxHeapDumper)  ..  and just have one single way (i.e. HeapDumper) that uses the IPC mechanism for both ... possible?
> * Does the "IPC dump" mechanism work for .NET Core on Windows, Linux and OSX?  After looking at the Diagnostics repo it looks like maybe this is not possible?
> * What does this code check for: if (typeof(object).Assembly.GetType("System.Index") != null)
> 
> Dave

That's a great idea, however, it appears to me that it will not work, because the capability for generating a core dump at runtime through RPC is enabled only under `FEATURE_PAL` (see [here](https://github.com/dotnet/coreclr/blob/d8bbcc485be92b2feaf392eb750b7b523a872b6c/src/vm/diagnosticserver.cpp#L79) for more information). If we used the same code path for Windows, it would just fail with `CORDIAGIPC_E_UNKNOWN_COMMAND`

dtillman:
@cshung : 
I tried to run the unit tests under WSL1 .. and it failed to create the dump.  HR=80004005 .. Is this expected? 

Dave

cshung:
> @cshung :
> I tried to run the unit tests under WSL1 .. and it failed to create the dump. HR=80004005 ... Is this expected?
> 
> Dave

Nope, this is not expected. Can you share with me more info about the failure? If I can repro what is going on, I can debug and take a look at what is wrong.

Fundamentally, the code uses the same mechanism `dotnet dump` does, can you please try running `dotnet dump` on your setup and see what is going on. The full instruction on running `dotnet dump` is available [here](https://github.com/dotnet/diagnostics/blob/master/documentation/dotnet-dump-instructions.md).

Earlier I was fixing the unit test failures on CI, and most of them have to do with unsupported scenarios. For example, it isn't running on .NET Core 3.0, or it is not Linux. It might be one of these cases.

dtillman:
@cshung 
The unit test that failed was :  public void Invoke_CreatesDump().

The filename didn't exist after calling DumpHeap().

I checked the HR returned from : 
DiagnosticsHelpers.GenerateCoreDump(Process.GetCurrentProcess().Id, fileName, DiagnosticsHelpers.DumpType.Full, false);  in DumpHeap() 

and found 0x80004005 as the returned value.
I'll do some more digging .. try the dotnet dump ... just wanted to verify it should all work on WSL.

Dave

cshung:
You could be hitting this issue: https://github.com/dotnet/diagnostics/issues/548

dtillman:
@cshung:

Ok, I updated to WSL2 and found the tests all pass fine .. so I was likely hitting the issue on WSL1 you mentioned above.

I also verified I could open the resulting dump file using "dotnet-dump analyze .." so all is looking good.

You had a couple TODO's in the LinuxHeapDumper .. if you would finish those up .. I'll merge this in..
Please add some logging of the HResult if it returns !S_OK.

Thanks again for the contrib!
Dave

cshung:
> @cshung:
> 
> Ok, I updated to WSL2 and found the tests all pass fine .. so I was likely hitting the issue on WSL1 you mentioned above.
> 
> I also verified I could open the resulting dump file using "dotnet-dump analyze .." so all is looking good.
> 
> You had a couple TODO's in the LinuxHeapDumper .. if you would finish those up .. I'll merge this in..
> Please add some logging of the HResult if it returns !S_OK.
> 
> Thanks again for the contrib!
> Dave

@dtillman 
I am not sure what's wrong with the Code Analysis task. That feels like a configuration issue, not related to my change.

The only remaining part seems to be `HeapDumpOptions`. It looks like `IHeapDumpOptions` is empty, I am not sure how do I tell what kind of dump the caller is looking for.

PR #124, opened 2019-09-25
Endpoints content headers

### LINKED ISSUE #120: Handle Content Negotiation on endpoints correctly
The spring actuator endpoints response depends on the request headers passed in.For example: 
`curl -v -H "`**Accept: application/vnd.spring-boot.actuator.v2+json, application/vnd.spring-boot.actuator.v1+json, application/json**`" http://localhost:8083/actuator`

Responds with: 
`< HTTP/1.1 200 OK
< Content-Type: `**application/vnd.spring-boot.actuator.v2+json;charset=UTF-8**

Whereas:
`curl -v -H "`**Accept: application/json" http://localhost:8083/actuator**
responds with: 

`< Content-Type: `**application/json;charset=UTF-8**

Even though, the schema for the json is the same. 

For Steeltoe endpoints, when the schema is different we respond with the highest level of implementation available based on configured media type. 

This breaks apps like codecentric/spring-boot-admin which perform strict checking of headers and might break future versions of apps manager. 

The middleware should set the content headers based on accept headers on request. 

A work around is to disable strict checking in other applications.

hananiel:
Adds correct headers for spring-boot compatible clients like spring-boot-admin. Resolves #120


src/Management/src/EndpointBase/ActuatorMediaTypes.cs
```diff
@@ -0,0 +1,60 @@
...
+                {
+                    if (acceptHeader == ANY)
+                    {
+                        contentHeader = AllowedAcceptHeaders(version).First();
+                    }
+                    else
+                    {
+                        contentHeader = AllowedAcceptHeaders(version)
+                            .FirstOrDefault(header => acceptHeader == header);
+                    }
```
Karql: Imo when contentHeader is found foreach should be brake

src/Management/src/EndpointCore/ContentNegotiation/ContentNegotiationExtensions.cs
```diff
@@ -0,0 +1,71 @@
...
+                }
+            }
+        }
+
+        public static void SetContentType(this IHeaderDictionary responseHeaders, IHeaderDictionary requestHeaders, ILogger logger, MediaTypeVersion version = MediaTypeVersion.V2)
+        {
+            var acceptMediaTypes = new List<string>();
+            if (requestHeaders.TryGetValue("Accept", out var acceptHeader))
+            {
+                acceptMediaTypes = acceptHeader.ToString().Split(';').ToList();
```
Karql: spring boot send headers delimeted by ','
"application/vnd.spring-boot.actuator.v2+json,application/vnd.spring-boot.actuator.v1+json,application/json"

PR #160, opened 2019-11-07
Prevent placeholderResolver from removing other propertysources

hananiel:


TimHess:
I was going to say this should go into 2.x, but just changing the base on the PR doesn't look like it would go well as-is

jkonicki:
Yes, this was supposed to go into 2.4.0.

PR #165, opened 2019-11-08
add more (Web)HostBuilder extensions for parity between the two

TimHess:
Adds extensions for both IHostBuilder and IWebHostBuilder for feature parity between the two

PR #168, opened 2019-11-11
Define CORS before activation, allow customization

TimHess:
Allows custom CORS definitions. 
Note: Using custom CORS prevents application of `.AllowAnyOrigin()` as that's the most likely piece that needs customization. 

Also changes the setup to define the CORS policy in ConfigureServices and activate during Configure, so that the policy can be retrieved by name later, specifically for testing. This change _could_ cause an issue for anybody that uses `.UseCloudFoundryActuators` but not `AddCloudFoundryActuators` as the CORS policy won't exist, but that is not something that is expected

This PR includes commits from #165 as this work affects some additions from that PR...


TimHess:
I'm going to re-do this to avoid commit weirdness with how the other PR merged

Issue #176, opened 2019-11-16
EurekaBase potential for NullReferenceException when no instance id exists


putridparrot:
If the applications XML contains an application without an instanceId element, Steeltoe correctly has an instanceId of null, however this causes a problem as the Applications.cs file (line 147) 

instances[info.InstanceId] = info;

this code will attempt to add a key to the ConcurrenctDictionary (instances) which is null. This in turn causes an exception. The result being that this exception causes other application objects with valid instanceId's to not get loaded/registered with Steeltoe and hence trying to get them results in a null return value.

I raised this issue previous (https://github.com/SteeltoeOSS/steeltoe/issues/13) and it's not related to the fix highlighted within that thread.

I have a fork with a commit and test for this, will raise a PR if you wish to accept or it will highlight the issue hopefully.

putridparrot:
Raised PR, feel free to ignore/reject it's more to demonstrate the problem https://github.com/SteeltoeOSS/steeltoe/pull/177

PR #177, opened 2019-11-16
EurekaBase potential for NullReferenceException when no instance id exists

putridparrot:
Test for issue https://github.com/SteeltoeOSS/steeltoe/issues/176 as well as a very basic fix, would be good to look to log such issues. Also fixed typo. whilst in this code.


TimHess:
This looks like a reasonable thing to check for, thanks for following up!

Issue #184, opened 2019-11-27
Configprovider : configuration changes are not reflecting when we delete any setting/feature toggle


vivekkjain:
When Steeltoe ConfigServer reads/Load config settings that get stored in the Dictionary of ConfigurationProvider Data (Base abstract class). Now if we change anything any config setting and trigger Reload/Load it will maintain the old data and only add/update the settings in the Dictionary.

## Step to Reproduce
1) Initial Configuration Value:  setting{ "1", "2","3"}
      Data Dictionary will have. below: 
              KEY        | Value
             setting:0 | "1"
             setting:1 | "2"
             setting:2 | "3"
2) Change the configurations like this: setting{ "none"} and reload the configuration while calling config.Reload().
       Now Data Dictionary will have. below: 
              KEY        | Value
             setting:0 | "none"
             setting:1 | "2"
             setting:2 | "3"

## Expected behavior
Data dictionary should have these value on step 2 above:
               KEY        | Value
             setting:0 | "none"

## Environment (please complete the following information):
 - Platform: Cloud Foundry
 - OS: Linux
- ..NET Core 2.2.0 
 - Steeltoe.Extensions.Configuration.ConfigServerBase, Version=2.1.1.0 
 - Any other library versions to note

dtillman:
@vivekkjain : Do you have a sample app we can run to reproduce the problem?

vivekkjain:
@dtillman I don't have any sample application which I can share here. 
Let me know if you have any sample which I can use and modify to replicate this issue.

vivekkjain:
@dtillman  Please find attached ConfigServerConfigurationProviderTest.zip with test project to reproduce this issue.  

[ConfigServerConfigurationProviderTest.zip](https://github.com/SteeltoeOSS/steeltoe/files/3903684/ConfigServerConfigurationProviderTest.zip)

dtillman:
Thanks for the sample test... I've pushed a fix .. and added your unit test .  Should see this appear in the next 2.x release.
Dave

vivekkjain:
@dtillman added review comment please address.

vivekkjain:
@dtillman any update on PR. It’s not yet merged.

PR #185, opened 2019-12-02
Config Server Data dictionary not cleared before Load

dtillman:
[#184] The Config Server's Data dicitonary needs to be cleared before loading it

### DIFF
--- a/src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
+++ b/src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
@@ -264,16 +264,16 @@ namespace Steeltoe.Extensions.Configuration.ConfigServer
                             "Located environment: {name}, {profiles}, {label}, {version}, {state}", env.Name, env.Profiles, env.Label, env.Version, env.State);
                         if (updateDictionary)
                         {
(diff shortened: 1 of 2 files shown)


src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
```diff
@@ -264,6 +264,8 @@ internal ConfigEnvironment DoLoad(bool updateDictionary = true)
                             "Located environment: {name}, {profiles}, {label}, {version}, {state}", env.Name, env.Profiles, env.Label, env.Version, env.State);
                         if (updateDictionary)
                         {
+                            Data.Clear();
```
vivekkjain: @dtillman I won't add this one here, this could impact requests which are currently reading the dictinoary value. I would add another local temp dictonary and add the new values init and replace with existing one .

dtillman: Ahh .. great point .. thanks!
Dave

jkonicki:
No worries, @vivekkjain.  We always merge the 2.x branch forward to master (3.x branch).  

Thanks for opening the issue and the review.

Issue #196, opened 2019-12-18
Automatically disable color in console logging on PCF


TimHess:
In Dynamic Console logging, `if (Platform.IsCloudFoundry)` then configure color to be disabled in the underlying console loggers so that Steeltoe users to not have to set `logging:console:disablecolors` to `true`

cesinhaugusto:
Would overriding the Configure method in ConsoleLoggerOptionsSetup (DynamicLoggingBuilder class) be a good approach to this?

Something like this:
```C#
internal class ConsoleLoggerOptionsSetup : ConfigureFromConfigurationOptions<ConsoleLoggerOptions>
{
    public ConsoleLoggerOptionsSetup(ILoggerProviderConfiguration<ConsoleLoggerProvider> providerConfiguration)
        : base(providerConfiguration.Configuration)
    {
    }

    public override void Configure(ConsoleLoggerOptions options)
    {
        if (Platform.IsCloudFoundry)
        {
            options.DisableColors = true;
        }

        base.Configure(options);
    }
}
`

TimHess:
Yeah @cesinhaugusto, I think that looks like a good place for it

cesinhaugusto:
Great, I'll prepare a pull request for that. I'm just having a little trouble creating a test to assert this.

PR #216, opened 2020-01-13
Disable console logging color automatically on PCF. Addresses #196

cesinhaugusto:
Guys, made the suggested modifications to solve #196. However, I was unable to find a way to unit test it, because I couldn't find a way to retrieve the DisableColors property. If you think it is extremely necessary to generate a unit test for this change, I'd ask you to help me find a way.

Additionally, as it was necessary to reference Platform, I had to add a reference to Steeltoe.Common, which also required to upgrade LoggingVersion from 2.1.0 to 2.20.

### DIFF
--- a/src/Logging/Logging.sln
+++ b/src/Logging/Logging.sln
@@ -34,6 +34,8 @@ Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "dependencies", "dependencie
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Steeltoe.Common.TestResources", "..\Common\test\Common.TestResources\Steeltoe.Common.TestResources.csproj", "{B78BAA02-862A-4DD3-8E9F-407FEDA3C950}"
 EndProject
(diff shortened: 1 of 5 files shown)


TimHess:
Hey, this looks functional and works in my PCF environment, thank you! There are just a couple changes needed for .csproj and .sln files, and we can test this code... it didn't work when I tried to push changes to your branch, but if you could either pull them in or re-implement them, this is about what I'm looking for: https://github.com/SteeltoeOSS/steeltoe/commit/bc106d94ccb94cc82bc58d86cb92bab0c776e57e 

If that's too much of a hassle I can just come back through and clean it up later

cesinhaugusto:
I'll gladly pull them, don't worry. My bad on the missing refs on .csproj and .sln. Simply forgot them.
Also, thanks for writing the tests. I wrote the same tests but was unable to think on how to write this line: `var options = services.GetService(typeof(IOptions<ConsoleLoggerOptions>)) as IOptions<ConsoleLoggerOptions>;`

It's always good to keep learning!

PR #217, opened 2020-01-16
Serilog Destructure fix

### LINKED ISSUE #182: SerilogDynamicProvider.CreateLogger negates Serilog destructuring configurations
## Describe the bug
When using the SerilogDynamicLogger serilog destructuring configurations are ignored.

## Steps to reproduce 
1. Create a project using Steeltoe SerilogDynamicLogger and Destructurama.Attributed
2. Add the UseAttribute destructuring configuration
3. Add an object model containing a destructuring attribute to the log scope to be destructured
4. Write to the log

## Expected behavior
The expected behavior is for properties tagged with [NotLogged] will not show up in the log and [LogMasked] will show up masked.

 - Platform: Any
 - OS: Any
 - .NET Version .NET Core 2.2.0 & 3.0.0
 - Steeltoe Version 2.3.0 & 2.4.0
 - Any other library versions to note

## Additional context or links
Serilog handles destructuring and building the property tree before writing to any sinks, the current implementation of SerilogDynamicProvider.CreateLogger creates a new logger for each category and sets the global logger as the sink, all of the destructuring happens before serilog pipes to the sink and since the category logger has no destructuring configs it uses the defaults. By the time the properties reach the global logger where your serilog configs are loaded they are already translated to key/value pairs of strings with no additional type information or metadata.

problem is with this line: https://github.com/SteeltoeOSS/steeltoe/blob/master/src/Logging/src/SerilogDynamicLoggerCore/SerilogDynamicProvider.cs#L106

nelsontbrandon:
Fixes #182 Extracts the Serilog:Destructure configs from the IConfiguration object if present and uses it to configure the category logger

### DIFF
--- a/src/Logging/src/SerilogDynamicLoggerCore/SerilogDynamicProvider.cs
+++ b/src/Logging/src/SerilogDynamicLoggerCore/SerilogDynamicProvider.cs
@@ -38,7 +38,7 @@ namespace Steeltoe.Extensions.Logging.SerilogDynamicLogger
         private ConcurrentDictionary<string, LogEventLevel> _runningLevels = new ConcurrentDictionary<string, LogEventLevel>();
         private LogEventLevel? _defaultLevel = null;
         private bool disposed = false;
-        
+
         private IConfiguration _configuration;
 
         /// <summary>
@@ -53,7 +53,7 @@ namespace Steeltoe.Extensions.Logging.SerilogDynamicLogger
             {
                 throw new ArgumentNullException(nameof(configuration));
             }
-            
+
             _configuration = configuration;
 
             _serilogOptions = options ?? new SerilogOptions(configuration);
@@ -86,7 +86,7 @@ namespace Steeltoe.Extensions.Logging.SerilogDynamicLogger
             {
                 throw new ArgumentNullException(nameof(configuration));
             }
-            
+
             _configuration = configuration;
 
             _serilogOptions = options ?? new SerilogOptions(configuration);


hananiel:
@nelsontbrandon I am not at all clear on how this is supposed to work. Is this an example of how this could work or does this work as is? I am not super familiar with serilog's configuration abilities - but destructurama has no documentation of serilog settings you are using (ex. Serilog:Destructure). I understood this to be a feature of destructurama - Destructure.UsingAttributes. Where is that happening? Would you mind giving an example of how this would be used? Tests would be good too.

nelsontbrandon:
This is how I worked around the issue internally, I don't know if it's the best way to solve the issue but is an example of one solution.

Destructuring is a serilog feature, destructurama is just a destructuring plugin that adds type attribute support.

Some documentation on destructuring is available here: https://github.com/serilog/serilog/wiki/Structured-Data#customizing-the-stored-data

The config key "Serilog:Destructure" is from how IConfiguration structures nested keys, that's the equivalent of { "Serilog":{"Destructure":{}} in the config file and the Serilog.Settings.Configuration (this package is what steeltoe uses to configure serilog) package expects the serilog configs inside the "Serilog" key.

Essentially what this PR does is checks the app configs for the existence of destructuring (Serilog:Destructure) and if it finds it then adds just the destructuring configurations and any assembly references (Serilog:Using, this is so serilog knows where to find the types) to the configs used to create the category loggers.

Hopefully I've explained this well, but please let me know if you have any more questions.

nelsontbrandon:
> @nelsontbrandon this sounds good, can we create a working sample around this (or check it in if you already have it).

Ya, I'm not really sure why this PR is failing to build, the code builds locally and the error details on the PR doesn't appear to be in the code I added?

I also haven't contributed any unit tests around the destructuring config because I'm not really sure how to go about testing that

hananiel:
@nelsontbrandon To merge this PR, I would need some help get a working example. We don't need unit tests, I can write those but need to see how exactly the changes you have suggested would work.

nelsontbrandon:
@hananiel the PR I submitted is a working example, i'm not sure why the build was failing in your azure pipeline but the build log indicated it wasn't failing due to the code failing to build. I pushed another commit to try and trigger the build again but all I did was remove some white space.

The only thing this PR does really is adds this logic:
```csharp
            var seriloggerConf = new Serilog.LoggerConfiguration();
            if (_configuration != null && _configuration.GetSection("Serilog:Destructure").Exists())
            {
                var confBuilder = new ConfigurationBuilder();
                confBuilder.AddInMemoryCollection(_configuration.GetSection("Serilog")
                    .AsEnumerable().Where((kv) => kv.Key.StartsWith("Serilog:Destructure") || kv.Key.StartsWith("Serilog:Using")));
                seriloggerConf = seriloggerConf.ReadFrom.Configuration(confBuilder.Build());
            }

            var serilogger = seriloggerConf.MinimumLevel.ControlledBy(levelSwitch)
```

if your configuration contains a Serilog:Destructure key then it builds a serilog config object that includes destructuring, otherwise it operates the same as before.

hananiel:
@nelsontbrandon I was able to test it with this configuration below: `
   "Destructure": [
          
            {
                "Name": "ToMaximumDepth",
                "Args": { "maximumDestructuringDepth": 10 }
            },
            {
                "Name": "ToMaximumStringLength",
                "Args": { "maximumStringLength": 3 }
            },
            {
                "Name": "ToMaximumCollectionCount",
                "Args": { "maximumCollectionCount": 20 }
            }
        ],` 
and was able to recreate this issue. I now realize what the problem is. However it appears to me that a general fix might be to do this below.

```
 var serilogger = new Serilog.LoggerConfiguration()
                .ReadFrom.Configuration(_configuration)
                .MinimumLevel.ControlledBy(levelSwitch)
                .WriteTo.Logger(_globalLogger)
                .CreateLogger();
```
instead of :
```            var seriloggerConf = new Serilog.LoggerConfiguration();
             if (_configuration != null && _configuration.GetSection("Serilog:Destructure").Exists())
             {
                var confBuilder = new ConfigurationBuilder();
                confBuilder.AddInMemoryCollection(_configuration.GetSection("Serilog")
                    .AsEnumerable().Where((kv) => kv.Key.StartsWith("Serilog:Destructure") || kv.Key.StartsWith("Serilog:Using")));
                seriloggerConf = seriloggerConf.ReadFrom.Configuration(confBuilder.Build());
             }
```
This would take care of other configuration - I have tested this with destructure and it works. Ill test it tomorrow with other config and try to create some tests for this. Thanks for your work in identifying this issue.

nelsontbrandon:
Ah sorry @hananiel , I didn't understand what kind of example you were looking for at first.

The general fix is is what I had originally tried, but that has the problem of duplicating the log messages since it will duplicate the WriteTo configs.

category logger -> sinks -> global logger -> sinks

Since it would be hitting all the sinks twice, that would include writing to console twice if that's the configured sink, this is why I extract only the specific configs for destructuring.

hananiel:
Closing in favor of #235 which is these commits plus few enhancements and tests

PR #222, opened 2020-01-22
Update PlaceholderResolverExtensions.cs

antonpopov:
loggerFactory not passed to AddPlaceholderResolver extension method for IConfigurationBuilder.


TimHess:
Good catch, thanks for the PR

PR #223, opened 2020-01-27
Fix usages of IPAddress.Parse.

eerhardt:
In ConsulServerUtils.FixIPv6Address, it is possible to get a non-valid IPAddress - ex. "localhost". This would cause an unnecessary exception, which was caught and discarded. Changed to use TryParse instead.

In InetUtils, the underlying method has an IPAddress, which it ToStrings, only to parse it again later. The change is to simply flow the IPAddress object.

### DIFF
--- a/src/Common/src/Common/Net/InetUtils.cs
+++ b/src/Common/src/Common/Net/InetUtils.cs
@@ -102,13 +102,7 @@ namespace Steeltoe.Common.Net
                 return result;
             }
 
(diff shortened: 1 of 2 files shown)


jkonicki:
@eerhardt Thank you for the contribution.  We will be reviewing it soon.  

General Note to Steeltoe team: This fix is targeted for 2.x and 3.x.  Once merged into master it needs to be brought into 2.x branch.

PR #224, opened 2020-01-29
Commons updates for Streams

dtillman:


.editorconfig
```diff
@@ -57,3 +57,6 @@ csharp_new_line_before_members_in_anonymous_types = true
 
 # CS0168: Variable is declared but never used
 dotnet_diagnostic.CS0168.severity = suggestion
+
+# S3925: "ISerializable" should be implemented correctly
+dotnet_diagnostic.S3925.severity = none
```
TimHess: I don't see any `ISerializable` in this PR, so I don't think anything in here is affected by this change, but if you do want to turn off this rule completely we should also alter the [configuration for SonarCloud](https://github.com/SteeltoeOSS/steeltoe/blob/master/Steeltoe.ruleset#L269)

src/Common/src/Abstractions/Converter/IConverterRegistry.cs
```diff
@@ -0,0 +1,28 @@
...
+
+namespace Steeltoe.Common.Converter
+{
+    /// <summary>
+    /// For registering converters with a type conversion system.
+    /// </summary>
+    public interface IConverterRegistry
+    {
+        /// <summary>
+        /// Adds a generic convert to this registry
```
TimHess:
```suggestion
        /// Adds a generic converter to this registry
```

src/Common/src/Abstractions/Converter/IConverterRegistry.cs
```diff
@@ -0,0 +1,28 @@
...
+        /// </summary>
+        /// <param name="converter">the convert to add</param>
```
TimHess:
```suggestion
        /// <param name="converter">the converter to add</param>
```

src/Common/src/Abstractions/Converter/IGenericConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+using System.Collections.Generic;
+
+namespace Steeltoe.Common.Converter
+{
+    /// <summary>
+    /// Generic converter interface for converting betweeen two or more types
```
TimHess:
```suggestion
    /// Generic converter interface for converting between two or more types
```

src/Common/src/Abstractions/Converter/IGenericConverter.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    public interface IGenericConverter
+    {
+        /// <summary>
+        /// Gets the soruce and target types this converter can convert between.
```
TimHess:
```suggestion
        /// Gets the source and target types this converter can convert between.
```

src/Common/src/Common/Converter/ArrayToCollectionConverter.cs
```diff
@@ -0,0 +1,95 @@
...
+                return null;
+            }
+
+            var len = asArray.GetLength(0);
+            var arrayElementType = ConversionUtils.GetElementType(asArray.GetType());
+
+            var list = ConversionUtils.CreateCompatListFor(targetType);
+            if (list == null)
+            {
+                throw new InvalidOperationException("Unable to create compatable list");
```
TimHess:
```suggestion
                throw new InvalidOperationException("Unable to create compatible list");
```

src/Common/src/Abstractions/Util/IRunnable.cs
```diff
@@ -0,0 +1,28 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Steeltoe.Common.Util
+{
+    /// <summary>
+    /// An object that can be invoked syncronously via a run method.
```
TimHess:
```suggestion
    /// An object that can be invoked synchronously via a run method.
```

src/Common/src/Abstractions/Util/IAsyncRunnable.cs
```diff
@@ -0,0 +1,30 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System.Threading.Tasks;
+
+namespace Steeltoe.Common.Util
+{
+    /// <summary>
+    /// An object that can be invoked asyncronously via a run method.
```
TimHess:
```suggestion
    /// An object that can be invoked asynchronously via a run method.
```

src/Common/src/Common/Converter/CollectionToCollectionConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+                return ConversionUtils.CanConvertElements(
+                    ConversionUtils.GetElementType(sourceType), ConversionUtils.GetElementType(targetType), _conversionService);
+            }
+
+            return false;
+        }
+
+        public override object Convert(object source, Type sourceType, Type targetType)
+        {
+            // ICollection sourceCollection = source as ICollection;
```
TimHess:
```suggestion
```

src/Common/src/Common/Converter/CollectionToCollectionConverter.cs
```diff
@@ -0,0 +1,102 @@
...
+            var sourceCollection = source as IEnumerable;
+            if (sourceCollection == null)
+            {
+                return null;
+            }
+
+            var list = ConversionUtils.CreateCompatListFor(targetType);
+            if (list == null)
+            {
+                throw new InvalidOperationException("Unable to create compatable list");
```
TimHess:
```suggestion
                throw new InvalidOperationException("Unable to create compatible list");
```

src/Common/src/Common/Expression/CSharp/EvaluatioinContext.cs
```diff
@@ -0,0 +1,23 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+using System;
+
+namespace Steeltoe.Common.Expression.CSharp
+{
+    public class EvaluatioinContext : IEvaluationContext
```
TimHess:
```suggestion
    public class EvaluationContext : IEvaluationContext
```
Filename should be updated, too

src/Common/src/Common/Converter/StringToBooleanConverter.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return null;
+            }
+
+            value = value.ToLowerInvariant();
+            if (_trueValues.Contains(value))
+            {
+                return true;
+            }
+            else if (_falseValues.Contains(value))
```
TimHess: What about using `.Contains(value, StringComparison.InvariantCultureIgnoreCase)` instead of `.ToLowerInvariant`?

.editorconfig
dtillman: Yes .. lets turn this off

src/Common/src/Common/Converter/StringToBooleanConverter.cs
dtillman: I don't find that method on the ISet  (i.e. .Contains(value, StringComparison.InvariantCultureIgnoreCase)  what am I missing?

TimHess: Ah, right - that would be on a string. In the case of `HashSet` we can pass the IEqualityComparer in the constructor... eg: `new HashSet<string>(StringComparer.InvariantCultureIgnoreCase)`

Issue #246, opened 2020-02-14
ConfigServerConfigurationProvider not working with GetSection()


Karql:
Hi,

After this commit: https://github.com/SteeltoeOSS/steeltoe/commit/1f1279a821ade60d1f4a12c57e2405bad16e012f

ConfigServerConfigurationProvider not working with GetSection()
v2.4.0
![conf-2 4 0](https://user-images.githubusercontent.com/173098/74505422-6107a100-4ef7-11ea-81e2-a7b9f12d4b38.png)

v2.4.2
![conf-2 4 2](https://user-images.githubusercontent.com/173098/74505438-682eaf00-4ef7-11ea-8d36-5e0c06ed08de.png)

Like you can see property Realms.Source.AuthServerUrl should be overriden by config server but in new version this is not happening.

The problem is this line:
`IDictionary<string, string> newData = new Dictionary<string, string>();`
Dictionary without comparer.

Should be: `new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);` like in base class: https://github.com/aspnet/Configuration/blob/master/src/Config/ConfigurationProvider.cs,

Best regards,
Mateusz

Karql:
I've made PR https://github.com/SteeltoeOSS/steeltoe/pull/247 ;)

PR #247, opened 2020-02-14
Fix comparer in ConfigServerConfigurationProvider (#246)

Karql:
Fix https://github.com/SteeltoeOSS/steeltoe/issues/246


TimHess:
Thanks for the PR! 

*This should also be applied to the 2.x branch

PR #249, opened 2020-02-18
Fix content negotiation #120

Karql:
https://github.com/SteeltoeOSS/steeltoe/issues/120#issuecomment-587351045

### DIFF
--- a/src/Management/src/EndpointBase/ActuatorMediaTypes.cs
+++ b/src/Management/src/EndpointBase/ActuatorMediaTypes.cs
@@ -30,31 +30,20 @@ namespace Steeltoe.Management.EndpointBase
 
         public static string GetContentHeaders(List<string> acceptHeaders, MediaTypeVersion version = MediaTypeVersion.V2)
         {
-            string contentHeader = null;
+            string contentHeader = APP_JSON;
+            string versionContentHeader = GetContentTypeHeaderForVersion(version);
 
-            if (acceptHeaders != null && acceptHeaders.Count > 0)
+            if (acceptHeaders != null && acceptHeaders.Any(x => x == ANY || x == versionContentHeader))
             {
-                foreach (var acceptHeader in acceptHeaders)
-                {
-                    if (acceptHeader == ANY)
-                    {
-                        contentHeader = AllowedAcceptHeaders(version).First();
-                    }
-                    else
-                    {
-                        contentHeader = AllowedAcceptHeaders(version)
-                            .FirstOrDefault(header => acceptHeader == header);
-                    }
-                }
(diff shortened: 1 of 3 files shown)


src/Management/src/EndpointBase/ActuatorMediaTypes.cs
```diff
@@ -45,6 +45,11 @@ public static string GetContentHeaders(List<string> acceptHeaders, MediaTypeVers
                         contentHeader = AllowedAcceptHeaders(version)
                             .FirstOrDefault(header => acceptHeader == header);
                     }
+
```
hananiel: This is doing the correct thing if the highest version comes first in the list from client. I am not sure we can depend on that necessarily. Can you rework it to find the highest matching version. Eventually, we need to make the change to setup both versions and respond with the correct version of the response based on header. So the result needs to be(for now):  
```
                    // Configured version  | Accept        | Result
                    //          V1         |  ANY          |   V1   
                    //                     |  No Match     |   APPJSON
                    //                     |  V1           |   V1
                    //                     |  V2           |   V1
                    //                     |  V1,V2/V2/V1  |   V1
                    //          V2         |  ANY          |   V2
                    //                     |  No Match     |   APPJSON
                    //                     |  V1           |   APPJSON
                    //                     |  V2           |   V2
                    //                     |  V1,V2/V2,V1  |   V2

```
What you have does this, it wont work if it comes in a different order.

Karql: Done ;)

I've also refactored tests for it.

src/Management/src/EndpointCore/ContentNegotiation/ContentNegotiationExtensions.cs
```diff
@@ -48,24 +49,14 @@ public static void LogContentType(this ILogger logger, IHeaderDictionary request
...
+            var acceptMediaTypes = headers.Accept?.Select(x => x.MediaType.Value).ToList();
 
             var contentType = ActuatorMediaTypes.GetContentHeaders(acceptMediaTypes, version);
 
             responseHeaders.Add("Content-Type", contentType);
 
             logger?.LogContentType(requestHeaders, contentType);
         }
-
-        public static bool HasSupportedAcceptHeaders(this IHeaderDictionary requestHeaders, MediaTypeVersion version = MediaTypeVersion.V2)
```
Karql: It was not used anywhere

hananiel: We can drop that one.

Issue #208, opened 2020-01-09
Provide service to service authentication using rotating certificates


jkonicki:
Use Case:  Provide a way to do service to service authentication utilizing the certificates that are automatically provided (and rotated) by PCF

TimHess:
With this item, a mechanism should be provided for generating certificates during local development. These local certificates should look similar to certificates provided by PCF, such that authorization policies requiring the apps be in the same space and/or org work locally as well as on the platform.

TimHess:
Blocked on Linux on PCF by [this issue](https://github.com/dotnet/runtime/issues/32558)

Potential workaround could be to duplicate most of the functionality provided by `Microsoft.AspNetCore.Authentication.Certificate` using BouncyCastle for all certificate components

TimHess:
Relevant links for easy retrieval in case this work gets mothballed:

- https://docs.pivotal.io/platform/2-8/security/pcf-infrastructure/certificates-index.html
- https://docs.cloudfoundry.org/devguide/deploy-apps/instance-identity.html

jkonicki:
Waiting on Docs

jkonicki:
Closing now.  Docs to come in #297

PR #255, opened 2020-02-25
Service to Service Client Certificate Authentication

TimHess:
Built on [Microsoft.AspNetCore.Authentication.Certificate](https://docs.microsoft.com/en-us/aspnet/core/security/authentication/certauth?view=aspnetcore-3.1) but adds:

- support for specifying known valid signing certificates
- auto-discovery (and rotation) of Diego-provided identity certificates
- auto-generated client certificates that resemble those provided by Diego
- authorization policies for requiring client certificate from the same space and/or org

#208


src/Common/src/Common.Security/PemConfigureCertificateOptions.cs
```diff
@@ -60,71 +55,50 @@ public void Configure(string name, CertificateOptions options)
...
-            {
-                _logger?.LogError(e, "Unable to read PEM encoded certificate");
+                var der = Convert.FromBase64String(base64);
+                var rsa = RSA.Create();
+                rsa.ImportRSAPrivateKey(der, out _);
+                return rsa;
             }
 
-            return null;
+            // "BEGIN PRIVATE KEY" (ImportPkcs8PrivateKey),
```
hananiel: Could we delete this (since if we did do this - at least we would use a different name for method)?

src/Security/test/Authentication.CloudFoundryCore.Test/CloudFoundryContainerIdentityMtlsTest.cs
```diff
@@ -0,0 +1,20 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+namespace Steeltoe.Security.Authentication.CloudFoundryCore.Test
+{
+    public class CloudFoundryContainerIdentityMtlsTest
```
hananiel: Do we need this/ Do we need more tests ?

src/Security/src/Authentication.CloudFoundryCore/ConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,47 @@
...
+using System;
+using System.IO;
+
+namespace Steeltoe.Security.Authentication.CloudFoundry
+{
+    public static class ConfigurationBuilderExtensions
+    {
+        public static IConfigurationBuilder AddCloudFoundryContainerIdentity(this IConfigurationBuilder builder, string orgId = null, string spaceId = null)
+        {
+            if (!Platform.IsCloudFoundry)
```
TimHess: Open question: does this seem like the best way to handle certificate creation?

PR #265, opened 2020-03-25
Adding spring placeholder resolution

TheBritz:
Added a '.' to ':' replacement on identified placeholder before resolution to ensure that spring placeholders can be resolved. Attempted only after other resolution attempts have failed.


TimHess:
Thanks for your contribution @TheBritz!

PR #291, opened 2020-04-14
Fix failing tests

mmartell:
- Fixes Reflection of C# properties
- Adds support for PoolFactory


mmartell:
Also adds RegionFactory support

Issue #209, opened 2020-01-09
Retrieve all Kubernetes configuration data


jkonicki:
Use case:  Create a similar experience to `AddCloudFoundry` functionality, except `AddKubernetes` and the information from ConfigMap and Secrets is retrieved.

PR #304, opened 2020-04-24
Initial Kubernetes config support

TimHess:
Builds on official Kubernetes client, adding:
- `Steeltoe.Common.Kubernetes` with general Kubernetes-related app configuration and extensions (like a single way to get an `IKubernetes` for any new feature)
- `Steeltoe.Extensions.Configuration.KubernetesBase` with config providers for ConfigMaps and Secrets #209
- `Steeltoe.Extensions.Configuration.KubernetesCore` with extensions for `Host` and `WebHost`


src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
```diff
@@ -0,0 +1,108 @@
...
+            K8sClient = kubernetes;
+            Settings = settings;
+        }
+
+        public override void Load()
+        {
+            try
+            {
+                var configMapResponse = K8sClient.ListNamespacedConfigMapWithHttpMessagesAsync(Settings.Namespace ?? "default", fieldSelector: $"metadata.name={Settings.Name}", watch: Settings.Watch).GetAwaiter().GetResult();
+                ProcessData(configMapResponse.Body.Items?.FirstOrDefault());
```
macsux, line 55: This is incorrect, if `watch` is true the result will not be deserialized into a V1List type and must be read through watcher

TimHess: This is a good point, its a leftover from a previous approach that didn't end up making sense, thanks

Issue #299, opened 2020-04-21
OpenTelemetry/.NET Diagnostics library implementation for our metric subsystem


jkonicki:
Moving from OpenCensus (no longer active project) to OpenTelemetry or .NET Diagnostic library implementations.

Targeting support for : Metric actuator, CLR runtime metrics, Distributed Tracing, and Circuit Breakers events throughout Steeltoe.

jkonicki:
#134, #256, #264 we added in 3.0.0.  
#276 will be added once the OTel beta/GA is available along with any necessary changes.

hananiel:
Need to update #711 when done with this work.

hananiel:
This is resolved by https://github.com/SteeltoeOSS/Steeltoe/pull/828

PR #305, opened 2020-04-25
Add CLR Eventsource and EventCounter Metrics

### LINKED ISSUE #256: Update CLR Metrics to use EventCounters
Currently we use an "IPolledDiagnosticSource" which is a steeltoe abstraction that polls CLR metrics. 
EventCounters give a new way to get these metrics in process (and also expose out of proc via EventPipe).  We need to update the CLR Metrics Observer to use EventCounter. Other observers such aspnetcore still publish via DiagnosticSource/Listener and that code can remain the same. 
https://github.com/dotnet/diagnostics/blob/eb13265352569304912eedaa66894eb760d6c158/documentation/design-docs/eventcounters.md

hananiel:
Metrics end-end with metrics and prometheus actuators
Make updates to MetricsEndpoint to serialize metrics correcty
Resolves #256 and Partially #299


src/Management/test/EndpointCore.Test/ContentNegotiation/ContentNegotiationTests.cs
```diff
@@ -66,16 +66,15 @@ public static IEnumerable<object[]> EndpointMiddleware_ContentNegotiation_TestCa
...
                     }
                 }
             }
         }
 
-        /// <param name="version">For now there is no way to configure version - defined for future use</param>
         [Theory]
         [MemberData(nameof(EndpointMiddleware_ContentNegotiation_TestCases))]
-        public async void EndpointMiddleware_ContentNegotiation(EndpointNames epName, string epPath, MediaTypeVersion version, string[] accepts, string contentType, string name)
+        public async void EndpointMiddleware_ContentNegotiation(EndpointNames epName, string epPath, string[] accepts, string contentType, string name)
```
TimHess, line 77: @hananiel This might be one to add an exception for (in tests)

src/Management/src/EndpointBase/Metrics/Observer/EventSourceListener.cs
```diff
@@ -0,0 +1,69 @@
...
+
+                var metricName = $"{eventSourceName}.{nameEnumerator.Current}";
+                var currentMetric = MeasureMetrics.GetOrAddEx(
+                    metricName,
+                    (name) => _stats.Meter.CreateInt64Measure(name));
+
+                var actualValue = Convert.ToInt64(payloadEnumerator.Current, CultureInfo.InvariantCulture);
+                currentMetric.Record(default(SpanContext), actualValue, labels.ToList());
+
+             // _logger.LogDebug($"MetricName= {metricName} value = {actualValue}");
```
TimHess, line 65: This appears to be the only logger entry in this file and now it's commented out... if you're going to leave the logger in the class, you might consider this change
```suggestion
               _logger?.LogDebug($"MetricName= {metricName} value = {actualValue}");
```

Issue #310, opened 2020-04-28
MongoDB Health Check Always Fails


TimHess:
## Describe the bug
`MongoDbHealthContributor` always reports "down" status

> In Steeltoe sources, a CancellationTokenSource is passed to ListDatabases method, when it is expecting a CancellationToken instead, causing an exception that is interpreted as a failure in MongoDB connection

https://steeltoeteam.slack.com/archives/C0ENXS0M8/p1587978471050300

PR #312, opened 2020-04-28
Fix MongoDb Health Check (2.x)

TimHess:
Provide a CancellationToken instead of its Source #310


TimHess:
That is so weird... thanks `git` for the merge "help" and thanks @terolaakso for the real help!

PR #313, opened 2020-04-29
Fix heartbeat interval calculation

Vakadavr:
It seems like interval calculations for consul's heartbeat does not respect TtlUnit parameter.
F.e. ttl "10m" with ratio 0.1 will produce 1 second interval instead of 1 minute.
Also I've merged multiple tests into one theory and extent cases a bit.


jkonicki:
Thanks for the contribution.  
Note for team:  This needs to pulled into the 2.x branch also.

Issue #320, opened 2020-05-05
Enable Deterministic Builds


TimHess:
Deterministic builds are important as they enable verification that the resulting binary was built from the specified source and provides traceability.

Follow this example https://github.com/clairernovotny/DeterministicBuilds

TimHess:
Not merged in 3.0 yet

PR #325, opened 2020-05-08
Enable deterministic builds, fix comment

TimHess:
#320 for 2.x branch, fix a comment that isn't accurate


TimHess:
The problem with the comment on `Port` was raised here: https://github.com/SteeltoeOSS/Documentation/pull/13

Issue #319, opened 2020-05-04
Consul client should not defaults to TTL when health is specified


hananiel:
Currently consul defaults to using TTL even if health endpoints are configured. It should default to TTL only if no other health check options are configured. Workaround is to use this here: https://github.com/SteeltoeOSS/Steeltoe/issues/309#issuecomment-622485201_

PR #332, opened 2020-05-19
Fix DeregisterCriticalServiceAfter configuration on ttl healthchecks - Consul

Vakadavr:
DeregisterCriticalServiceAfter property doesn't set properly if ttl healthcheck enabled.
I think it is not correct, because DeregisterCriticalServiceAfter feature work pretty well with ttl healthchecks too.
Note: this pr modify default behaviour and can broke backward compatibility a bit. I don't know if it is critical or not =)


hananiel:
We need to fix this in a way to fix #319. I'll add these changes.

Issue #331, opened 2020-05-15
Consul discovery compute the loopback address first


zh6335901:
Hi, I have a service instance using consul discovery. But the disovery component register the loopback address to consul. I think it should be register the internal ip address to consul first, not loopback address.
My service  discovery configuraion:
```
  "Consul": {
    "Host": "10.1.3.63",
    "Port": 8500,
    "Discovery": {
      "Heartbeat": {
        "Enabled": false
      },
      "ServiceName": "ClientManager",
      "Port": 9030,
      "PreferIpAddress": true,
      "HealthCheckPath": "/liveness",
      "HealthCheckInterval": "20s"
    }
  }
```
Consul dashboard shows:
![image](https://user-images.githubusercontent.com/3895438/82044168-6daab980-96df-11ea-8802-197bc0ed21d4.png)
My service's machine info:
![image](https://user-images.githubusercontent.com/3895438/82044345-c37f6180-96df-11ea-80f3-644eefb24a18.png)

zh6335901:
I know it can set specified IpAddress in the consul configuraion section. But it need set every time on different machine.

TimHess:
@zh6335901 we would agree that setting the IP address for each machine isn't a workable solution. I'm not exactly sure how that's happening for you, but we do already have some code in Steeltoe.Common that should provide additional flexibility in this area, I should be able to start a PR that uses it today

PR #334, opened 2020-05-20
Use InetUtils for Discovery

TimHess:
Provides more flexibility with address registration, should help with #331


dtillman:
LGTM... did just a quick scan..

PR #339, opened 2020-05-29
Update Consul dependency on System.Net.Http.WinHttpHandler

hananiel:


src/Discovery/src/ConsulBase/Steeltoe.Discovery.ConsulBase.csproj
```diff
@@ -14,6 +14,7 @@
     <PackageReference Include="Microsoft.Extensions.Logging" Version="$(ExtensionsVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="$(ExtensionsVersion)" />
     <PackageReference Include="Newtonsoft.Json" Version="$(JsonNetVersion)" />
+    <PackageReference Include="System.Net.Http.WinHttpHandler" Version="$(HttpVersion)" />
```
TimHess, line 17: `$(HttpVersion)` = 4.3.3, we may need to bump it also

hananiel: The latest available version of System.Net.Http is 4.3.4. I think I will make a new variable for WinHttpHandler and bump it to latest stable version of 4.7.0

PR #340, opened 2020-05-29
Health async interfaces

hananiel:
Draft for adding a new IAsyncHealthContributor with CheckHealthAsync() method. Steeltoe already supports adding IHealthContributor implementations in addition to Microsoft [HealthCheckRegistration](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.diagnostics.healthchecks.healthcheckregistration?view=dotnet-plat-ext-3.1)s. In order to be flexible to Steeltoe consumers, who want to continue using IHealthContributor, we want to support IAsyncHealthContributor in addition to the other options.


hananiel:
I will recreate this by cherry-picking the last commit.

PR #345, opened 2020-06-01
Use streamlined, dateless copyright headers

TimHess:


TimHess:
CodeCov will fail until this PR is merged: https://github.com/codecov/codecov-exe/pull/105

jkonicki:
Copyright update looks good.  I think you need to fix the comment referring to 3.1.300 to `post 3.1.300`, since it doesn't address all the problems we have.

PR #351, opened 2020-06-08
Bug: Token exchanger RequiredScopes

macsux:
Scopes string is incorrectly formed when required RequiredScopes is specified


jkonicki:
This needs to be brought into 2.x.   We can release this for 2.4.4, if we get it in today.

TimHess:
to be added to #350 after #352

PR #354, opened 2020-06-16
Skip InetUtils tests on hosted Mac, target netcoreapp3.1

TimHess:
Several tests recently started failing on MacOS CI builds though no Steeltoe code changed, this PR skips those on Mac
Also:
- updates dependencies and tests to use netcoreapp3.1 instead of netcoreapp3.0
- .csproj and pipeline maintenance


TimHess:
Our test failures seem to be caused by [this issue](https://github.com/actions/virtual-environments/issues/1042), revisit after azure pipelines get the update (next week?)

Issue #355, opened 2020-04-11
RabbitMQ health check creates a new connection during health check.


VenkateshSrini:
## Describe the bug
The points that I'm going to state is opinionated based on the feedback given my CF cluster management team. We are not using rabbit mq shipped along with PCF platform, instead have hosted the a Rabbit MQ in separate Infrastructure as a service. Issue is when I use the rabbitmq health check in my app ( the default one provided by SteelToe) for performing health check it creates a new connection every time. This causes lots of connection churn. The recommendation was use to use an already existing connection factory and check the connection state. If it is broken then try creating a new connection and that too fails report a failure. 

In fact, this recommendation was given in consensus with Pivotal application expert team

## Steps to reproduce 
1. Create a application with rabbit MQ health check
2. Observer the connection tab in RabbitMQ UI

## Expected behavior
The rabbit MQ health should use an existing connection not new one. There should not be any additional connection churns created because of health check

## Environment (please complete the following information):
 - Sample name: [e.g. RabbitMQ Connector, Discovery, etc.]:Rabbit MQ
 - Platform: [e.g. Cloud Foundry, Azure, etc.) : Pivotal cloud foundry
 - OS: [e.g. Windows, Linux, Mac OSX]:Linux
 - .NET Version [e.g. .NET Core 2.1.0, .NET Framework 4.7.1, etc.] :.NET core 3.1
 - Steeltoe Version [e.g. 2.3.0]: Latest
 - Any other library versions to note

PR #356, opened 2020-06-16
keep IConnection between RMQ health check requests

TimHess:
Also include server version in status

#355

### DIFF
--- a/src/Connectors/src/ConnectorBase/Queue/RabbitMQHealthContributor.cs
+++ b/src/Connectors/src/ConnectorBase/Queue/RabbitMQHealthContributor.cs
@@ -7,7 +7,9 @@ using Microsoft.Extensions.Logging;
 using Steeltoe.CloudFoundry.Connector.Services;
 using Steeltoe.Common.HealthChecks;
 using System;
+using System.Collections.Generic;
 using System.Reflection;
+using System.Text;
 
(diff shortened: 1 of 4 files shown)


src/Connectors/src/ConnectorBase/Queue/RabbitMQHealthContributor.cs
```diff
@@ -49,19 +50,31 @@ public HealthCheckResult Health()
             var result = new HealthCheckResult();
             try
             {
-                var connection = ConnectorHelpers.Invoke(RabbitMQTypeLocator.CreateConnectionMethod, _connFactory, null);
+                connection ??= ConnectorHelpers.Invoke(RabbitMQTypeLocator.CreateConnectionMethod, _connFactory, null);
```
macsux, line 53: I don't think this is right. You're injecting RabbitMQProviderConnectorFactory, which serves up instances of `ConnectionFactory`, which in turn are used to create an actual connection. It is the actual connection that you should be reusing, hence this should be a field not a local variable. Alternatively, the connection must be disposed of or it will keep TCP session open to the server till it's garbage callected and Dispose is called implicitly.

TimHess: Connection is stored in a field and is re-used... will look at enforcing `_` on fields in the future so it is more clear

PR #376, opened 2020-07-01
Update Actuator Extensions to enable Security

hananiel:
* The MapXXXActuator and MapAllActuators now return IEndpointConventionBuilder allowing MapXXXActuator().RequireAuthorization()
* The MapAllActuators takes an optional EndpointConventionBuilderCollection to apply the added convention to all
of the endpoints.

### DIFF
--- a/src/Management/src/EndpointCore/ActuatorRouteBuilderExtensions.cs
+++ b/src/Management/src/EndpointCore/ActuatorRouteBuilderExtensions.cs
@@ -21,6 +21,7 @@ using Steeltoe.Management.Endpoint.ThreadDump;
 using Steeltoe.Management.Endpoint.Trace;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Steeltoe.Management.Endpoint
 {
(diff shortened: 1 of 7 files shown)


src/Management/src/EndpointCore/ActuatorRouteBuilderExtensions.cs
```diff
@@ -54,19 +55,69 @@ public static (Type middleware, Type options) LookupMiddleware(Type endpointType
...
             if (endpoints == null)
             {
                 throw new ArgumentNullException(nameof(endpoints));
             }
 
-            var (middleware, optionsType) = LookupMiddleware(typeof(TEndpoint));
+            var (middleware, optionsType) = LookupMiddleware(typeEndpoint);
             var options = endpoints.ServiceProvider.GetService(optionsType) as IEndpointOptions;
             var mgmtOptionsCollection = endpoints.ServiceProvider.GetServices<IManagementOptions>();
+            var builder = conventionBuilder ?? new EndpointCollectionConventionBuilder();
```
TimHess, line 120: it might be clearer to do `conventionBuilder ??= new EndpointCollectionConventionBuilder();` here

src/Management/test/EndpointCore.Test/ManagementHostBuilderExtensionsTest.cs
```diff
@@ -504,7 +505,25 @@ public async Task AddAllActuators_IHostBuilder_IStartupFilterFires()
         }
 
         [Fact]
-        public void AddaLLActuators_IHostBuilder()
+        public async Task AddAllActuatorsWithConventions_IHostBuilder_IStartupFilterFires()
```
TimHess, line 508: Should there be another test or at least assertion that confirms failure when an unauth'd request hits an auth required actuator?

TimHess: After further review of `MapTestAuthSuccess` this suggestion would likely be excessive

PR #378, opened 2020-07-08
Rebuild Discovery.ClientCore, Add NoOp and Config-based Options

TimHess:
With these changes:
- ClientCore no longer requires a package/project reference to every client
- A NoOp client is available if none is provided (assumes DNS or other platform-based fallback is in place)
- Adds a helper to check if the app is containerized (based on a setting from MSFT-provided images so not **100%** reliable)
- Removes some setup extensions that were never even updated to support Consul


TimHess:
For review... does the NoOp client make sense when we have the config-based provider? There would be a similar pass-through effect, but with the benefit of being able to plug something in via config

src/Discovery/src/ClientCore/SimpleClients/NoOpDiscoveryClient.cs
```diff
@@ -0,0 +1,42 @@
...
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace Steeltoe.Discovery.Client
+{
+    internal class NoOpDiscoveryClient : IDiscoveryClient
+    {
+        internal NoOpDiscoveryClient(ILogger<NoOpDiscoveryClient> logger = null)
+        {
+            logger?.LogWarning("A NoOp Discovery Client has been created! Running in container: {IsContainerized}", Platform.IsContainerized);
```
ccheetham, line 18: nit ... The "!" seems redundant in that the message is warning level

src/Discovery/src/EurekaBase/EurekaDiscoveryClientBuilderExtension.cs
```diff
@@ -0,0 +1,22 @@
...
+
+namespace Steeltoe.Discovery.Eureka
+{
+    public static class EurekaDiscoveryClientBuilderExtension
+    {
+        /// <summary>
+        /// Configures <see cref="EurekaDiscoveryClient"/> as the <see cref="IDiscoveryClient"/> of choice
+        /// </summary>
+        /// <param name="clientBuilder"><see cref="DiscoveryClientBuilder"/></param>
+        /// <param name="serviceInfoName">Optionally specify the name of a specific Eureka service binding to look for</param>
```
ccheetham, line 15: remove "to look for"?

src/Discovery/src/EurekaBase/EurekaDiscoveryClientExtension.cs
```diff
@@ -0,0 +1,109 @@
...
+
+                return eurekaService;
+            });
+
+            services.AddSingleton<IHealthContributor, EurekaServerHealthContributor>();
+
+            var serviceProvider = services.BuildServiceProvider();
+            var certOptions = serviceProvider.GetService<IOptions<CertificateOptions>>();
+            var existingHandler = serviceProvider.GetService<IHttpClientHandlerProvider>();
+            if (certOptions != null && existingHandler is null)
```
ccheetham, line 103: is this mix of null/not null a common csharp practice?

Alternatives that _seem_ more consistent:
```
(certOptions != null && existingHandler == null)
(!(certOptions is null) && existingHandler is null)
```

src/Discovery/src/ClientCore/SimpleClients/NoOpDiscoveryClient.cs
jkonicki: I think in general this message is not clear.  As a user, I would be confused as to why this was a warning.  I think it is good to inform why this is being used.  An example message that would be helpful 
-- Information message: "No discovery client has been configured, using default no-op discovery client." 
-- Information message 2: "Running in container: {IsContainerized}".

src/Discovery/src/ClientCore/SimpleClients/NoOpDiscoveryClient.cs
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        public string Description => "The IDiscoveryClient that didn't";
```
jkonicki, line 21: I don't think this is a good description.

src/Discovery/src/EurekaBase/EurekaDiscoveryClientExtension.cs
TimHess:
```suggestion
            if (certOptions is object && existingHandler is null)
```
StackOverflow has this as a suggestion... thoughts?

src/Discovery/src/ClientCore/SimpleClients/NoOpDiscoveryClient.cs
TimHess: Fair point. I wasn't taking it very seriously because I don't think description is actually used anywhere, but we could do something like this
```suggestion
        public string Description => "An IDiscoveryClient that passes through to underlying infrastructure";
```

src/Discovery/src/EurekaBase/EurekaDiscoveryClientExtension.cs
ccheetham: looks good

src/Discovery/src/ClientCore/SimpleClients/NoOpDiscoveryClient.cs
jkonicki: Much better.

src/Discovery/src/EurekaBase/EurekaDiscoveryClientBuilderExtension.cs
TimHess:
```suggestion
        /// <param name="serviceInfoName">Optionally specify the name of a specific Eureka service binding</param>
```

src/Discovery/src/ClientCore/SimpleClients/NoOpDiscoveryClient.cs
TimHess:
```suggestion
            logger?.LogWarning("No discovery client has been configured, using default no-op discovery client.");
            logger?.LogInformation("Running in container: {IsContainerized}", Platform.IsContainerized);
```

src/Discovery/src/ClientCore/DiscoveryServiceCollectionExtensions.cs
```diff
@@ -2,327 +2,120 @@
...
         }
 
-        public static IServiceCollection AddDiscoveryClient(this IServiceCollection services, IConfiguration config, string serviceName, IDiscoveryLifecycle lifecycle = null)
+        public static IServiceCollection AddDiscoveryClient(this IServiceCollection services, IConfiguration config, string serviceName = null, IDiscoveryLifecycle lifecycle = null)
         {
-            if (services == null)
-            {
-                throw new ArgumentNullException(nameof(services));
-            }
+            Action<DiscoveryClientBuilder> builderAction = null;
```
TimHess, line 36: New reflection-based client setup option here, combining original config-driven approach with `IDiscoveryClientExtension`

Issue #381, opened 2020-07-24
Remove Newtonsoft-based HttpClient Json Extensions


TimHess:
Steeltoe.Common.Http includes several extension methods that handled (de)serializion to/from Json (via Newtonsoft) with HTTP interactions. New versions of these extensions are now shipped in `System.Net.Http.Json` and ours are no longer needed.

Remove these extensions, update code that used them and convert as much to System.Text.Json as is feasible

Issue #281, opened 2020-04-07
Add UserAgent headers for SCS services


jkonicki:
Consider adding useragent headers to SCS services like Config Server, eureka discovery, circuitbreaker, and maybe SC gateway.

PR #382, opened 2020-07-24
Add User-Agent, Reduce Newtonsoft

TimHess:
- Set a user-agent header on most/all outgoing Steeltoe requests #281
- Remove HttpClientExtensions that aren't needed anymore #381
- Remove Newtonsoft from Discovery, Configuration, Management (tests) and Security


TimHess:
SonarCloud is failing due to detected duplication in `EurekaHttpClient` on what is effectively old code. The scanner is detecting it as new/changed because of the change to simple `using`s

Issue #285, opened 2020-04-08
Refactoring of connectors


jkonicki:
Originally connectors were built for CF.  With our move to a more platform agnostic approach to Steeltoe, we need to take a look at how we can better abstract out connector functionality for usage on other platforms.  

related ticket: #26

TimHess:
Consider better compatibility with [`GetConnectionString` ](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationextensions.getconnectionstring?view=dotnet-plat-ext-3.1)

Issue #26, opened 2017-07-10
How do you add a custom ServiceInfoFactory?


tillig:
We are using the Vault service broker for PCF and I'm trying to follow the pattern in this repo for a nice, robust connection mechanism.

The roadblock I'm hitting is that most of these start with something like:

```csharp
config.GetSingletonServiceInfo<RedisServiceInfo>();
```

This invokes, a roundabout way, [the `CloudFoundryServiceInfoCreator.BuildServiceInfoFactories` method](https://github.com/SteeltoeOSS/Connectors/blob/master/src/Steeltoe.CloudFoundry.Connector/CloudFoundryServiceInfoCreator.cs#L117), which scans _only the `Steeltoe.CloudFoundry.Connector` assembly_ for service info factory implementations.

This `ServiceInfo` that gets created is later used in [extension methods like this one for Redis](https://github.com/SteeltoeOSS/Connectors/blob/master/src/Steeltoe.CloudFoundry.Connector.Redis/RedisCacheServiceCollectionExtensions.cs#L45) as the "base configuration" for a connection, left to be enhanced by additional client configuration.

Problem is, I can't figure out how to get my service info factory into play to be used when service info is getting retrieved. The set of service info objects gets cached with individual `CloudFoundryServiceInfoCreator` instances so while I could create some sort of additional extension method that might work like this...

```csharp
config
  .AddVaultServiceInfo()
  .GetSingletonServiceInfo<VaultServiceInfo>();
```

...having to remember to add that call every time isn't so cool.

If there's _not_ already a solution for this, I could imagine one possible solution being the `ServiceInfo` implementation getting marked with an attribute indicating its factory type and having the `CloudFoundryServiceInfoCreator` discover factories in that way rather than (or in addition to?) assembly scanning.

However, I didn't see anything in the samples repo showing this so I figured I'd ask.

jkonicki:
Need to keep in mind during the 3.x refactoring.

PR #388, opened 2020-07-31
Connectors Enhancements

TimHess:
More enhancement than refactor #285 at this point, but:
- allows external `IConnectionInfo`, `ServiceInfoFactory` and `ServiceInfoCreator` #26 
- cuts ConnectorBase's direct dependency on CloudFoundryConfig
- adds assembly attributes (and helper logic) for simple identification of assemblies that contain types with an attribute or base type Steeltoe might be looking for
- adds more reflection helpers
- adds new config provider that allows access to Connector-built connection strings  from `IConfiguration.GetConnectionString()` using either the service id/name or binding type
- allow setting timeout on MongoDb Healthcheck


src/Connectors/src/ConnectorBase/ConnectionTypeLocator.cs
```diff
@@ -0,0 +1,55 @@
...
+
+namespace Steeltoe.Connector
+{
+    internal static class ConnectionTypeLocator
+    {
+        internal static Type GetConnectionInfoType(string input)
+        {
+            return true switch
+            {
+                bool _ when input.Equals("cosmosdb", StringComparison.InvariantCultureIgnoreCase) => typeof(CosmosDbConnectionInfo),
```
ccheetham, line 24: This feels wrong, but hard to put finger on it.  E.g., the fact that "mysql" relates to `MySqlConnectionInfo` should be encapsulated in the `MySqlConnectionInfo` class (or related if using a factory of some sort).

So something more along lines of:
```
for info in connection infos:
  if connection info accepts input:
    return connection info
throw exception
```

ccheetham: But even knowing the connection classes seems wrong as well.  Could the connection classes register them selves upon init?  Otherwise wouldn't this code need to be updated when a new connection type is added?

src/Connectors/src/ConnectorBase/DocumentDB/MongoDb/MongoDbConnectorOptions.cs
```diff
@@ -47,7 +47,7 @@ public MongoDbConnectorOptions(IConfiguration config)
 
         public string Database { get; set; }
 
-        public Dictionary<string, string> Options { get; set; }
+        public Dictionary<string, string> Options { get; set; } = new Dictionary<string, string>(StringComparer.InvariantCultureIgnoreCase);
```
ccheetham, line 50: Does the invariant culture aspect provide value here?

src/Connectors/src/ConnectorBase/DocumentDB/MongoDb/MongoDbHealthContributor.cs
```diff
@@ -25,18 +26,29 @@ public static IHealthContributor GetMongoDbHealthContributor(IConfiguration conf
             var info = configuration.GetSingletonServiceInfo<MongoDbServiceInfo>();
             var mongoOptions = new MongoDbConnectorOptions(configuration);
             var factory = new MongoDbConnectorFactory(info, mongoOptions, MongoDbTypeLocator.MongoClient);
-            return new MongoDbHealthContributor(factory, logger);
+            var timeout = 5000;
```
ccheetham, line 29: Should default `timeout` be externally configurable?  Or at least a named constant?

src/Connectors/src/ConnectorBase/Queue/RabbitMQConnectionInfo.cs
```diff
@@ -7,20 +7,29 @@
...
+
+        private Connection GetConnection(RabbitMQServiceInfo info, IConfiguration configuration)
+        {
             var rabbitConfig = new RabbitMQProviderConnectorOptions(configuration);
             var configurer = new RabbitMQProviderConfigurer();
             return new Connection
             {
                 ConnectionString = configurer.Configure(info, rabbitConfig),
-                Name = "RabbitMQ" + serviceName?.Insert(0, "-")
+                Name = "RabbitMQ" + info?.Id?.Insert(0, "-")
```
ccheetham, line 32: This snippet appears several times over in connectors:
```
  name = <connector> + <info> + some mod
```

Is it possible to refactor?

src/Connectors/src/ConnectorBase/ServiceInfoCreator.cs
```diff
@@ -15,22 +16,28 @@ namespace Steeltoe.Connector
...
     public class ServiceInfoCreator
     {
+        private static readonly object _lock = new object();
         private static IConfiguration _config;
         private static ServiceInfoCreator _me = null;
-        private static object _lock = new object();
 
-        internal ServiceInfoCreator(IConfiguration config)
-        {
 #pragma warning disable S3010 // Static fields should not be updated in constructors
```
ccheetham, line 23: The warning seems valid.  In parallel or concurrent environment, `_config` becomes undetermined.

src/Connectors/test/ConnectorBase.Test/ConnectionStringConfigurationProviderTest.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Steeltoe.Connector.Test
+{
+    public class ConnectionStringConfigurationProviderTest
```
ccheetham, line 16: Purpose of this class?

src/Connectors/test/ConnectorBase.Test/ConnectionStringConfigurationSourceTest.cs
```diff
@@ -0,0 +1,14 @@
...
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Steeltoe.Connector.Test
+{
+    public class ConnectionStringConfigurationSourceTest
```
ccheetham, line 13: Purpose of this class?

TimHess: I think I pushed contents to this file while you were reviewing

src/Connectors/src/Abstractions/ServiceInfoCreatorAssemblyAttribute.cs
```diff
@@ -0,0 +1,23 @@
...
+namespace Steeltoe.Connector
+{
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public sealed class ServiceInfoCreatorAssemblyAttribute : AssemblyContainsTypeAttribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ServiceInfoCreatorAssemblyAttribute"/> class.
+        /// Used to override the default ServiceInfoCreator
+        /// </summary>
+        /// <param name="creatorType">The type of your info creator that descends from Steeltoe.Connector.ServiceInfoCreator</param>
```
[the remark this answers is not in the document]
TimHess: just the word my brain picked that day, `inherits` is fine

src/Connectors/src/Abstractions/Services/IServiceInfoFactory.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Steeltoe.Connector.Services
+{
+    public interface IServiceInfoFactory
+    {
+        /// <summary>
+        /// Check if this factory can create <see cref="IServiceInfo"/> from the given binding
+        /// </summary>
+        /// <param name="binding">A service binding to evaluate</param>
+        /// <returns>Gets a value indicating whether or not the binding is compatible with this factory</returns>
+        bool Accept(Service binding);
```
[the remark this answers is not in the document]
TimHess: yeah, I suppose - this looks like a new file, but it's not, it was moved from ConnectorBase

src/Connectors/src/ConnectorBase/ConnectionStringConfigurationProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+    {
+        public IConfiguration Configuration { get; private set; }
+
+        internal IList<IConfigurationProvider> _providers;
+
+        internal ConnectionStringManager ConnectionStringManager { get; set; }
+
+        internal ServiceInfoCreator ServiceInfoCreator { get; set; }
+
+        public ConnectionStringConfigurationProvider(IList<IConfigurationProvider> providers)
```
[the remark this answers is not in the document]
TimHess: this is a constructor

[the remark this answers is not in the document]
ccheetham: hah! duh.

src/Connectors/src/ConnectorBase/ConnectionStringConfigurationProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public IList<IConfigurationProvider> Providers => _providers;
+
+        public new IChangeToken GetReloadToken()
+        {
+            EnsureInitialized();
+            return Configuration.GetReloadToken();
+        }
+
+        public override void Load()
```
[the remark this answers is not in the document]
TimHess: I hear you, but `Load` is called by framework code and this is an internal class

src/Connectors/src/ConnectorBase/ConnectionStringConfigurationProvider.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            Configuration ??= new ConfigurationRoot(_providers);
+            ConnectionStringManager ??= new ConnectionStringManager(Configuration);
+        }
+
+        public override bool TryGet(string key, out string value)
+        {
+            if (key.StartsWith("ConnectionStrings"))
+            {
+                var searchKey = key.Split(':')[1];
```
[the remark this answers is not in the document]
TimHess: fair point, though I don't know if there's a realistic case for that happening - this config provider is only intended for use with `GetConnectionString()` [as seen in Method 1 here](https://www.connectionstrings.com/store-and-read-connection-string-in-appsettings-json/)

src/Connectors/src/ConnectorBase/DocumentDB/MongoDb/MongoDbConnectorOptions.cs
TimHess: Ignoring case is the important part, since we need to specify something I usually just go all the way to invariantcultureignorecase because we want to be pretty flexible about the matching

src/Connectors/src/ConnectorBase/DocumentDB/MongoDb/MongoDbHealthContributor.cs
TimHess: It could be a named constant, but I think making the *default* be externally configurable wouldn't be worth the effort... this was at least an improvement over having it inline and non-configurable

src/Connectors/src/ConnectorBase/ConnectionStringConfigurationProvider.cs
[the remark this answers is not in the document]
ccheetham: I hear ya.  So it'd have to be a config block like:

```
{
  "ConnectionStrings": {
  }
}
```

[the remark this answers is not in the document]
TimHess: That's the general expectation. I'll add some error handling anyway so we don't blow up if somebody does something unexpected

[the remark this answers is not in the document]
TimHess: Actually, to clarify - we don't expect anybody to use build that block when they're using this config provider, we're stuffing Steeltoe-built connection strings into a virtual config section that will emulate that behavior

src/Connectors/src/ConnectorBase/ServiceInfoCreator.cs
TimHess: ServiceInfoCreator is supposed to be a static singleton - there should only be one, retrieved via `Instance` below

ccheetham: Maybe cleaner to split this class into 2 classes then. One of the factory and one for the configuration.

TimHess: Is latest version better?

src/Connectors/src/CloudFoundry/CloudFoundryServiceInfoCreator.cs
```diff
@@ -12,50 +13,48 @@ namespace Steeltoe.Connector.CloudFoundry
...
         {
-            if (config == null)
+            if (configuration == null)
             {
-                throw new ArgumentNullException(nameof(config));
+                throw new ArgumentNullException(nameof(configuration));
             }
 
-            if (config == _config)
+            if (configuration == _me?.Configuration)
```
ccheetham, line 32: This section, from here to the final return can be simplified into a single return by converting the == into !=

src/Connectors/src/CloudFoundry/CloudFoundryServiceInfoCreator.cs
```diff
@@ -12,50 +13,48 @@ namespace Steeltoe.Connector.CloudFoundry
...
             {
                 return _me;
             }
 
             lock (_lock)
             {
-                if (config == _config)
+                if (configuration == _me?.Configuration)
```
ccheetham, line 39: I suspect that you only expect to call this static method once.  Or, if more than once, with the same logical configuration.  If so, if the compared configs are not equal, should an error be thrown?  Or a warning logged?

src/Connectors/src/CloudFoundry/CloudFoundryServiceInfoCreator.cs
```diff
@@ -12,50 +13,48 @@ namespace Steeltoe.Connector.CloudFoundry
     public class CloudFoundryServiceInfoCreator : ServiceInfoCreator
     {
         private static readonly object _lock = new object();
-        private static IConfiguration _config;
         private static CloudFoundryServiceInfoCreator _me = null;
 
-        internal CloudFoundryServiceInfoCreator(IConfiguration config)
-            : base(config)
+        internal CloudFoundryServiceInfoCreator(IConfiguration configuration)
```
ccheetham, line 18: Since a singleton, should this be `private`?

src/Connectors/src/CloudFoundry/CloudFoundryServiceInfoCreator.cs
TimHess: I know this is hit by the unit tests, its possibly it is here _only_ for testing purposes

src/Connectors/src/CloudFoundry/CloudFoundryServiceInfoCreator.cs
TimHess: I think this return exists to avoid the (presumably negligible) perf hit from the lock. I think this return makes sense but inverting the logic below is also good

ccheetham: To be clear, you still need to the `if` checks.  

I was suggesting:
```
if (configuration != _me?.Configuration)
{
    lock (_lock)
    {
        if (configuration != _me?.Configuration)
        {
            _me = new CloudFoundryServiceInfoCreator(configuration);
        }
    }
}
return _me;
```
From performance perpspective, 2 blocks identical.

src/Connectors/src/CloudFoundry/CloudFoundryServiceInfoCreator.cs
```diff
@@ -12,50 +13,46 @@ namespace Steeltoe.Connector.CloudFoundry
...
-            if (config == null)
+            if (configuration == null)
             {
-                throw new ArgumentNullException(nameof(config));
+                throw new ArgumentNullException(nameof(configuration));
             }
 
-            if (config == _config)
+            if (configuration == _me?.Configuration)
             {
```
ccheetham, line 33: This branch of the `if`, that is the one returning `_me`,  is unnecessary.

src/Connectors/src/ConnectorBase/ConnectionStringConfigurationSource.cs
```diff
@@ -0,0 +1,29 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Configuration;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Steeltoe.Connector
+{
+    public class ConnectionStringConfigurationSource : IConfigurationSource
```
TimHess, line 12: The ConnectionString Configuration source/provider _could_ be in a separate package, grouped under the other `/Configuration` packages, should it move to a dedicated package? I put it here because it only exists as an entry point to the logic in ConnectorBase

src/Common/src/Common/Reflection/ReflectionHelpers.cs
```diff
@@ -30,6 +33,87 @@ public static Assembly FindAssembly(string name)
             return null;
         }
 
+        /// <summary>
```
TimHess, line 36: @dtillman There are some helpers in here that might also be interesting in relation to some of the streams work

src/Connectors/src/ConnectorBase/ServiceInfoCreator.cs
ccheetham: yep

PR #393, opened 2020-08-05
Rename Discovery Client/Extension Packages as they're no longer the base

TimHess:


src/Discovery/src/ClientBase/DiscoveryHostBuilderExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        /// <exception cref="ConnectorException">Thrown if no service info with expected name or type are found or when multiple service infos are found and a single was expected</exception>
+        public static IHostBuilder AddServiceDiscovery(this IHostBuilder hostBuilder, Action<DiscoveryClientBuilder> optionsAction)
+        {
+            return hostBuilder.ConfigureServices((context, collection) => AddServices(collection, optionsAction));
+        }
+
+        private static void AddServices(IServiceCollection collection, Action<DiscoveryClientBuilder> optionsAction)
+        {
+            collection.AddServiceDiscovery(optionsAction);
+            collection.AddHostedService(services => new DiscoveryClientService(services.GetRequiredService<IDiscoveryLifecycle>()));
```
TimHess, line 32: I'm not a huge fan of how this is working... We don't currently need `DiscoveryClientService` for anything other than a mechanism to make sure `IDiscoveryLifeCycle` gets loaded up

Issue #407, opened 2020-08-12
DiscoveryClient assembly not always loaded


TimHess:
When using the reflection-based approach to load Discovery Client assemblies, assemblies that have not been referenced in code won't be found. Same root cause as #401

PR #408, opened 2020-08-12
Fix discovery client assemblies not found with reflection approach

TimHess:
addresses #407 and ensures parity between (Web)HostBuilder discovery extensions

### DIFF
--- a/src/Discovery/src/ClientBase/DiscoveryClientAssemblyAttribute.cs
+++ b/src/Discovery/src/ClientBase/DiscoveryClientAssemblyAttribute.cs
@@ -16,7 +16,7 @@ namespace Steeltoe.Discovery
     {
         /// <summary>
         /// Initializes a new instance of the <see cref="DiscoveryClientAssemblyAttribute"/> class.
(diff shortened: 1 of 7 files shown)


src/Discovery/src/ClientBase/DiscoveryClientAssemblyAttribute.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Identify assemblies containing ServiceInfoCreators
+    /// </summary>
+    [AttributeUsage(AttributeTargets.Assembly)]
+    public sealed class DiscoveryClientAssemblyAttribute : AssemblyContainsTypeAttribute
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DiscoveryClientAssemblyAttribute"/> class.
+        /// Used to override the default ServiceInfoCreator
```
TimHess, line 19:
```suggestion
        /// Used to identify assemblies that contain a discovery client
```

PR #412, opened 2020-08-17
Re-use Httpclients

TimHess:
Bring #409 over to 3.0

### DIFF
--- a/src/Common/src/Common.Http/HttpClientHelper.cs
+++ b/src/Common/src/Common.Http/HttpClientHelper.cs
@@ -67,6 +67,14 @@ namespace Steeltoe.Common.Http
             return client;
         }
 
+        public static HttpClient GetHttpClient(HttpMessageHandler handler, int timeout = 15)
+        {
+            var client = handler == null ? new HttpClient() : new HttpClient(handler);
+            client.Timeout = TimeSpan.FromMilliseconds(timeout);
+            client.DefaultRequestHeaders.UserAgent.ParseAdd(SteeltoeUserAgent);
+            return client;
+        }
+
         public static void ConfigureCertificateValidation(
             bool validateCertificates,
             out SecurityProtocolType protocolType,
@@ -203,7 +211,7 @@ namespace Steeltoe.Common.Http
                 throw new ArgumentException("Access token Uri is not well formed", nameof(accessTokenUri));
             }
 
-            return GetAccessTokenInternal(accessTokenUri, clientId, clientSecret, timeout, validateCertificates, logger, additionalParams, httpClient);
+            return GetAccessTokenInternal(accessTokenUri, clientId, clientSecret, timeout, validateCertificates, httpClient, additionalParams, logger);
         }
 
         private static async Task<string> GetAccessTokenInternal(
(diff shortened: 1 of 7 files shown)


src/Common/src/Common.Http/HttpClientHelper.cs
```diff
@@ -67,6 +67,14 @@ public static HttpClient GetHttpClient(bool validateCertificates, HttpClientHand
             return client;
         }
 
+        public static HttpClient GetHttpClient(HttpMessageHandler handler, int timeout = 15)
```
ccheetham, line 70: In lieu of XML docs ...

```
, int timeoutMillis = 15)
```

src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
```diff
@@ -834,7 +828,6 @@ protected internal async void RefreshVaultTokenAsync(object state)
             }
             finally
             {
-                client.Dispose();
```
ccheetham, line 837: Does this client get disposed elsewhere? I see a lot of dispose removal (and the comment about "after token") but am missing where they ultimately get cleaned.

TimHess: With this change, the client lives until the app shuts down

src/Common/src/Common.Http/HttpClientHelper.cs
TimHess: oh, wow... this should probably not be 15 ms, that seems a bit short for an http timeout!

TimHess:
I'm not sure if [this post](https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/) is the origin of this conversation, but it is why this issue has been floating around my brain for a while... Long story short, [the guidance](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netcore-3.1#remarks) is that we'd be better off with singleton `HttpClient`s than a new one for every request 

> I didn't look through details of this change .. but my first thought was have you double checked/verified a couple things:
> 
> 1. Not disposing of HttpClient will likely leave the connection open to whatever the other end is (e.g. config server, eureka server, etc.)  and will that cause any issues for our users.
> 2. Also, if the other end fails (i.e. stops .. doesn't restart, or stops and restarts) does the affected Steeltoe code properly handle those failure cases (i.e. cycles to next URI if the Steeltoe client supports that, etc.).  Did you test those failure cases for the various Steeltoe components to verify all works properly?
> 3. HttpClients have state so. .. are there any multithreading issues associated with your changes... e.g. two threads using the same HttpClient concurrently, etc.
> 4. Some of the Steeltoe components do a OAuth flow to get a token using HttpClient.. does this all still work properly with this change... may not even be an issue as I don't remember how that is all working today

1. HttpClient does its own connection pool management, but my understanding is that it will initiate a new connection as-needed and should not fail if the tcp connection goes away between requests. The flip side is where we are now, with .Dispose not always closing the connections immediately or predictably 
2. ConfigServer & Eureka code have retry built-in that should handle anything that comes up as a result of this change, but I haven't done any extensive testing
3. Everything we're doing should be thread safe per [the docs](https://docs.microsoft.com/en-us/dotnet/api/system.net.http.httpclient?view=netcore-3.1#remarks)
4. I don't think this change will break anything there

dtillman:
Ok .. if you're confident in the change.. I'll let @hananiel  and @ccheetham finish their reviews and approve

Issue #372, opened 2020-06-30
Enhance Serilog Integration with Management


TimHess:
Enhance the Serilog integration so it is as easy/seamless to use as the dynamic console logger (eg: can Serilog be used in combination with `HostBuilder.AddCloudFoundryActuators`?)

PR #414, opened 2020-08-17
Management and Dynamic logging enhancements

TimHess:
Better (more definitive) handling of unsupported dynamic logging configurations (should help with https://github.com/SteeltoeOSS/Samples/issues/163)
Also improves consistency with dynamic logger setup methods

Doesn't provide a strong linking, but might be considered a resolution for #372  (the auto-adding of DynamicConsoleLogger will now just silently fail if DynamicSerilog is previously configured, adding Serilog after DynamicConsole throws at run-time)


src/Logging/src/DynamicLogger/DynamicLoggingBuilder.cs
```diff
@@ -27,24 +27,29 @@ public static ILoggingBuilder AddDynamicConsole(this ILoggingBuilder builder)
...
-            // make sure logger provider configurations are available
-            if (!builder.Services.Any(descriptor => descriptor.ServiceType == typeof(ILoggerProviderConfiguration<ConsoleLoggerProvider>)))
-            {
-                builder.AddConfiguration();
+                // make sure logger provider configurations are available
+                if (!builder.Services.Any(descriptor => descriptor.ServiceType == typeof(ILoggerProviderConfiguration<ConsoleLoggerProvider>)))
+                {
+                    builder.AddConfiguration();
+                }
+
```
ccheetham, line 45: maybe a high-level comment here of what's being done

src/Logging/src/DynamicSerilogBase/SerilogHostBuilderExtensions.cs
```diff
@@ -0,0 +1,55 @@
...
+    public static class SerilogHostBuilderExtensions
+    {
+        /// <summary>
+        /// Configure Serilog as the <see cref="IDynamicLoggerProvider"/> to enable dynamically controlling log levels via management endpoints
+        /// </summary>
+        /// <param name="hostBuilder">The <see cref="IHostBuilder"/> to configure</param>
+        /// <param name="configureLogger">The delegate for configuring the <see cref="DynamicLoggerConfiguration" /> that will be used to construct a <see cref="Serilog.Core.Logger" /></param>
+        /// <param name="preserveStaticLogger">Indicates whether to preserve the value of <see cref="Log.Logger"/>.</param>
+        /// <returns>The <see cref="IHostBuilder"/></returns>
+        public static IHostBuilder AddDynamicSerilog(
```
ccheetham, line 20: Perhaps intended ... one method name (below) has `Console` and one doesn't (this)

src/Logging/src/DynamicSerilogBase/SerilogLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+
+namespace Steeltoe.Extensions.Logging.DynamicSerilog
+{
+    public static class SerilogLoggingBuilderExtensions
+    {
+        /// <summary>
+        /// Add Steeltoe logger wrapped in a <see cref="IDynamicLoggerProvider"/> that supports
+        /// dynamically controlling the minimum log level via management endpoints
+        /// </summary>
+        /// <param name="builder">The <see cref="ILoggingBuilder"/> for configuring the LoggerFactory  </param>
```
ccheetham, line 22: extra whitespace in comment tail

src/Logging/src/DynamicSerilogBase/SerilogLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        public static ILoggingBuilder AddSerilogDynamicConsole(this ILoggingBuilder builder)
+        {
+            return builder.AddSerilogDynamicConsole(null, false);
+        }
+
+        /// <summary>
+        /// Add Steeltoe logger wrapped in a <see cref="IDynamicLoggerProvider"/> that supports
+        /// dynamically controlling the minimum log level via management endpoints
+        /// </summary>
+        /// <param name="builder">The <see cref="ILoggingBuilder"/> for configuring the LoggerFactory  </param>
```
ccheetham, line 33: extra whitespace

src/Logging/src/DynamicSerilogBase/SerilogLoggingBuilderExtensions.cs
```diff
@@ -0,0 +1,64 @@
...
+        /// <param name="loggerConfiguration">An initial <see cref="LoggerConfiguration"/></param>
+        /// <param name="preserveStaticLogger">Indicates whether to preserve the value of <see cref="Log.Logger"/>.</param>
+        /// <returns>The configured <see cref="ILoggingBuilder"/></returns>
+        public static ILoggingBuilder AddSerilogDynamicConsole(this ILoggingBuilder builder, LoggerConfiguration loggerConfiguration, bool preserveStaticLogger = false)
+        {
+            if (builder.Services.Any(sd => sd.ServiceType == typeof(IDynamicLoggerProvider)))
+            {
+                throw new InvalidOperationException("An IDynamicLoggerProvider has already been configured! Call 'AddSerilogDynamicConsole' earlier in program.cs (before adding Actuators) or remove duplicate IDynamicLoggerProvider entries.");
```
ccheetham, line 41: (deja vu?) The exclamation point seems redundant.

src/Management/src/EndpointCore/ManagementWebHostBuilderExtensions.cs
```diff
@@ -0,0 +1,278 @@
...
+using System;
+
+namespace Steeltoe.Management.Endpoint
+{
+    public static class ManagementWebHostBuilderExtensions
+    {
+        /// <summary>
+        /// Adds the Database Migrations actuator to the application
+        /// </summary>
+        /// <param name="hostBuilder">Your HostBuilder</param>
```
ccheetham, line 33: Haven't looked at other comments ... "Your" seems informal.  Is that what similar comments in code base use?

src/Logging/src/DynamicSerilogBase/SerilogHostBuilderExtensions.cs
TimHess: I could be wrong, but I thought this integration wasn't actually exclusively for console logging, so that's why I dropped it... maybe @hananiel can weigh in here to clarify ~ I would agree that it should be there if that is the case, and it should probably be made consistent across the board in this PR

src/Logging/src/DynamicSerilogBase/SerilogLoggingBuilderExtensions.cs
TimHess: This isn't new code, it just looks like it because the file rename shows up as a delete/add... this?
```suggestion
                throw new InvalidOperationException("An IDynamicLoggerProvider has already been configured. Call 'AddSerilogDynamicConsole' earlier in program.cs (before adding Actuators) or remove duplicate IDynamicLoggerProvider entries.");
```

src/Management/src/EndpointCore/ManagementWebHostBuilderExtensions.cs
TimHess: Contents of this file were copied from `ManagementHostBuilderExtensions` so it is done this way at least there, probably elsewhere as well... there's not a lot to say here... we could do `<see cref="IWebHostBuilder" />` instead

src/Management/src/CloudFoundryCore/CloudFoundryServiceCollectionExtensions.cs
```diff
@@ -19,7 +19,7 @@ public static class CloudFoundryServiceCollectionExtensions
         /// <param name="services">Service collection</param>
         /// <param name="config">Application Configuration</param>
```
TimHess, line 20: @ccheetham something like this?
```suggestion
        /// <param name="config">Application Configuration. Retrieved from the <see cref="IServiceCollection" /> if not provided</param>
```

src/Logging/src/DynamicSerilogBase/SerilogHostBuilderExtensions.cs
hananiel: @TimHess that is right - it isn't for exclusively for console logging. It allows you to have a variety of sinks.

src/Logging/src/DynamicSerilogBase/Steeltoe.Extensions.Logging.DynamicSerilogBase.csproj
```diff
@@ -1,17 +1,17 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <TargetFrameworks>netstandard2.0</TargetFrameworks>
-    <Description>Steeltoe library for enabling dynamic management of Serilog</Description>
-    <PackageTags>logging;dynamic logging;serilog;aspnetcore;management;monitoring;Spring Cloud;</PackageTags>
+    <Description>Steeltoe library for enabling dynamic management of Serilog. Includes Console sink.</Description>
```
TimHess, line 4: Added note about the Console sink to the description... If we remove the reference to Serilog.Sinks.Console, the packages still built but 3 tests fail... I could have moved the reference to the tests, but I _think_ console logging is used in most/all Steeltoe use cases, so it probably makes sense to leave it here and just call it out

TimHess:
> A couple general comments.
> 
> 1. The doc comments have inconsistent ending punctuation. Some end in periods, some do not.  I believe the recommended practice is to end all doc comments (summary, params, etc) with peried.
> 2. The Add*Actuator comments do not desribed the behavior of `null` values.  If the behavior were same for all params, that might be OK.  But now there's additional behavior for `config` params that are `null`.

1. Good point, but possibly addressed separately in a more comprehensive fashion?
2. `        /// <param name="config">Application configuration. Retrieved from the <see cref="IServiceCollection"/> if not provided</param>` ?

src/Management/src/EndpointCore/ManagementWebHostBuilderExtensions.cs
ccheetham: Best left for future left then

src/Logging/src/DynamicSerilogBase/SerilogLoggingBuilderExtensions.cs
ccheetham: Looks good

PR #434, opened 2020-09-11
Minor Eureka enhancements and Fix ApplicationInstanceInfo binding

TimHess:
For Eureka:
- Better handling of empty `OK` response from heartbeat request
- Allow use of app addresses from "urls" in configuration
- Use secure port in app id if enabled

Port-related issues in Tye (Env variable `PORT` might be a semicolon delimited string) turned out to be related to bare/default `ApplicationInstanceInfo`  binding to the root of configuration instead of a base path. The last commit in this PR adds a new constructor that will result in binding `ApplicationInstanceInfo` to keys under `application` instead of the root

### DIFF
--- a/src/Common/src/Common.Kubernetes/KubernetesApplicationOptions.cs
+++ b/src/Common/src/Common.Kubernetes/KubernetesApplicationOptions.cs
@@ -30,8 +30,6 @@ namespace Steeltoe.Common.Kubernetes
 
         public bool Enabled { get; set; } = true;
 
-        public string Name { get; set; }
-
         public override string ApplicationName => Name;
 
         public string NameSpace { get; set; } = "default";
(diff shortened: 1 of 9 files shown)


cecilphillip:
I've been staring at this code for hours trying to figure out why I was getting `System.InvalidOperationException: Failed to convert configuration value at 'Port' to type 'System.Int32'.` with Tye. Guess I'll wait til this fix gets pushed.

ddieruf:
@cecilphillip this was sparked from using tye in this proj - https://github.com/spboyer/steeltoe-sample.

TimHess:
@cecilphillip sorry about that, it made a pretty crappy experience. A short-term workaround to use Steeltoe in Tye is to specify that only http or https should be used so that you don't get two ports stuffed into a variable where only one port was previously expected:
```
services:
- name: musicstoreservice
  project: src/MusicStoreService/MusicStoreService.csproj
  bindings:
    - protocol: http
```

cecilphillip:
@ddieruf nice. I have a sample [here](https://github.com/cecilphillip/ContosoCrafts/tree/steeltoe) that I'm trying to add Tye support to.

@TimHess Thanks! I'll try that binding setting

dtillman (approved):

Issue #418, opened 2020-08-20
Handle passthrough of error messages on IIS


jkonicki:
## Problem Summary:
* There was a change to Steeltoe quite some time ago to comply with the Spring spec for returning HTTP 503 in cases of HealthStatus.DOWN or HealthStatus.OUT_OF_SERVICE. 

* For 500 series errors, however, HWC is configured by default to substitute a canned message for the http response content - so, instead of getting back any failed health checks, the monitor just gets a 503 with “Service Not Available” instead of the healthcheck JSON. 

* Thus, in the default case, health monitoring with HWC and Steeltoe no longer works. People who were using AppsManager and were able to see rich information in their health checks might not be able to see this information anymore, and will likely not know the cause. 

## Exploration:
We validated in our exploration that HWC’s setting for 500 errors can be changed via a Web.config file in an application. 
```
<?xml version="1.0" encoding="utf-8"?>
<configuration>
    <system.webServer>
        <httpErrors existingResponse="PassThrough"/>
    </system.webServer>
</configuration>
```
This would be achieved by setting httpErrors existingResponse to “PassThrough”.

We explored changing this setting globally in HWC, but this would change the default error page behavior of every existing app using default settings for HWC. We decided this was a risk.

The defaults we have right now in HWC match the IIS defaults, so we’re more inclined to keep these default settings as they are.

Note: The Web API will return the correct HTTP response code regardless of this setting. It’s just the content of the API response that is different. 

## Possible Resolution:
* Add a note to the Steeltoe documentation about this, warning users of this issue and noting that a customer can change this setting in their application Web.config file.  A further solution would be for Steeltoe to check for this setting programmatically and fail with a message to correct this in the Web.config file.
* We could change add a property like so:
``` 
if (!config["management:endpoints:health:alwaysOK"])
{
    context.Response.StatusCode = ((HealthEndpoint)_endpoint).GetStatusCode(result);
} 
```
This would be a new config switch we'd add that would allow you to override our behavior of setting the status code so that it always returns 200 or at least the actual health response wouldn't affect what code we return... there could still be other unhandled exceptions making it not a 200.
This could be acceptable since the health response is generally more important than the status code.

Note: This same issue would exist for any K8s users using Microsoft dotnet/framework/aspnet images and apps with Steeltoe.

emgrig:
[...]

PR #439, opened 2020-09-14
Allow disabling health status as HTTP status

TimHess:
Add a setting that can disable use of health status for HTTP status so that IIS doesn't scrub data in the response body #418

### DIFF
--- a/src/Management/src/EndpointCore/Health/HealthEndpointMiddleware.cs
+++ b/src/Management/src/EndpointCore/Health/HealthEndpointMiddleware.cs
@@ -60,7 +60,8 @@ namespace Steeltoe.Management.Endpoint.Health
         protected internal string DoRequest(HttpContext context)
         {
             var result = _endpoint.Invoke(new CoreSecurityContext(context));
-            if (((HealthEndpointOptions)_mgmtOptions.FirstOrDefault().EndpointOptions.FirstOrDefault(o => o is HealthEndpointOptions)).HttpStatusFromHealth)
+            var managementOptions = _mgmtOptions.OptionsForContext(context.Request.Path, _logger);
+            if (((HealthEndpointOptions)managementOptions.EndpointOptions.FirstOrDefault(o => o is HealthEndpointOptions)).HttpStatusFromHealth)
             {
                 context.Response.StatusCode = ((HealthEndpoint)_endpoint).GetStatusCode(result);
             }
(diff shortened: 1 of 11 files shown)


src/Management/src/EndpointCore/Health/HealthEndpointMiddleware.cs
```diff
@@ -59,7 +60,11 @@ protected internal async Task HandleHealthRequestAsync(HttpContext context)
         protected internal string DoRequest(HttpContext context)
         {
             var result = _endpoint.Invoke(new CoreSecurityContext(context));
-            context.Response.StatusCode = ((HealthEndpoint)_endpoint).GetStatusCode(result);
+            if (((HealthEndpointOptions)_mgmtOptions.FirstOrDefault().EndpointOptions.FirstOrDefault(o => o is HealthEndpointOptions)).HttpStatusFromHealth)
```
hananiel, line 63: This does not look correct to me.  I am thinking either filter by type for ActuatorManagementOptions since you don't care about Cloud Foundry path (on `/cloudfoundryapplication`)  or find the matching management options like [here]( https://github.com/SteeltoeOSS/Steeltoe/blob/f2b82898a0f29a3a69d9128b09a33e9299ea4a1b/src/Management/src/EndpointBase/EndpointExtensions.cs#L73). Depending on the setup, the two collections of  EndpointOptions on IManagementOptions can be different and have weird behavior.

src/Management/src/EndpointOwin/Health/HealthEndpointOwinMiddleware.cs
```diff
@@ -40,7 +41,12 @@ public override async Task Invoke(IOwinContext context)
                 _logger?.LogTrace("Processing {SteeltoeEndpoint} request", typeof(HealthEndpoint));
                 var result = _endpoint.Invoke(new OwinSecurityContext(context));
                 context.Response.Headers.SetValues("Content-Type", new string[] { "application/vnd.spring-boot.actuator.v2+json;charset-UTF-8" });
-                context.Response.StatusCode = ((HealthEndpoint)_endpoint).GetStatusCode(result);
+
+                if (((HealthEndpointOptions)_mgmtOptions.FirstOrDefault().EndpointOptions.FirstOrDefault(o => o is HealthEndpointOptions)).HttpStatusFromHealth)
```
hananiel, line 45: Same comment as above.

hananiel (changes requested):
Please see the comments.

src/Management/src/EndpointCore/Health/HealthEndpointMiddleware.cs
TimHess: The only property we care about here is this new `HttpStatusFromHealth`... Could `HealthEndpointOptions` have different values for this property between ActuatorManagementOptions entries and/or is it even possible for these options to be present in only one of the two places?

FWIW, this work was specifically requested for Cloud Foundry use cases

hananiel (approved):

Issue #443, opened 2020-09-16
Backport Eureka port and host fixes from #434


jkonicki:
Backport port and host handling for Eureka.

Issue #442, opened 2020-09-16
Consul handling of registration of port and host 


jkonicki:
Allow auto-registration of host and port for Consul with ASP.NET Core configured urls.

Should be a faillback scenario is any configuration is specified.

See Eureka changes #434

PR #445, opened 2020-09-24
allow registration using ASP.NET Core urls

TimHess:
Allow (as a default) use of the urls ASP.NET Core is listening on when registering with a service registry #443 and #442

Defaults to HTTPS when present

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -92,6 +92,7 @@ jobs:
       arguments: '--blame --no-build -c $(buildConfiguration) -maxcpucount:1 /p:CopyLocalLockFileAssemblies=true $(skipFilter) --collect:"XPlat Code Coverage" --settings coverlet.runsettings --logger trx --results-directory $(Build.SourcesDirectory)'
       publishTestResults: false
   - task: PublishTestResults@2
+    condition: always()
     inputs:
       testResultsFormat: VSTest
       testResultsFiles: '*.trx'
--- a/src/Discovery/src/EurekaBase/EurekaInstanceOptions.cs
+++ b/src/Discovery/src/EurekaBase/EurekaInstanceOptions.cs
@@ -2,7 +2,11 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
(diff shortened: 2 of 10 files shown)


azure-pipelines.yml
```diff
@@ -92,6 +92,7 @@ jobs:
       arguments: '--blame --no-build -c $(buildConfiguration) -maxcpucount:1 /p:CopyLocalLockFileAssemblies=true $(skipFilter) --collect:"XPlat Code Coverage" --settings coverlet.runsettings --logger trx --results-directory $(Build.SourcesDirectory)'
       publishTestResults: false
   - task: PublishTestResults@2
+    condition: always()
```
hananiel, line 95: Did you mean to change this here?

hananiel (dismissed):

azure-pipelines.yml
TimHess: yeah, I had a random hystrix test fail, but because `dotnet test` failed this step didn't run and that failure wasn't published to the build results...  Previously the test step always auto-published the results so we didn't even have a discrete step here, but now we need to make sure it always runs so the results always show up in the report

src/Discovery/src/EurekaBase/EurekaInstanceOptions.cs
```diff
@@ -162,5 +136,54 @@ public override string GetHostName(bool refresh)
...
+                var springInstanceId = config.GetValue<string>(EurekaPostConfigurer.SPRING_APPLICATION_INSTANCEID_KEY);
+                if (!string.IsNullOrEmpty(springInstanceId))
+                {
+                    InstanceId = springInstanceId;
+                }
+                else
+                {
+                    if (SecurePortEnabled)
+                    {
+                        InstanceId = GetHostName(false) + ":" + AppName + ":" + SecurePort;
```
TimHess, line 179: @hananiel this line actually _should_ be hit by the same test that hits the block under line 152 above... should we set SecurePortEnabled=true when using urls from config that use https? Consul only supports a single port and this PR sets prefers https, so maybe we should add something like that here, what do you think?

hananiel (approved):
LGTM

PR #449, opened 2020-09-25
Add InitializeAsync()

pengweiqhca:


TimHess:
Thanks, this change does need to make it to 3.0, but rather than merge this PR (and end up with two commit hashes in history for the same change) I'm going to include #448 in the next merge-forward from 2.x, which should be in the next few days

PR #450, opened 2020-09-25
Updates from 2x

TimHess:
Bring pull requests from 2.x over to master


src/Management/src/EndpointBase/EndpointExtensions.cs
```diff
@@ -68,6 +74,35 @@ public static bool ShouldInvoke(this IEndpoint endpoint, IManagementOptions mgmt
             return enabled && exposed;
         }
 
+        public static IManagementOptions OptionsForContext(this IEnumerable<IManagementOptions> mgmtOptions, string requestPath, ILogger logger = null)
```
TimHess, line 77: @hananiel this doesn't seem to be currently needed in 3.0, there's no references to it and I hadn't added any explicit tests for it... should we drop it, or do you see a use for it somewhere?

hananiel: I am not seeing a need for it. We can drop it.

hananiel (approved):
I am ok with leaving OptionsForContext and removing later.

Issue #131, opened 2019-10-07
Eureka endpoint - online/offline service instances 


jkonicki:
Provide Blue/Green deployments through endpoint. Use the endpoint to set registered instances to offline

TimHess:
This functionality is provided in Spring via the `service-registry` endpoint: https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/serviceregistry/endpoint/ServiceRegistryEndpoint.java

TimHess:
Sounds good, have at it!

jkonicki:
@cesinhaugusto We are happy to have the help.  Let us know if there are any questions or if you have any issues getting started on this task.  Thanks.

cesinhaugusto:
@jkonicki Good to know that, because I'll probably need it a lot hahaha. I took some time to study Steeltoe a little bit more and get to know the project better in order to be able to make meaningful contributions.

In this specific issue, I'd like to ask you guys if the goal is to just create an Endpoint class, similar to the Spring one that @TimHess mentioned above, and add it to the EurekaBase library.

Initially, I thought that the goal was to prepare Steeltoe to completely handle Blue/Green deployments, but after further research, I understood this is responsibility of the platform, not the library. Am I correct?

TimHess:
Yeah, this is basically "make blue/green relatively easy or at least straightforward" but not like a 360 degree solution, so some form of platform or external tooling support would be expected.

I didn't get extremely far in exploratory work here, but I was expecting the bulk of the work to be done in Management in the form of a service-registry actuator... Ideally there would be little to no work required in either of the Discovery client libraries, but I'm not positive that's the case. `IServiceRegistry` has the methods needed, but it looks like it's only used by Consul at the moment

cesinhaugusto:
I see. This is a little more complex than I anticipated then. I'll learn more about Management and try to come up with a draft for this and return to get your opinion/suggestions.

bart-vmware:
The concept of blue/green deployments is explained at: https://martinfowler.com/bliki/BlueGreenDeployment.html.

PR #461, opened 2020-10-13
Allow eureka repeat register and unregister

pengweiqhca:


TimHess:
@pengweiqhca is this so that you can take an app (instance) offline or online on-demand, perhaps like what #131 is describing?

pengweiqhca:
> @pengweiqhca is this so that you can take an app (instance) offline or online on-demand, perhaps like what #131 is describing?

Yes, but we has implemented. 
``` C#
public class TuhuEurekaDiscoveryClient : EurekaDiscoveryClient
{
	public Task StartupAsync()
	{
		if (_shutdown == 0) return Task.CompletedTask;

		_appInfoManager.InstanceStatus = InstanceStatus.UP;

		return InitializeAsync().ContinueWith(_ => _shutdown = 0);
	}
}
```
I want make `StartupAsync()` is simpler. So, have this PR

TimHess:
I did [some preliminary work](https://github.com/SteeltoeOSS/Steeltoe/compare/registry_endpoint) in this area already - would it be cleaner to discretely change the status, or are you really wanting to start and stop the client?

pengweiqhca:
Yes, if don't stop, _heartBeatTimer will modify the status of the client

TimHess:
@pengweiqhca does it look like the work I started will work for you? I didn't have a chance to work on it today, but the intent is that heartbeats will continue to fire (but without changing the status back to up)

PR #469, opened 2020-10-15
Fix Loggers, Metrics and Hypermedia when Manangement:Endpoint:Path is…

### LINKED ISSUE #467: POST to loggers actuator fails if the base management path is /
## Describe the bug
When posting to the loggers endpoint when using the dynamic logger to change the log level, the request fails with a 400 bad request if the management endpoints path is `/`. 

## Steps to reproduce 
1. Create a simple AspNet Core app and add Steeltoe Dynamic logging and Management Endpoints.
2. Configure the management endpoints:

```
  "Management": {
    "Endpoints": {
      "Path": "/",
      "Actuator": {
        "Exposure": {
          "Include": [ "info", "health", "loggers" ]
        }
      }
    }
  }
```
3. Attempt to change the level:  POST https://localhost:5000/loggers/Microsoft

## Expected behavior
The request to return a 200 OK and the level should be changed.

 - Platform: Local
 - OS: Windows 10
 - .NET Version: Core 3.1
 - Steeltoe Version 3.0 (also fails in 2.x line)

hananiel:
… set to /

Resolves #467

### DIFF
--- a/src/Management/src/EndpointBase/Hypermedia/HypermediaService.cs
+++ b/src/Management/src/EndpointBase/Hypermedia/HypermediaService.cs
@@ -49,7 +49,8 @@ namespace Steeltoe.Management.Endpoint.Hypermedia
                     {
                         if (!links._links.ContainsKey(opt.Id))
                         {
-                            links._links.Add(opt.Id, new Link(baseUrl + "/" + opt.Path));
+                            var linkPath = $"{baseUrl.TrimEnd('/')}/{opt.Path}";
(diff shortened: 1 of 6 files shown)


TimHess:
We should merge #470 into 2.x and then bring it forward so the commits aren't in history twice

hananiel:
> We should merge #470 into 2.x and then bring it forward in #468 so the commits aren't (eventually) in history twice

Closing this in favor of #468

PR #470, opened 2020-10-15
Fix Loggers, Metrics and Hypermedia when Manangement:Endpoint:Path is…

### LINKED ISSUE #467: POST to loggers actuator fails if the base management path is /

hananiel:
… set to / 
resolves #467

### DIFF
--- a/src/Management/src/EndpointBase/Hypermedia/HypermediaService.cs
+++ b/src/Management/src/EndpointBase/Hypermedia/HypermediaService.cs
@@ -49,7 +49,8 @@ namespace Steeltoe.Management.Endpoint.Hypermedia
                 {
                     if (!string.IsNullOrEmpty(opt.Id) && !links._links.ContainsKey(opt.Id))
                     {
-                        links._links.Add(opt.Id, new Link(baseUrl + "/" + opt.Path));
+                        var linkPath = $"{baseUrl.TrimEnd('/')}/{opt.Path}";
+                        links._links.Add(opt.Id, new Link(linkPath));
                     }
                     else if (links._links.ContainsKey(opt.Id))
                     {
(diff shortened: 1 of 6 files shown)


src/Management/src/EndpointBase/Hypermedia/HypermediaService.cs
```diff
@@ -49,7 +49,8 @@ public Links Invoke(string baseUrl)
                 {
                     if (!string.IsNullOrEmpty(opt.Id) && !links._links.ContainsKey(opt.Id))
                     {
-                        links._links.Add(opt.Id, new Link(baseUrl + "/" + opt.Path));
+                        var linkPath = $"{baseUrl.TrimEnd('/')}/{opt.Path}";
```
TimHess, line 52:
```suggestion
                        var linkPath = $"{baseUrl}/{opt.Path}".Replace("//", "/");
```
For consistency with the other two changes

TimHess (approved):

src/Management/src/EndpointBase/Hypermedia/HypermediaService.cs
TimHess: ^ won't work b/c baseUrl is expected to include `://`

hananiel: `baseUrl` has Http(s?):// in it, so this change won't work here.  In the other place,  it seemed more efficient to do that since no replace occurs if `//` is not found.

PR #508, opened 2020-11-11
update config server image name for integration tests

TimHess:
`steeltoeoss/configserver` was replaced by `steeltoeoss/config-server`, this PR will update the integration tests

```
docker: Error response from daemon: pull access denied for steeltoeoss/configserver, repository does not exist or may require 'docker login': denied: requested access to the resource is denied.
```


ccheetham (approved):
lgtm

PR #511, opened 2020-11-13
Logging bug fix

### LINKED ISSUE #502: Dynamic Log filtering doesn't appear to be working
> Question about Log filtering, I have ILogger setup and it's outputs to the dynamic console, and "Configure Logging Levels" showing, when you output from controller how do you filter on Error and critical only ?  because I changed as it show but I am still getting the info messages.
<sub>[View in Slack](https://vmware.slack.com/archives/C1VKUERFH/p1604681075122400)</sub>

hananiel:
Resolves #502

### DIFF
--- a/src/Logging/src/SerilogDynamicLoggerCore/SerilogBuilderExtensions.cs
+++ b/src/Logging/src/SerilogDynamicLoggerCore/SerilogBuilderExtensions.cs
@@ -11,6 +11,7 @@ using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Configuration;
 using Microsoft.Extensions.Logging.Console;
 using Microsoft.Extensions.Options;
+using Serilog;
 using Serilog.Core;
 using System;
 using System.Linq;
@@ -66,7 +67,7 @@ namespace Steeltoe.Extensions.Logging.SerilogDynamicLogger
 
                 // Add a level switch that controls the "Default" level at the root
                 var levelSwitch = new LoggingLevelSwitch(serilogOptions.MinimumLevel.Default);
-                loggerConfiguration.MinimumLevel.ControlledBy(levelSwitch);
+                loggerConfiguration.MinimumLevel.ControlledBy(levelSwitch).WriteTo.Console(); // If there is no Console in the Configuration, or there is Configuration, go ahead and add console
 
                 var logger = loggerConfiguration.CreateLogger();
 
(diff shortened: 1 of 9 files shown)


src/Logging/src/SerilogDynamicLoggerCore/SerilogBuilderExtensions.cs
```diff
@@ -32,6 +33,12 @@ public static ILoggingBuilder AddSerilogDynamicConsole(this ILoggingBuilder buil
                 throw new InvalidOperationException("An IDynamicLoggerProvider has already been configured! Call 'AddSerilogDynamicConsole' earlier in program.cs (Before AddCloudFoundryActuators()) or remove duplicate IDynamicLoggerProvider entries.");
             }
 
+            var defaultConsoleDescriptor = builder.Services.FirstOrDefault(d => d.ImplementationType == typeof(ConsoleLoggerProvider));
+            if (defaultConsoleDescriptor != null)
+            {
+                builder.Services.Remove(defaultConsoleDescriptor);
+            }
+
```
TimHess, line 41: As-is, this seems like a change that might be too extreme for a patch release (actively removing something that wasn't added by our code). Can we make it opt-in and switch it to be default behavior down the road?

hananiel: We are just making it consistent with DynamicConsole -  https://github.com/SteeltoeOSS/Steeltoe/blob/4a396adf1069ccd1ad3bc8603719569579efae7b/src/Logging/src/DynamicLogger/DynamicLoggingBuilder.cs#L33
But if you feel strongly I can add a flag and allow opt in.

TimHess: I'm in favor of it being this way -- I'm just hesitating on adding it _in a patch release_ where it is on-by-default and there's no opt-out.

hananiel: Added an opt-out to preserve default logger.

src/Logging/src/SerilogDynamicLoggerCore/SerilogBuilderExtensions.cs
```diff
@@ -20,18 +21,28 @@ namespace Steeltoe.Extensions.Logging.SerilogDynamicLogger
     public static class SerilogBuilderExtensions
     {
         /// <summary>
-        /// Add Steeltoe logger wrapped in a <see cref="IDynamicLoggerProvider"/> that supports
+        /// Add Serilog logger wrapped in a <see cref="IDynamicLoggerProvider"/> that supports
         /// dynamically controlling the minimum log level via management endpoints
         /// </summary>
         /// <param name="builder">The <see cref="ILoggingBuilder"/> for configuring the LoggerFactory  </param>
+        /// <param name="preserveDefaultConsole">When true, do not remove the default console</param>
```
TimHess, line 28:
```suggestion
        /// <param name="preserveDefaultConsole">When true, do not remove Microsoft's ConsoleLoggerProvider</param>
```

TimHess (dismissed):

TimHess (approved):

PR #515, opened 2020-11-16
Update Config Server Integration Test

TimHess:
After updating the config server docker image used by the integration tests, the response body on the request in this test has changed to include some extra text at the end (` (document #0)`). This is an update for the assertion on that test so that the test will pass with either response

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -125,7 +125,7 @@ jobs:
     inputs:
       command: test
       projects: '**/*.ITest/*.csproj'
-      arguments: '--blame --no-build -c $(buildConfiguration) -maxcpucount:1 /p:CopyLocalLockFileAssemblies=true $(skipFilter) --collect:"XPlat Code Coverage" --settings coverlet.runsettings --logger trx --results-directory $(Build.SourcesDirectory)'
+      arguments: '--blame -c $(buildConfiguration) -maxcpucount:1 /p:CopyLocalLockFileAssemblies=true $(skipFilter) --collect:"XPlat Code Coverage" --settings coverlet.runsettings --logger trx --results-directory $(Build.SourcesDirectory)'
       publishTestResults: false
   - script: |
       docker kill configserver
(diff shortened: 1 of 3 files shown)


azure-pipelines.yml
```diff
@@ -125,7 +125,7 @@ jobs:
     inputs:
       command: test
       projects: '**/*.ITest/*.csproj'
-      arguments: '--blame --no-build -c $(buildConfiguration) -maxcpucount:1 /p:CopyLocalLockFileAssemblies=true $(skipFilter) --collect:"XPlat Code Coverage" --settings coverlet.runsettings --logger trx --results-directory $(Build.SourcesDirectory)'
```
TimHess, line 128: Removing `--no-build` seems to slow down test execution by enough time that the config server is ready for requests by the time the tests run (2 of the failing tests seemed to be for this reason), but it isn't the most scientific way to make sure the config server is ready

TimHess:
The discovery-first test should actually be failing as there is no discovery server running for these integration tests, so that needs to be fixed to be able to accurately show success and failure.

src/Configuration/test/ConfigServer.ITest/ConfigServerConfigurationExtensionsIntegrationTest.cs
```diff
@@ -433,7 +439,9 @@ public async void SpringCloudConfigServer_WithHealthEnabled_ReturnsHealth()
             using var server = new TestServer(builder);
             var client = server.CreateClient();
             var result = await client.GetStringAsync("http://localhost/Home/Health");
-            Assert.Equal("UP,https://github.com/spring-cloud-samples/config-repo/foo-development.properties,https://github.com/spring-cloud-samples/config-repo/foo.properties,https://github.com/spring-cloud-samples/config-repo/application.yml,", result);
+
+            // after switching to new config server image, this value now ends with " (document #0),"
```
TimHess, line 443: This new text is indeed a feature and not a bug - the # increments with stacked config files

hananiel (approved):

PR #538, opened 2020-12-04
Fix addresses

### LINKED ISSUE #537: DetermineAddresses never returns a empty or null value
## Describe the bug
The `DetermineAddresses()` method always returns a value. Because of this the calling method `SetAddresses()` will always try to set an address. This forces the use of `AmqpTcpEndpoint` to connect to RabbitMQ during the `IConnectionFactory` connection calls. Rather than the service name connection.

## Steps to reproduce 
1. Configure `RabbitOptions` without setting the `addresses` property.
2. Run App

## Expected behavior
Library should not set addresses if not configured. 

 - Steeltoe Version: 3.0.1
 - Platform: Tanzu/PCF
 - OS: Windows
 - .NET Version: 3.1

TravisBoatman:
This fixes #537

### DIFF
--- a/src/Messaging/src/RabbitMQ/Config/RabbitOptions.cs
+++ b/src/Messaging/src/RabbitMQ/Config/RabbitOptions.cs
@@ -60,7 +60,7 @@ namespace Steeltoe.Messaging.RabbitMQ.Config
             var parsed = ParsedAddresses;
             if (parsed.Count == 0)
             {
-                return Host + ":" + Port;
+                return null;
             }
 
             var addressStrings = new List<string>();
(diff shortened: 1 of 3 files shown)


dtillman (commented):
Would it not be enough to just make the change to RabbitOptions (i.e. return null) and leave the rest the same?

dtillman:
@TravisBoatman : Any thoughts ?

TravisBoatman:
Good catch. Yes, you could just include the RabbitOption changes and leave everything else.

jkonicki:
Closing in favor of #552

Issue #526, opened 2020-11-24
Cant Get Eureka Service Discovery To Work


TheLazyLemur:
## Question
I cant seem to get Eureka work, I am not listing it as a bug as I am not sure if it is. The issue im having is that my services do not register with Eureka. Im running Eureka with ```docker run --publish 8761:8761 steeltoeoss/eureka-server``` Im am using a webapi project generated by the Steeltoe Initialize. 

## Environment
 - .NET Core 3.1.403
 - Steeltoe 3.0.1

## Additional context or links
Eureka says it is running on 172.17.0.3 although it is still accessible from localhost, I don't think this is the issue though.
My appsettings.json looks like this:

```json
{
  "Logging": {
    "LogLevel": {
      "Default": "Debug",
      "System": "Information",
      "Microsoft": "Information"
    }
  },
  "spring": {
    "application": {
      "name": "TestApi"
    }
  },
  "eureka": {
    "client": {
      "serviceUrl": "http://localhost:8761/eureka/",
      "shouldFetchRegistry": "false",
      "shouldRegisterWithEureka": true,
      "validateCertificates": false
    },
    "instance": {
      "port": "5000",
      "ipAddress": "127.0.0.1",
      "preferIpAddress": true
    }
  }
}
```

and my Startup Configurations :

```cs
        public void ConfigureServices(IServiceCollection services)
        {
            services.AddDiscoveryClient(Configuration);
            services.AddControllers();
        }

        // This method gets called by the runtime. Use this method to configure the HTTP request pipeline.
        public void Configure(IApplicationBuilder app, IWebHostEnvironment env)
        {
            if (env.IsDevelopment())
            {
                app.UseDeveloperExceptionPage();
            }

            app.UseDiscoveryClient();
            app.UseRouting();
            app.UseEndpoints(endpoints =>
            {
                endpoints.MapControllers();
            });
        }
```

I do notice this warning on startup

```
 No discovery client has been configured, using default no-op discovery client.
```

TheLazyLemur:
Okay... So I found the solution to my problem. I switched  to using Consul and then I just needed to install and extra package.. namely
```
Steeltoe.Discovery.Consul
```
after that it worked perfectly. 

I have not tested it but using 
```
Steeltoe.Discovery.Eureka
```
may fix the issue if you stick to Eureka.
This step is not in the official getting started guide, therefore I will leave this open for now.

TimHess:
[...]

Issue #459, opened 2020-10-07
Write error log when deployed to cloud and default service information is used


TimHess:
Due to the changes in connectors to separate out Cloud Foundry functionality, some Steeltoe users are struggling with service bindings in cloud hosted scenarios because they don't realize they need another package reference.

In order to improve this experience, (at a minimum for Eureka on Cloud Foundry), write error logs if service bindings aren't detected but are effectively required.

* Use `Platform.IsCloudHosted` or `Platform.IsCloudFoundry` to detect running on the cloud
* Log errors if/when default service connection values are used (eg: http://localhost:8761)
* This change might go best in [EurekaDiscoveryClientExtension](https://github.com/SteeltoeOSS/Steeltoe/blob/master/src/Discovery/src/Eureka/EurekaDiscoveryClientExtension.cs) or [EurekaPostConfigurer](https://github.com/SteeltoeOSS/Steeltoe/blob/master/src/Discovery/src/Eureka/EurekaPostConfigurer.cs)

TimHess:
Logging isn't going to be very natural (or appropriate) where this needs to run (`services.PostConfigure<T>(...)`) so this is probably going to be best accomplished by throwing an exception, which isn't ideal but still seems valid to me

PR #539, opened 2020-12-04
Discovery Runtime/Deployed Configuration Hints

TimHess:
- Throw InvalidOperationException when a discovery server is configured with `localhost` in a containerized or cloud environment to help with #459 
- Add a log entry for any discovery client that might be configured and still needs a nuget ref #526


src/Discovery/test/ClientBase.Test/NoOpDiscoveryClientTest.cs
```diff
@@ -0,0 +1,55 @@
...
+            {
+                { "spring:application:name", "myName" },
+                { "Consul:somekey", "somevalue" },
+                { "Eureka:somekey", "somevalue" },
+                { "spring:cloud:kubernetes:discovery:somekey", "somevalue" },
+                { "DiscoveryClients:TestClient", "TestClientConfigPath" },
+                { "TestClientConfigPath:Uri", "http://someserver:1234" }
+            };
+            var config = new ConfigurationBuilder().AddInMemoryCollection(appsettings).Build();
+            var logger = new Mock<ILogger<NoOpDiscoveryClient>>();
```
hananiel, line 31: This works and is fine as is. I did run into a few issues with logging, testing logging and logging while debugging tests and we have solved this in several ways in the code base. perhaps we can standardize on this in the future if it works for us : https://github.com/roryprimrose/Divergic.Logging.Xunit

hananiel (approved):

PR #561, opened 2021-01-10
Assist IDE to create members that start with underscore

kimbell:
* Coding guidelines state members should start with _
* Not enabled by default in Visual Studio
* Extend .editorconfig to support this


TimHess (approved):
LGTM, thanks!

PR #562, opened 2021-01-12
Update Coverlet and SonarAnalyzer

TimHess:
- Update Coverlet to a GA version, drop the dev feed reference from nuget.config #563 
- Ignore or resolve several new warnings that show up as a result of the SonarAnalyzer update

I made a mistake. I need to make a PR for 2.x and merge that into this branch to keep the history clean. The changes will be the same, but the git info will change

### DIFF
--- a/src/CircuitBreaker/src/Hystrix.MetricsStreamCore/RabbitMetricsStreamPublisher.cs
+++ b/src/CircuitBreaker/src/Hystrix.MetricsStreamCore/RabbitMetricsStreamPublisher.cs
@@ -32,7 +32,7 @@ namespace Steeltoe.CircuitBreaker.Hystrix.MetricsStream
             if (sslOption != null && sslOption.Enabled)
             {
                 logger?.LogInformation("Hystrix Metrics using TLS");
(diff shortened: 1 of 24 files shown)


src/CircuitBreaker/src/Hystrix.MetricsStreamCore/RabbitMetricsStreamPublisher.cs
```diff
@@ -32,7 +32,7 @@ public RabbitMetricsStreamPublisher(IOptions<HystrixMetricsStreamOptions> option
             if (sslOption != null && sslOption.Enabled)
             {
                 logger?.LogInformation("Hystrix Metrics using TLS");
-                sslOption.Version = SslProtocols.Tls | SslProtocols.Tls11 | SslProtocols.Tls12;
+                sslOption.Version = SslProtocols.Tls12 | SslProtocols.Tls13;
```
TimHess, line 35: Tests aren't likely to notice this change, I'm not sure about real-world impact but everything less than TLS 1.2 has been deprecated

hananiel (dismissed):

hananiel (approved):

PR #576, opened 2021-01-21
Fix HttpClient using expired client certificate after rotation

macsux:
Fix HttpClient using expired client certificate after rotation


TimHess:
This PR makes `ClientCertificateHttpHandlerProvider` obsolete, but still uses it. Before merging, `ClientCertificateHttpHandler` should be used directly. We were planning to do some other work with Eureka for the 3.1 release, but I'm not sure that it will fit in this timeframe anymore.

@macsux is this still a priority? If the current version of the changes in this PR fix the problem then we could remove the [obsolete] and merge now, otherwise it might be waiting a bit longer

TimHess:
Replaced by #778

Issue #578, opened 2021-01-25
OAuth Service Binding URLs


kenswan:
Currently, if SSO service bindings are not found during CloudFoundry OAuth extension or CloudFoundry JWT Bearer extension, default URL are kept and not leveraging the current auth domain setting (if available). OAuth URL domains should default to Auth Domain app setting if SSO service bindings are not available.

**AddCloudFoundryOAuth**
- App Setting
  - "Security -> OAuth2 -> Client -> AuthDomain"
- Properties
  - Authorization Endpoint
  - Token Endpoint
  - User Information Endpoint
  - Token Info Url

**AddCloudFoundryJwtBearer**
- App Setting
  - "Security -> OAuth2 -> Client -> AuthDomain"
- Properties:
  - Jwt Key Url

PR #586, opened 2021-02-02
OAuth Service URL Fallback (3.x version)

kenswan:
Currently, if SSO service bindings are not found during CloudFoundry OAuth extension or CloudFoundry JWT Bearer extension, default URL are kept and not leveraging the current auth domain setting (if available). OAuth URL domains should default to Auth Domain app setting if SSO service bindings are not available.

Addresses #578 for 3.x

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -15,7 +15,7 @@ jobs:
   variables:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     DOTNET_CLI_TELEMETRY_OPTOUT: 1
-    PackageVersion: $[format('2.5.2-ci{0}', variables['Build.BuildId'])]
+    PackageVersion: $[format('2.5.3-ci{0}', variables['Build.BuildId'])]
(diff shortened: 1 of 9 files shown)


azure-pipelines.yml
```diff
@@ -15,7 +15,7 @@ jobs:
   variables:
     DOTNET_SKIP_FIRST_TIME_EXPERIENCE: true
     DOTNET_CLI_TELEMETRY_OPTOUT: 1
-    PackageVersion: $[format('2.5.2-ci{0}', variables['Build.BuildId'])]
+    PackageVersion: $[format('2.5.3-ci{0}', variables['Build.BuildId'])]
```
TimHess, line 18: This diff doesn't make sense to me... This value should currently be 3.1.0....? Regardless, it shouldn't be changed in this PR

TimHess (changes requested):
The build produced the wrong package version

kenswan:
Closing PR to start a new 2.x based branch

Issue #580, opened 2021-01-28
(Web)Hostbuilder.AddKubernetesActuators doesn't pass mediaTypeVersion everywhere it should


TimHess:
mediaTypeVersion is passed to the startupfilter but not IServiceCollection.AddKubernetesActuators
```
                .ConfigureServices((context, collection) =>
                {
                    collection.AddKubernetesActuators();
                    collection.AddSingleton<IStartupFilter>(new KubernetesActuatorsStartupFilter(mediaTypeVersion));
                });
```

This results in a runtime error if V1 is passed in (attempts to startup Trace when HttpTrace is what was setup)

jkonicki:
Completed in #580

Issue #497, opened 2020-11-04
Usage of UseEndpoints in IStartupFilter


kimbell:
We are building common libraries for use in our .NET Core services infrastructure. Some of these need to expose endpoints or add custom middleware. We are using Steeltoe for Eureaka, so I did some exploring of your codebase. 

I see a pattern of using an `IStartupFilter `for modifying the endpoints using `UseEndpoints()`. This is not an approach I've seen anywhere else. I asked a question related to this on the ASP.NET Core repo, and got a reply stating that this was not ideal since it would add multiple EndpointMiddlware to the pipeline. 

https://github.com/dotnet/aspnetcore/discussions/27503

I'm curious on how you landed on this approach? I'm also struggling to find a built in way of making the middleware pipeline extendable based on what services have been added. I have some ideas, so if they pan out, I'll let you know how we approached it.

TimHess:
While we define multiple `IStartupFilter`s, I think the intent is that only one should ever actually be used. There could be room for improvement here, at a minimum addressing the issue in the documentation, but for anybody that wants to avoid `UseEndpoints()` being called multiple times, it is also possible to map the endpoints more directly:
```
app.UseEndpoints(endpoints =>
{
    // map individual endpoints
    endpoints.Map<CloudFoundryEndpoint>();
    // map all endpoints
    endpoints.MapAllActuators(MediaTypeVersion);
});
```

cc @hananiel in case I missed anything here

kimbell:
The extension methods for `IHostBuilder  `and `IWebHostBuilder `contains methods like `AddEnvActuator `and `AddHealthActuator`. These methods register an `IStartupFilter` each. 

If you use these methods instead of `AddAllActuators`, a number of `IStartupFilters `will be registered. 

The way I understand the use of these AddXYZActuator is to make it a 'one liner' to set things up, rather than adding things to `IServiceCollection `and `IEndpointRouteBuilder` individually.

TimHess:
Correct. But just because they've been added doesn't mean they've been activated, so for cases where you want more than one you should be able to add them all and then selectively enable only the ones you want. By default there's only a couple that are actually enabled

kimbell:
In the documentation at https://docs.steeltoe.io/api/v3/management/using-endpoints.html 

> Extensions for both IHostBuilder and IWebHostBuilder are included to configure actuators with a single line of code in program.cs:

[...]

PR #591, opened 2021-02-04
Use AllActuatorsStartupFilter more

TimHess:
- Dial down to one in-use `IStartupFilter` for almost all actuators #497
- `[Obsolete]` actuator startup filters that don't do anything special. 
- MediaTypeVersion param in k8s extensions #580


src/Management/src/EndpointCore/ManagementHostBuilderExtensions.cs
```diff
@@ -92,182 +86,195 @@ public static IHostBuilder AddHealthActuator(this IHostBuilder hostBuilder, Type
...
-                    collection.AddTransient<IStartupFilter, AllActuatorsStartupFilter>(provider => new AllActuatorsStartupFilter(configureEndpoints));
+                    ActivateActuatorEndpoints(collection, configureEndpoints);
                 });
+
+        /// <summary>
+        /// Registers an <see cref="IStartupFilter" /> that will map all configured actuators, initialize health
+        /// </summary>
+        /// <param name="collection"><see cref="IServiceCollection" /> that has actuators to activate</param>
+        /// <param name="configureEndpoints">IEndpointConventionBuilder customizations (such as auth policy customization)</param>
+        public static void ActivateActuatorEndpoints(this IServiceCollection collection, Action<IEndpointConventionBuilder> configureEndpoints = null)
```
TimHess, line 253: @hananiel this is new, please focus here

src/Management/src/EndpointCore/AllActuatorsStartupFilter.cs
```diff
@@ -18,6 +20,21 @@ public AllActuatorsStartupFilter(Action<IEndpointConventionBuilder> configureCon
 
         private readonly Action<IEndpointConventionBuilder> _configureConventions;
 
+        public static void InitializeAvailability(IServiceProvider serviceProvider)
```
hananiel, line 23: Can this be internal? may be we shouldn't expose these at all  - if possible.

hananiel (approved):
LGTM; See comment on the AllActuatorsStartupFilter.

src/Management/src/EndpointCore/AllActuatorsStartupFilter.cs
TimHess: This method should be available for anybody that doesn't want to use the Hostbuilder/StartupFilter approach, so it should be public... We could put it somewhere else if that would address your concern

src/Management/src/EndpointCore/ManagementHostBuilderExtensions.cs
TimHess: Note: this method allows application of one unified set of endpoint convention customization for all actuator endpoints. If anybody wants to do per-endpoint customization they can/should use [`Map<TEndpoint>`](https://github.com/SteeltoeOSS/Steeltoe/blob/master/src/Management/src/EndpointCore/ActuatorRouteBuilderExtensions.cs#L62) directly for each endpoint instead of this more general one-size-fits-most option

hananiel (approved):

Issue #545, opened 2020-12-07
Refactoring the Management endpoints


kimbell:
The discussion on #503 has drifted from the original topic. This issue has been created for the general discussion on management endpoints so that the original issue can resume it's original duty.

My thoughts on improving the management endpoint features can be broken down into the following phases.

Phase 1 is to get #543 merged since that makes testing easier.

Phase 2 is to update the tests so that DI is the primary mechanism for creating objects. This allows us to refactor the internals without breaking too many tests. A lot of tests are currently using `TestHost` and could be updated to use `WebApplicationFactory` that was created by Microsoft for testing purposes; wraps `TestHost` and sets up other things.
Since we are updating tests, it allows us to fix #542. I have an approach that doesn't depend on external libraries that works for most cases; depends on how much you want to analyze the logging output during the test.

Phase 3 is to update the internals of the management endpoint code. I have some ideas on how this can be achieved. One idea I'm toying around with is using an `EndpointDataSource` for the actuator and CloudFoundry endpoints.

kimbell:
I'm looking at this from a .NET Core perspective. Great that there was support for .NET Framework, but long term, Core will be the only game in town. Based on that, there are things that should be done to align it with how .NET Core does things. The big challenge is how do we get there, and how can we incrementally make changes that don't break everything. At some point in the future, the code will just have to accept that .NET Framework is dead and buried. As long as `EndpointBase` is constrained by history, it restricts what type of changes can be made. When that time comes, I don't know. 

I have played around with several ideas on what can be changed, but usually hit a dead end that would cause too many changes that are hard to track or become a binary breaking change. I'm trying to find small changes that don't break things completely and can be used to aid the long term goals of the project; makes it easier to merge between branches. This to delay the split from .NET Framework as long as possible. 

So what types of changes would I like to make? 

The code contains a lot of Option objects. For .NET Core developers, that comes with a lot assumptions since Microsoft has the `IOptions` pattern. It would not be unreasonable for .NET Core developers to think that they could leverage all the possibilities of the IOptions pattern. i.e. `Configure()` and `PostConfigure()`. 

Since Steeltoe isn't using this pattern, it may become a developer surprise. I couldn't find anything in the documentation stating that `IOptions` is not supported. The options pattern is to expose configuration values consumers can change. The Options classes in Steeltoe contain options that can't be changed by consumers; `AllowedVerbs` and `ExactMatch`. Another small surprise. 



[...]

PR #600, opened 2021-02-06
Reduce the number of direct constructor callers to management endpoint base

kimbell:
This is another step on the journey to refactoring the management endpoint code. 
It contains a lot of changes, but they are similar for each endpoint. 

- A public interface is added for each endpoint
- Interface can be used for public consumption so that we can make classes internal in the future. 
- Service collection contains entry that maps interface to actual implementation

At some point we can mark code as obsolete to trigger consumers to move over to interface. 
MetricsEndpoint is using some protected internal code that prevented some code from using an interface.

In the test project I've created a `TestContext` class that contains the state for each test. This adds Options and Logging infrastructure. Tests are  now using `IServiceProvider `to create endpoint rather than using the constructor directly.

Tests that log information is now available in the XUnit output.

> Unable to access the Dynamic Logging provider, log configuration unavailable
> Health check requested for a group that is not configured

related to #545


TimHess (commented):
Looks fine to me - we can merge after conflicts are resolved, unless @hananiel has any thoughts

dtillman (commented):
LGTM

TimHess (approved):

Issue #607, opened 2021-02-23
(Web)Hostbuilder.AddDynamicSerilog default configuration issues


TimHess:
As a Steeltoe user, I expect `(Web)HostBuilder.AddDynamicSerilog()` to use appsettings to configure Serilog and add a console sink.

Because of the way some refactoring for 3.0 was done around [the action that customizes the loggerConfiguration](
https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Logging/src/DynamicSerilogCore/SerilogWebHostBuilderExtensions.cs#L30), a default/new `LoggerConfiguration` is passed into other methods instead of `null` and efforts to [centralize the default config](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Logging/src/DynamicSerilogBase/SerilogLoggingBuilderExtensions.cs#L57) of the refactor are lost as a result.

Related, it looks like there's yet another level of default logger configuration at the [LoggerProvider](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Logging/src/DynamicSerilogBase/SerilogDynamicProvider.cs#L71) level that does not seem like it should be necessary

PR #608, opened 2021-02-23
Use app's configuration when adding Serilog to a hostbuilder

TimHess:
Use configuration from the hostcontext for Serilog config, add a console sink if no LoggerConfiguration action was provided #607

### DIFF
--- a/src/Logging/src/DynamicSerilogBase/SerilogConfigurationExtensions.cs
+++ b/src/Logging/src/DynamicSerilogBase/SerilogConfigurationExtensions.cs
@@ -4,11 +4,26 @@
 
 using Serilog;
 using Serilog.Configuration;
+using Serilog.Core;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
 
 namespace Steeltoe.Extensions.Logging.DynamicSerilog
 {
     internal static class SerilogConfigurationExtensions
     {
         public static LoggerConfiguration SerilogConsole(this LoggerSinkConfiguration sinkConfiguration) => sinkConfiguration.Console();
+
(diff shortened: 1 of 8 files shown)


src/Logging/src/DynamicSerilogBase/SerilogConfigurationExtensions.cs
```diff
@@ -4,11 +4,26 @@
...
 
 namespace Steeltoe.Extensions.Logging.DynamicSerilog
 {
     internal static class SerilogConfigurationExtensions
     {
         public static LoggerConfiguration SerilogConsole(this LoggerSinkConfiguration sinkConfiguration) => sinkConfiguration.Console();
+
+        internal static LoggerConfiguration AddConsoleIfNoSinksFound(this LoggerConfiguration loggerConfiguration)
+        {
+            var configuredSinksField = loggerConfiguration.GetType().GetField("_logEventSinks", BindingFlags.NonPublic | BindingFlags.Instance);
```
TimHess, line 20: SonarCloud is (rightfully) complaining about use of reflection here to access a non-public field, but I'm not sure that there is any other way to detect whether a sink has already been added or not

TimHess: This _may_ be excessive as having two ConsoleSinks does not appear to result in duplicate logging, so maybe this is all just excessive ?

hananiel: One of the reasons I remember adding this is people trying it would remove the configured(sample) sink and wonder why it wasn't working (myself included). But I wasn't sure about dual logging - that has been an issue as well. I was thinking may be check for at console sink and add it if not present - so if they add some remote sink then still be able to see something in console - unless someone wants its really not to be present.

hananiel (approved):

PR #615, opened 2021-03-03
Fix dynamicLoggerProvider resolution in EndpointOwinAutofac

### LINKED ISSUE #460: EndpointOwinAutofac doesn't correctly set up DynamicLoggerProvider
The logging actuator isn't working when setup via `EndpointOwinAutofac` since around 2.4.0 when all [constructors that don't use IOptionsMonitor were removed](https://github.com/SteeltoeOSS/Steeltoe/pull/142/files#diff-0991b423adf292b31e6788c3f35316b1L42)

```
None of the constructors found with 'Autofac.Core.Activators.Reflection.DefaultConstructorFinder' on type 'Steeltoe.Extensions.Logging.DynamicConsoleLoggerProvider' can be invoked with the available services and parameters:
Cannot resolve parameter 'Microsoft.Extensions.Options.IOptionsMonitor`1[Microsoft.Extensions.Logging.Console.ConsoleLoggerOptions] options' of constructor 'Void .ctor(Microsoft.Extensions.Options.IOptionsMonitor`1[Microsoft.Extensions.Logging.Console.ConsoleLoggerOptions], Microsoft.Extensions.Options.IOptionsMonitor`1[Microsoft.Extensions.Logging.LoggerFilterOptions], System.Collections.Generic.IEnumerable`1[Steeltoe.Extensions.Logging.IDynamicMessageProcessor])'.
```

The [RegisterLogging](https://github.com/SteeltoeOSS/Steeltoe/blob/2.x/src/Common/src/Common.Autofac/Logging/LoggingContainerBuilderExtensions.cs#L33) extension looks like it tries to setup IOptions for `LoggerFilterOptions`, but IOptionsMonitor<LoggerFilterOptions> does not seem to be resolveable. Additionally it could use the [extension in `Common.Autofac`](https://github.com/SteeltoeOSS/Steeltoe/blob/2.x/src/Common/src/Common.Autofac/Options/OptionsContainerBuilderExtensions.cs#L28) to avoid duplication

It might be worth changing this up to use a temporary Microsoft DI container to build logging and/or the options and dropping that into the Autofac container

TimHess:
Register IOptions types so that the IDynamicLoggerProvider can be resolved #460

### DIFF
--- a/src/Common/src/Common.Autofac/Logging/LoggingContainerBuilderExtensions.cs
+++ b/src/Common/src/Common.Autofac/Logging/LoggingContainerBuilderExtensions.cs
@@ -7,6 +7,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Logging.Console;
 using Microsoft.Extensions.Options;
+using Steeltoe.Common.Options.Autofac;
 using System;
 
 namespace Steeltoe.Common.Logging.Autofac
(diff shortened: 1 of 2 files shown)


hananiel (approved):
LGTM

Issue #384, opened 2020-07-27
Remove Json* Classes in EurekaBase/Transport


TimHess:
The Eureka discovery client includes a handful of files that might be replaceable by reworking the serialization and de-serialization. Most/all of these files have `Json` in the name and are `internal`, and could potentially be completely removed.

kenswan:
When digging into the code change, there were a few reasons to hold on this issue

- This required removing some internal setters which exposes the client to new methods of the api. This could potentially cause harm if the client were to override certain aspects of these internal objects.
- Keeping the json classes as DTOs allows us to flow with the changes of Eureka behind the scenes without affecting our clients
- With the amount of changes that would be needed for this conversion it introduced more risk and regression testing than reward for making the change

PR #644, opened 2021-04-22
Eureka Transport Json* Class Removal

kenswan:
Moving away from having json classes for deserialization purposes and using base models instead.

Addresses #384


dtillman:
@kenswan .... Hey, lets get on a zoom and discuss this further!

kenswan:
When digging into the code change, there were a few reasons to hold on this issue

- This required removing some internal setters which exposes the client to new methods of the api. This could potentially cause harm if the client were to override certain aspects of these internal objects.
- Keeping the json classes as DTOs allows us to flow with the changes of Eureka behind the scenes without affecting our clients
- With the amount of changes that would be needed for this conversion it introduced more risk and regression testing than reward for making the change

PR #653, opened 2021-04-30
Fix error message; fix finding binder in current directory

hananiel:
* Sometimes the cause on shutdowneventArgs is empty, but the replyText and Status have information which is exposed by ToString() - this leaves a better error message when failures occur (such as reading from an exchange that doesn't exist)
* We exclude the CurrentDirectory (which is added by default in the constructor) for testing. But this interferes with generic hosts when the binder is in the current directory.

### DIFF
--- a/src/Stream/src/StreamBase/Binder/DefaultBinderTypeRegistry.cs
+++ b/src/Stream/src/StreamBase/Binder/DefaultBinderTypeRegistry.cs
@@ -165,11 +165,8 @@ namespace Steeltoe.Stream.Binder
         {
             var paths = new List<string>();
             paths.AddRange(Directory.GetFiles(RuntimeEnvironment.GetRuntimeDirectory(), "*.dll"));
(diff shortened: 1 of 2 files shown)


src/Stream/src/StreamBase/Binder/DefaultBinderTypeRegistry.cs
```diff
@@ -165,11 +165,8 @@ internal static PathAssemblyResolver GetAssemblyResolver(string directory)
         {
             var paths = new List<string>();
             paths.AddRange(Directory.GetFiles(RuntimeEnvironment.GetRuntimeDirectory(), "*.dll"));
-            if (!Environment.CurrentDirectory.Equals(directory, StringComparison.InvariantCultureIgnoreCase))
-            {
-                paths.AddRange(Directory.GetFiles(directory, "*.dll"));
-            }
-
+            paths.AddRange(Directory.GetFiles(directory, "*.dll"));
+
```
TimHess, line 169:
```suggestion

```

TimHess (approved):

Issue #664, opened 2021-05-19
TypeConverter does not convert type List<dictionary<string, object>> 


hananiel:
When passing a List<Dictionary<string, object>>, as when the x-death header is passed Streams Method that expects an IDictionary<string, object>, it throws a conversion error. 
The correct behavior is to find the first dictionary element and convert that to a dictionary.

Issue #663, opened 2021-05-19
Error when sending messages to Streams Listener without Content type


hananiel:
When a Stream Listener is setup and a message is published with a string body (payload) without setting "content_type" property an exception is thrown. 
The correct response should be to default to "application/json" and correctly call the stream listener method.

PR #665, opened 2021-05-19
Fix errors in type conversion (from x-death sample) 

hananiel:
Address issues with default content type detection (#663) and type conversion of list<dictionary> (#664)


src/Messaging/src/RabbitMQ/Support/DefaultMessageHeadersConverter.cs
```diff
@@ -95,11 +95,6 @@ public virtual void FromMessageHeaders(IMessageHeaders headers, RC.IBasicPropert
                 target.ContentType = source.ContentType();
             }
 
-            if (string.IsNullOrEmpty(target.ContentType))
-            {
-                target.ContentType = RabbitHeaderAccessor.DEFAULT_CONTENT_TYPE;
```
hananiel, line 100: @dtillman setting default this in messaging code, causes issues in the Streams code with its default content type detection. I removed this here and in other places, All the tests are passing so I assume there is no issue, but do you see any problem with this?

kenswan (approved):

src/Messaging/src/RabbitMQ/Support/DefaultMessageHeadersConverter.cs
dtillman: OK .. sounds good ... can't remember why I did that... might want to double check the Java code to make sure it's all right!

hananiel: Yes the Java code was not doing this, was just making sure the messaging aspects did not need this.

PR #670, opened 2021-05-21
Async tests should return Task

TimHess:
The primary purpose of this PR is to prevent async tests from operating as fire and forget (Identified by sonar update in previous PR) other issues addressed:
- remove a few instances of `[Trait("Category", "FlakyOnHostedAgents")]` which hasn't been used for a while
- skip some tests that appear to be causing testhost crashes due to underlying issue in the runtime in Mac + .NET 5


kenswan (approved):
LGTM

Issue #639, opened 2021-04-21
Kubernetes ConfigMap file parsing


DaviGia:
## Describe problem
If I declare a file within a ConfigMap the `KubernetesConfigProvider` reads it as a single property and does not evaluate the value as json. Spring provides a similar functionality where you can specify a yaml or properties file within a ConfigMap and it will be parsed correctly ([link](https://docs.spring.io/spring-cloud-kubernetes/docs/current/reference/html/#configmap-propertysource)).

## Describe solution
I would like to be able to specify a file within ConfigMap and have all the content parsed correctly.

For example, provided the following ConfigMap:

``` yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: service-config
data:
  appsettings.json: |
    {
      "Test0": "Value0",
      "Test1": [
         {
            "Test2": "Value1"
         }
      ]
    }
```

the `KubernetesConfigProvider` should translate it in the following configuration properties:

|Key|Value|
|--|--|
|Test0|Value0|
|Test1:0:Test2|Value1|

If possible, it would be nice allow the provider to parse not only json but even yaml. 

## Describe alternatives
At the moment I will fallback into using a simple map in the ConfigMap.

``` yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: service-config
data:
  "Test0": "Value0"
  "Test1:0:Test2": "Value1"
```

PR #671, opened 2021-05-22
Add ConfigMap appsettings file parsing

tscrypter:
- Add support for ConfigMap entry with key `appsettings.json` being parsed using the JsonConfigurationProvider
- ConfigMap entries with keys duplicated in the `appsettings.json`'s value will use the values specified in the ConfigMap
#639

### DIFF
--- a/src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
+++ b/src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
@@ -19,7 +19,8 @@ namespace Steeltoe.Extensions.Configuration.Kubernetes
 {
     internal class KubernetesConfigMapProvider : KubernetesProviderBase, IDisposable
     {
-        private const string ConfigFileKey = "appsettings.json";
+        private const string ConfigFileKeyPrefix = "appsettings";
+        private const string ConfigFileKeySuffix = "json";
 
         private Watcher<V1ConfigMap> ConfigMapWatcher { get; set; }
 
@@ -134,14 +135,12 @@ namespace Steeltoe.Extensions.Configuration.Kubernetes
             {
                 foreach (var data in item?.Data)
                 {
-                    // Should this check for application.{env.Environment}.json as well?
-                    if (string.Equals(data.Key, ConfigFileKey, StringComparison.InvariantCultureIgnoreCase))
+                    if (IsAppsettingsKey(data.Key))
                     {
                         using var stream = this.GenerateStreamFromString(data.Value);
                         var jsonConfiguration = ParseConfigMapFile(stream);
 
-                        // Give precedence to overrides in the configmap, so ignore duplicate keys
-                        foreach (var jsonKey in jsonConfiguration.Keys.Where(k => !item.Data.Keys.Contains(k)))
+                        foreach (var jsonKey in jsonConfiguration.Keys)
                         {
                             configMapContents[NormalizeKey(jsonKey)] = jsonConfiguration[jsonKey];
                         }
@@ -154,6 +153,12 @@ namespace Steeltoe.Extensions.Configuration.Kubernetes
             }
 
             Data = configMapContents;
+        }
+
+        private bool IsAppsettingsKey(string key)
+        {
+            return key.StartsWith(ConfigFileKeyPrefix, StringComparison.InvariantCultureIgnoreCase) &&
+                key.EndsWith(ConfigFileKeySuffix, StringComparison.InvariantCultureIgnoreCase);
         }
 
         private Stream GenerateStreamFromString(string s)
--- a/src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
+++ b/src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
@@ -4,11 +4,14 @@
 
 using k8s;
 using k8s.Models;
+using Microsoft.Extensions.Configuration.Json;
 using Microsoft.Extensions.Logging;
 using Microsoft.Rest;
 using Steeltoe.Common.Kubernetes;
 using System;
 using System.Collections.Generic;
+using System.IO;
+using System.Linq;
 using System.Net;
 using System.Threading;
 
@@ -16,6 +19,8 @@ namespace Steeltoe.Extensions.Configuration.Kubernetes
 {
     internal class KubernetesConfigMapProvider : KubernetesProviderBase, IDisposable
     {
+        private const string ConfigFileKey = "appsettings.json";
+
         private Watcher<V1ConfigMap> ConfigMapWatcher { get; set; }
 
         internal KubernetesConfigMapProvider(IKubernetes kubernetes, KubernetesConfigSourceSettings settings, CancellationToken cancellationToken = default)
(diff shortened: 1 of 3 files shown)


src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
```diff
@@ -121,11 +134,52 @@ private void ProcessData(V1ConfigMap item)
...
                 {
-                    configMapContents[NormalizeKey(data.Key)] = data.Value;
+                    // Should this check for application.{env.Environment}.json as well?
+                    if (string.Equals(data.Key, ConfigFileKey, StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        using var stream = this.GenerateStreamFromString(data.Value);
+                        var jsonConfiguration = ParseConfigMapFile(stream);
+
+                        // Give precedence to overrides in the configmap, so ignore duplicate keys
+                        foreach (var jsonKey in jsonConfiguration.Keys.Where(k => !item.Data.Keys.Contains(k)))
```
TimHess, line 144: Is this necessary? I would think that if you want to use appsettings.json in a configmap that that is the only key you'd use in that one and then override it with a separate configmap... I could be wrong but that's my initial thought

TimHess (commented):
Thank you for taking the time to open this PR, it's looking good

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
tscrypter: I considered this necessary to make determining which settings more easily expressed. With this approach, it can just be documented that values in the configmap will override any values in the appsettings.json config value. Without this, it will be dependent on the order key/values are processed in the configmap.

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
```diff
@@ -121,11 +134,52 @@ private void ProcessData(V1ConfigMap item)
             {
                 foreach (var data in item?.Data)
                 {
-                    configMapContents[NormalizeKey(data.Key)] = data.Value;
+                    // Should this check for application.{env.Environment}.json as well?
```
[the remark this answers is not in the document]
tscrypter: Just to make sure I implement this in the expected manner, settings would be applied as followed:

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
TimHess: But this code executes within the context of a single configmap ~ likely at least 2 instances of the `KubernetesConfigMapProvider` have been created ([appname and appname.envName](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Configuration/src/KubernetesBase/KubernetesConfigurationBuilderExtensions.cs#L45)). It seems odd to me that anybody would want to combine standard key/value pairs and one or more appsettings.json style sets of configuration within a single configmap.

Overall this isn't a huge issue and isn't a great deal of complexity whether its needed or not, but it goes against the ways I would envision configmaps being used. If this class expects either standard key/value pairs or json provided under a single key (and eventually yaml or props or whatever else) then we don't need to worry about the last-entry-wins type of logic already provided by IConfiguration

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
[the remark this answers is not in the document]
TimHess: Maybe this should just look for starts with `appsettings` and ends with `json`? 

Let's look at the example in the linked issue:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: service-config
data:
  appsettings.json: |
    {
      "Test0": "Value0",
      "Test1": [
         {
            "Test2": "Value1"
         }
      ]
    }
```

This class is scoped to look at just one configmap, so is it more likely that multiple configmaps would be configured to mirror the multiple-file expectation we have with appsettings.json, or sould anybody want to combine those entries under a single configmap?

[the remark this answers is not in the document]
tscrypter: Good point, I had overlooked the [KubernetesConfigurationBuilderExtensions.cs](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Configuration/src/KubernetesBase/KubernetesConfigurationBuilderExtensions.cs#L45) already providing an environment specific configuration provider.
So to expand on the example from the issue to also include environment specific config settings as well, there would be 2 config maps:
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: service-config
data:
  appsettings.json: |
    {
      "Test0": "Value0",
      "Test1": [
         {
            "Test2": "Value1"
         }
      ]
    }
```
```
apiVersion: v1
kind: ConfigMap
metadata:
  name: service-config.demo
data:
  appsettings.demo.json: |
    {
      "Test1": [
         {
            "Test2": "Value2"
         }
      ]
    }
```
If the application is started with no EnvironmentName specified, the KubernetesConfigMapProvider bound to `service-config` will be the only provider to process data.
If the application is started with an EnvironmentName of `demo` specified, both KubernetesConfigMapProvider created will provide data, and the one bound to `service-config.demo` will override the former because it was added 2nd [here](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Configuration/src/KubernetesBase/KubernetesConfigurationBuilderExtensions.cs#L45).

The expected values would be as follows then:
| EnvironmentName | Test0 value | Test1:0:Test2 value|
|----------------------|-------------|---------------------|
| Production             | Value 0       |  Value 1                 | 
| Demo                     | Value 0       |  Value 2                 |

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
tscrypter: I had overlooked 2 instances of the `KubernetesConfigMapProvider` being created. I can make that update.

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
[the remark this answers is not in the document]
TimHess: Yeah, that explanation matches what I would expect. What do you think of changing L138 to something along the lines of  `if the key ends with 'json' and value contains '{'` so that we aren't really caring what the key is so long as it's labeled like and looks to contain json? I don't have an overly strong opinion here, but it might be redundant to require the env name in both the configmap name and the key, especially if there's only one key

src/Configuration/src/KubernetesBase/KubernetesConfigMapProvider.cs
tscrypter: Resolved in [97f89f3e3e64ce33b778f90b1f8a54585c334ce1](https://github.com/SteeltoeOSS/Steeltoe/pull/671/commits/97f89f3e3e64ce33b778f90b1f8a54585c334ce1)

TimHess (approved):
LGTM, thanks very much for you contribution!

Issue #673, opened 2021-05-26
Bad encoding for credentials with SQL Server connector since Steeltoe 2.5.0


tremblaysimon:
## Describe the bug
Deploying an app using Cloud Foundry and MASB as service broker for Azure SQL database, the connection will not be established when the password contains special characters like `+` character. 

## Steps to reproduce 
1. Deploy an app on Cloud Foundry using an Azure SQL database (provided by MASB). 
2. Unbind / bind the app on Azure SQL database until you get a password with a `+` in it.
3. Observe that app can't connect to Azure SQL database in your app log.

## Expected behavior
The connection should work properly even if there is a special character in the password.

## Environment :
 - Steeltoe Version: reproduced on 2.5.x and 3.x.x
 - Platform: Cloud Foundry using MASB
 - .NET Version: .NET Core 3.1

## Additional context or links
Here what the VCAP_SERVICES looks like:
```
System-Provided:
{
 "VCAP_SERVICES": {
  "azure-sqldb": [
   {
    "binding_name": null,
    "credentials": {
     "databaseLogin": "fakeUsername",
     "databaseLoginPassword": "fakePasswordWith+",
     "hostname": "custom-server.database.windows.net",
     "jdbcUrl": "jdbc:sqlserver://custom-server.database.windows.net:1433;database=fakeDbName;user=fakeUsername;password=fakePasswordWith+;Encrypt=true;TrustServerCertificate=false;HostNameInCertificate=*.database.windows.net;loginTimeout=30;",
     "jdbcUrlForAuditingEnabled": "jdbc:sqlserver://custom-server.database.secure.windows.net:1433;database=fakeDbName;user=fakeUsername;password=fakePasswordWith+;Encrypt=true;TrustServerCertificate=false;HostNameInCertificate=*.database.secure.windows.net;loginTimeout=30;",
     "name": "fakeDbName",
     "password": "fakePasswordWith+",
     "port": 1433,
     "sqlServerFullyQualifiedDomainName": "custom-server.database.windows.net",
     "sqlServerName": "custom-server",
     "sqldbName": "fakeDbName",
     "uri": "mssql://fakeUsername:fakePasswordWith%2B@custom-server.database.windows.net:1433/fakeDbName?encrypt=true\u0026TrustServerCertificate=false\u0026HostNameInCertificate=%2A.database.windows.net",
     "username": "fakeUsername"
    },
    "instance_name": "fakeServiceInstanceName",
[...]

PR #676, opened 2021-05-27
Handle space and plus in user/password

TimHess:
Deal with #673 in 2.x

### DIFF
--- a/src/Connectors/src/ConnectorBase/Services/UriInfo.cs
+++ b/src/Connectors/src/ConnectorBase/Services/UriInfo.cs
@@ -19,8 +19,8 @@ namespace Steeltoe.CloudFoundry.Connector.Services
             Scheme = scheme;
             Host = host;
             Port = port;
-            UserName = WebUtility.UrlEncode(username?.Replace("+", "-plus-")?.Replace(" ", "-space-"));
-            Password = WebUtility.UrlEncode(password?.Replace("+", "-plus-")?.Replace(" ", "-space-"));
+            UserName = WebUtility.UrlEncode(username);
+            Password = WebUtility.UrlEncode(password);
             Path = path;
             Query = query;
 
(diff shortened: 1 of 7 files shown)


src/Connectors/src/ConnectorBase/Services/UriInfo.cs
```diff
@@ -66,8 +66,8 @@ public UriInfo(string uristring, string username, string password)
                 Query = GetQuery(uri.PathAndQuery);
             }
 
-            UserName = username;
-            Password = password;
+            UserName = WebUtility.UrlEncode(username?.Replace("+", "-plus-")?.Replace(" ", "-space-"));
```
hananiel, line 69: @TimHess Originally, the bug appears that we did not have a `UrlEncode` here but we do now. So if there is a `+` or `<space>` in the username or password, it will now be converted to `%2B` and a `+` respectively. And the subsequent `UrlDecode` will revert to the original. So it doesn't look like we need the `Replace("+", "-plus-")?.Replace(" ", "-space-"))`. Am I missing something ?

hananiel (changes requested):
Please see comment.

src/Connectors/src/ConnectorBase/Services/UriInfo.cs
TimHess: Yeah, you're right, this solution was over-thought

hananiel (dismissed):

ccheetham (commented):
The fix for space/plus LGTM.

Should the pipeline and nuget config changes be part of a different PR?  They don't seem germane to the fix.

TimHess:
> The fix for space/plus LGTM.
> 
> Should the pipeline and nuget config changes be part of a different PR? They don't seem germane to the fix.

Technically, probably, but the builds for this PR were failing without the nuget changes at a minimum

PR #680, opened 2021-06-02
Merge from 2.x to main

TimHess:
Includes:
- support for EF6 on .NET Core 3+
- fix for #673


src/Connectors/test/Connector.EF6Core.Test/OracleDbTypeLocatorTest.cs
```diff
@@ -0,0 +1,26 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using Steeltoe.Connector.Oracle;
+using Xunit;
+
+#if NET461
```
kenswan, line 8: Should we remove this directive for the 3.x line?

TimHess: @kenswan this test would fail... The latest Oracle package for EF6 still only works with .NET Framework. Alternatively we could skip or remove the test until such time as they do support Core or .Net 5, but leaving it this way is a reminder that it only works on Framework

kenswan: Ah, thanks for that clarification!

kenswan (approved):

Issue #606, opened 2021-02-22
RabbitMQ custom RabbitListenerErrorHandler wont get called in async context


mhehle:
## Describe the bug
If an async RabbitListener method throws an exception, the custom RabbitListenerErrorHandler wont get called and exception is swallowed.

## Steps to reproduce 

1. Create new ErrorHandlerClass `CustomErrorHandler`
``` c#
public class CustomErrorHandler : IRabbitListenerErrorHandler, IServiceNameAware
{

  public string ServiceName { get; set; } = nameof (CustomErrorHandler);

  public object HandleError(IMessage origMessage, IMessage message, ListenerExecutionFailedException exception)
  {
    Console.WriteLine("Error handled");
    return null;
  }
}
```

2. Create new Listener class 
``` c#
public class RabbitListener
{

  [DeclareQueue(Name = "MyQueue", Durable = "True")]
  [DeclareQueueBinding(QueueName = "MyQueue", ExchangeName = "MyExchange", RoutingKey = "MyRoutingKey")]
  [RabbitListener(Queue = "MyQueue", ErrorHandler = nameof(CustomErrorHandler))]
  public async Task HandleMessage(IMessage<MyMessage> msg)
  {
    throw new Exception();
  }    
}
```
3. Call handle method `await RabbitListener.HandleMessage()`

## Expected behavior
`CustomErrorHandler.HandleError()` method should be called.

## Environment (please complete the following information):
 - Steeltoe 3.0.2
 - OS: [Windows, Linux]
 - .NET Version NET 5.0

PR #682, opened 2021-06-03
RabbitListenerErrorHandler Async Context

kenswan:
Resolves issue with RabbitListenerErrorHandler not being called when errors inside async methods.

Addresses #606

### DIFF
--- a/src/Messaging/src/MessagingBase/Handler/Invocation/InvocableHandlerMethod.cs
+++ b/src/Messaging/src/MessagingBase/Handler/Invocation/InvocableHandlerMethod.cs
@@ -4,7 +4,9 @@
 
 using Microsoft.Extensions.Logging;
 using System;
+using System.Linq;
 using System.Reflection;
+using System.Threading.Tasks;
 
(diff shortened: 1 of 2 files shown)


dtillman (approved):
You checked that the result is unwrapped from Task higher up in the call graph?  If so, LGTM.

kenswan:
> You checked that the result is unwrapped from Task higher up in the call graph? If so, LGTM.

Yes, everything checks out. All paths lead back to [AbstractMessageListenerAdapter.cs - HandleResult method](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Messaging/src/RabbitMQ/Listener/Adapters/AbstractMessageListenerAdapter.cs#L182) where the return type of Task is evaluated. There was also an existing test that verified intended use of returning a Task - [InvocableHandlerMethodTest.cs - Line 250 (HandleAsyncStringMethod)](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Messaging/test/MessagingBase.Test/Handler/Invocation/InvocableHandlerMethodTest.cs#L250)

PR #688, opened 2021-06-09
Add temp file/directory that can cleanup themselves

ccheetham:
Couple of utility classes that can cleanup after themselves, thus simplifying temp file usage.

Sample:
```
    using var tempFile = new TempFile();
```
A temp file is created during construction and subsequently deleted when the var goes out of scope.

Similar behavior for `TempDirectory`.


ccheetham:
Usurped by PR #693

Issue #532, opened 2020-11-30
Metrics observers - reduce the number of default metrics configured 


hananiel:
Currently by default we add all available metrics from the runtime and observers. This creates an issue with someone getting started with metrics. Reduce this to a few well known and expected metrics and make others configurable.

TimHess:
Toggles already exist at the observer level, and only [three out of seven observers](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/src/EndpointBase/Metrics/MetricsObserverOptions.cs#L51) are enabled by default.

There is currently no way to toggle which metrics are tracked within a given observer, but most of the observers only record a couple metrics. `EventCounterListener` is the only exception, reporting over 20 metrics

PR #699, opened 2021-06-23
Upgrade Open Telemetry

TimHess:
- Includes a copy of the Open Telemetry metrics implementation from around 0.2.0-alpha
- Updates to OpenTelemetry 1.1.0-rc1 for core components and starts using 1.0.0-rc5 for instrumentation
- Adds a way to exclude specific System.Runtime metrics (but does not currently exclude any) #532


src/Management/src/TracingBase/TracingBaseServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,84 @@
...
+                            options.Filter += req => !pathMatcher.IsMatch(req.RequestUri.PathAndQuery);
+                        });
+
+                    if (traceOpts.PropagationType.Equals("B3", StringComparison.InvariantCultureIgnoreCase))
+                    {
+                        var propagators = new List<TextMapPropagator> { new B3Propagator(traceOpts.SingleB3Header), new BaggagePropagator() };
+                        Sdk.SetDefaultTextMapPropagator(new CompositeTextMapPropagator(propagators));
+                    }
+
+                    // To Discuss: Should these remain Steeltoe options or downstream set directly?
```
TimHess, line 67: Samplers can be set via the exposed `TracerProviderBuilder`, should we keep the config-based option?

src/Management/src/OpenTelemetryBase/Steeltoe.Management.OpenTelemetryBase.csproj
```diff
@@ -15,11 +13,10 @@
 
   <ItemGroup>
     <PackageReference Include="OpenTelemetry" Version="$(OpenTelemetryVersion)" />
-    <PackageReference Include="OpenTelemetry.Exporter.Zipkin" Version="$(OpenTelemetryVersion)" />
```
TimHess, line 18: Dropped this to avoid forcing a dependency on Zipkin package... Autoconfig can still bootstrap Zipkin, but any other exporter would require tracing to be manually setup.

Side affect is UseShortTraceId only affects log processor now

src/Management/src/OpenTelemetryBase/Trace/ITracingOptions.cs
```diff
@@ -6,24 +6,54 @@ namespace Steeltoe.Management.OpenTelemetry.Trace
...
+        /// <remarks>Default value: "/cloudfoundryapplication/.*|.*\\.png|.*\\.css|.*\\.js|.*\\.html|/favicon.ico|/hystrix.stream|.*\\.gif"</remarks>
         string IngressIgnorePattern { get; }
 
+        /// <summary>
+        /// Gets a regex pattern for requests leaving this application that should not be traced
+        /// </summary>
+        /// <remarks>Default value: "/api/v2/spans|/v2/apps/.*/permissions|/eureka/*"</remarks>
         string EgressIgnorePattern { get; }
 
-        int MaxNumberOfAttributes { get; }
```
hananiel, line 15: So there is no way to set these anymore - or were they enver used?

TimHess: They _were_ used [here](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/src/OpenTelemetryBase/Trace/OpenTelemetryTracing.cs#L23), but I don't think they're available options anymore. Searching for them in the OTel repo comes back 0 results of any kind, and the `TracerConfiguration` class is gone

src/Management/src/TracingBase/TracingLogProcessor.cs
```diff
@@ -11,13 +11,11 @@ namespace Steeltoe.Management.Tracing
...
         private readonly ITracingOptions _options;
 
-        public TracingLogProcessor(ITracingOptions options, ITracing tracing)
+        public TracingLogProcessor(ITracingOptions options)
         {
             _options = options;
-            _tracing = tracing;
         }
 
         public string Process(string inputLogMessage)
```
hananiel, line 21: Does the below TODO still apply? that is, OTEL api still has this limitation?

TimHess: the context available here doesn't make it obvious if it is available... I have a recollection of reading that it still wasn't available, but I can look again

hananiel (approved):
LGTM

src/Management/src/TracingBase/TracingLogProcessor.cs
TimHess: ParentId is available on `Activity` but not anything Span related, so unless we switched over to reading the Activity.Current this TODO still applies

src/Management/src/OpenTelemetryBase/Steeltoe.Management.OpenTelemetryBase.csproj
jkonicki: It seems to make sense, but I don't have a full grasp of the implications.  Seems like we can document the manual setup?  Do you agree @TimHess?   @hananiel Do you have any thoughts on this?

TimHess: TracingBase can now auto-wire 3 different OTel exporters, without forcing the nuget dependency

hananiel (approved):

hananiel (approved):

Issue #705, opened 2021-06-25
Possible exception on configuration reload - Collection was modified; enumeration operation may not execute.


mzakrzewski:
## Describe the bug
Same thing as #704 but for the CloudFoundryConfigurationProvider

https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Configuration/src/CloudFoundryBase/CloudFoundryConfigurationProvider.cs

## Expected behavior
I would expect a call to reload configuration would not throw an invalid operation exception because a collection has been modified.

## Environment (please complete the following information):
Steeltoe Version: 3.0.2
Platform: Cloud Foundry
OS: Linux
.NET Version: .net core 3.1.13
Any other library versions to note: N/A

PR #709, opened 2021-06-29
Cf config provider atomic reload

mzakrzewski:
Should address #705 with a unit test to prevent regressing


TimHess (approved):
LGTM, thanks @mzakrzewski!

Issue #703, opened 2021-06-25
AddCloudFoundryIdentityCertificate on authentication builder is missing scheme name overload


macsux:
AddCloudFoundryIdentityCertificate on authentication builder is missing scheme name overload. This is expected when juggling multiple schemes and is a standard pattern for authentication registration pipelines.

https://github.com/SteeltoeOSS/Steeltoe/blob/fc5ebe963718ef922c32fad6895adb38ef5975af/src/Security/src/Authentication.CloudFoundryCore/AuthenticationBuilderExtensions.cs#L278

Issue #667, opened 2021-05-20
Convert IAppBuilder extensions to IHostedService(s)


TimHess:
We have many IAppBuilder extensions that may not need to do anything with the request pipeline and can/should potentially be converted to IHostedService. This could increase the number of packages that can be auto-bootstrapped

Extensions to review:
```
Steeltoe.CircuitBreaker.Hystrix.HystrixApplicationBuilderExtensions.UseHystrixMetricsStream -- x2
Steeltoe.Discovery.Client.DiscoveryApplicationBuilderExtensions.UseDiscoveryClient
Steeltoe.Management.Endpoint.SpringBootAdminClient.SpringBootAdminApplicationBuilderExtensions.RegisterWithSpringBootAdmin
Steeltoe.Security.Authentication.Mtls.CertificateApplicationBuilderExtensions.UseCertificateRotation

PR #714, opened 2021-07-01
AddCloudFoundryIdentityCertificate improvements

### LINKED ISSUE #702: Calling AddCloudFoundryIdentityCertificate causes service container to be built
## Describe the bug
Calling AddCloudFoundryIdentityCertificate on ApplicationBuilder causes a service container to be built.
https://github.com/SteeltoeOSS/Steeltoe/blob/fc5ebe963718ef922c32fad6895adb38ef5975af/src/Security/src/Authentication.CloudFoundryCore/AuthenticationBuilderExtensions.cs#L280

## Expected behavior
Avoid premature container creation during its building stage. Resolution of the logger should done from the container by adding a `Configure` call for `MutualTlsAuthenticationOptions` outside of AddMutualTls call (as it's just a shorthand way to register options with values known at the container registration phase).

TimHess:
Add some missing overloads for #703
Remove usage of temporary service container #702 
Also removes IConfiguration param that we can work without
Fix for relevant piece of #667


dtillman (approved):
LGTM

src/Security/src/Authentication.CloudFoundryCore/MutualTlsAuthenticationOptionsPostConfigurer.cs
```diff
@@ -2,24 +2,50 @@
...
 
         public void PostConfigure(string name, MutualTlsAuthenticationOptions options)
         {
             options.IssuerChain = _containerIdentityOptions.CurrentValue.IssuerChain;
+            options.Events = new CertificateAuthenticationEvents()
+            {
+                OnCertificateValidated = context =>
+                {
+                    var claims = new List<Claim>(context.Principal.Claims);
+                    if (CloudFoundryInstanceCertificate.TryParse(context.ClientCertificate, out var cfCert, _logger))
```
hananiel, line 35: Tim, I did not see a test, but in running this did you see the log message showing up? I don't see why the logger wouldn't get resolved by the container, but just making sure.

hananiel (approved):

src/Security/src/Authentication.CloudFoundryCore/MutualTlsAuthenticationOptionsPostConfigurer.cs
TimHess: Yes, the log message does show up if there is a logger and cert validation fails

PR #716, opened 2021-07-08
Convert IApplicationBuilder extensions where possible

### LINKED ISSUE #667: Convert IAppBuilder extensions to IHostedService(s)

TimHess:
Resolves #667, #647
Also addresses some testing minor testing deficiencies and removes some extension method params that aren't needed

### DIFF
--- a/src/CircuitBreaker/src/Hystrix.MetricsEventsCore/HystrixServiceCollectionExtensions.cs
+++ b/src/CircuitBreaker/src/Hystrix.MetricsEventsCore/HystrixServiceCollectionExtensions.cs
@@ -4,6 +4,7 @@
 
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection.Extensions;
 using Steeltoe.CircuitBreaker.Hystrix.Config;
 using Steeltoe.CircuitBreaker.Hystrix.Metric;
 using Steeltoe.CircuitBreaker.Hystrix.Metric.Consumer;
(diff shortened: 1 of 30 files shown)


src/CircuitBreaker/src/Hystrix.MetricsEventsCore/HystrixServiceCollectionExtensions.cs
```diff
@@ -32,50 +37,67 @@ public static void AddHystrixMetricsEventSource(this IServiceCollection services
                 throw new ArgumentNullException(nameof(services));
             }
 
+            services.AddHystrixMetricsStream();
```
TimHess, line 40: Added this line (and used TryAdd above) because `HystrixEventSourceService` requires `HystrixDashboardStream`, so this should help avoid an exception without causing any problems

src/Management/src/EndpointBase/SpringBootAdminClient/SpringBootAdminClientHostedService.cs
```diff
@@ -0,0 +1,74 @@
...
+            _httpClient.Timeout = TimeSpan.FromMilliseconds(_options.ConnectionTimeoutMS);
+            var result = await _httpClient.PostAsJsonAsync($"{_options.Url}/instances", app);
+            if (result.IsSuccessStatusCode)
+            {
+                RegistrationResult = await result.Content.ReadFromJsonAsync<RegistrationResult>();
+            }
+            else
+            {
+                _logger.LogError($"Error registering with SpringBootAdmin {result}");
+            }
```
kenswan, line 61:
```suggestion
            }
            
            return Task.CompletedTask;
```

TimHess: That results in a compiler error `CS1997 Since 'SpringBootAdminClientHostedService.StartAsync(CancellationToken)' is an async method that returns 'Task', a return keyword must not be followed by an object expression. Did you intend to return 'Task<T>'`

hananiel (approved):

Issue #575, opened 2021-01-20
Make HttpTimeout Configurable for Auth


TimHess:
Follow-up on #571 by making the timeout configurable

PR #722, opened 2021-07-13
Make auth server timeouts configurable

TimHess:
Start building v3.1.1
Set auth server timeout using `security:oauth2:client:timeout` #575


src/Security/src/Authentication.CloudFoundryCore/CloudFoundryOpenIdConnectOptions.cs
```diff
@@ -41,14 +41,20 @@ public CloudFoundryOpenIdConnectOptions()
         /// </summary>
         public bool ValidateCertificates { get; set; } = true;
 
+        /// <summary>
+        /// Gets or sets the timeout (in ms) for calls to the auth server
+        /// </summary>
+        public int Timeout { get; set; } = 100000;
```
TimHess, line 47: This change (combined with line 57) is likely to result in a higher timeout for at least one interaction

kenswan (approved):

PR #724, opened 2021-07-14
Remove GitInfoContributor unnecessary value processing

### LINKED ISSUE #706: [GitInfoContributor] commit time is not human readable
The _GitInfoContributor_ transforms the `git.commit.time` into a timestamp (in general any value associated to a key named "time"). This appears to be done on purpose (see permalink below), so I don't know if it's needed for a specific case. Nonetheless the timestamp is not human readable as the ISO8601 data representation. Currently Spring does not perform this value transformation and from SpringBootAdmin this value does not seem to be processed at all. The result is that SBA shows `git.commit.time` value with different formats:

``` sh
# spring
2021-06-28T07:46:08Z
# steeltoe
1624613217000
```

In my opinion datetime values retrieved from this actuator should always be sent using a human readable format. In any case using ISO8601 format will result in a value that it's both human and machine readable. 

## Additional content
Here is the custom code that performs the value transformation:

https://github.com/SteeltoeOSS/Steeltoe/blob/9322d0627673c9c96d71b584d190284649c4005d/src/Management/src/EndpointBase/Info/Contributor/GitInfoContributor.cs#L84-L99

DaviGia:
Removes unnecessary time-related values processing from GitInfoContributor. Resolves #706.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/SteeltoeOSS/Steeltoe?pullRequest=724) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/SteeltoeOSS/Steeltoe?pullRequest=724) before we can accept your contribution.<br/><br/>:x: DaviGia [sign now](https://cla.dotnetfoundation.org/SteeltoeOSS/Steeltoe?pullRequest=724)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/SteeltoeOSS/Steeltoe?pullRequest=724) it.</sub>

TimHess:
After further review, I understand a bit more what's happening here. Steeltoe is missing a piece that Spring has, and that's configurable serialization of this value (and others). Spring converts it to epoch time and lets the serializer decide what it should look like over the wire. This piece must have just been missed originally 

Your current change is good, and we internally went back and forth on it but are thinking the time should still be parsed, and as a follow-up the serialization options should be made configurable

src/Management/src/EndpointBase/Info/Contributor/GitInfoContributor.cs
```diff
@@ -80,20 +78,5 @@ public virtual IConfiguration ReadGitProperties(string propFile)
 
             return null;
         }
-
-        protected override void AddKeyValue(Dictionary<string, object> dict, string key, string value)
-        {
-            var keyToInsert = key;
```
TimHess, line 86: How about if we bring this method back and have it look more like this instead?
```csharp
            object valueToInsert = value;

            if ("time".Equals(key))
            {
                valueToInsert = DateTime.Parse(value).ToString("yyyy-MM-ddTHH:mm:ssZ");
            }

            dict[key] = valueToInsert;
```

DaviGia: That's ok, but I have some thoughts:
- I would move the key "time" into a constant (or I would keep it similar to Spring implementation)
- I would define an array of strings as constants instead of a single string (
- I would make sure that DateTime.Parse is culture invariant to avoid incorrect parsing (the values should always be ISO8601 compliant I hope)

TimHess: Sure, that all sounds good. I think it would be fine to use a single constant rather than an array, but if you want to define it as an array that's not likely to hurt

DaviGia: I've implemented the changes we discussed. Let me know if you have any further comment.

src/Management/src/EndpointBase/Info/Contributor/GitInfoContributor.cs
```diff
@@ -15,9 +16,10 @@ public class GitInfoContributor : AbstractConfigurationContributor, IInfoContrib
     {
         private const string GITSETTINGS_PREFIX = "git";
         private const string GITPROPERTIES_FILE = "git.properties";
+        private const string DATETIME_OUTPUT_FORMAT = "yyyy-MM-ddTHH:mm:ssZ";
 
-        private static readonly DateTime BaseTime = new DateTime(1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
-
+        private readonly List<string> DATETIME_INPUT_KEYS = new List<string> { "time" };
+
```
TimHess, line 22:
```suggestion
        private static readonly List<string> DATETIME_INPUT_KEYS = new List<string> { "time" };
```
Extremely minor stuff here

TimHess (dismissed):

TimHess (approved):

TimHess:
Thanks @DaviGia

Issue #723, opened 2021-07-14
Startup failure if Kubernetes actuators are enabled


DaviGia:
## Describe the bug

Kubernetes actuators are enabled without checking whether the application is inside Kubernetes or not. This causes the following yaml parse exception at startup when I try to debug the application locally from Visual Studio: 

``` log
   at YamlDotNet.Core.ParserExtensions.Expect[T](IParser parser)
   at YamlDotNet.Serialization.NodeDeserializers.DictionaryNodeDeserializer.DeserializeHelper(Type tKey, Type tValue, IParser parser, Func`3 nestedObjectDeserializer, IDictionary result)
   at YamlDotNet.Serialization.NodeDeserializers.DictionaryNodeDeserializer.YamlDotNet.Serialization.INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func`3 nestedObjectDeserializer, Object& value)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.AliasValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.<>c__DisplayClass3_0.<DeserializeValue>b__0(IParser r, Type t)
   at YamlDotNet.Serialization.NodeDeserializers.ObjectNodeDeserializer.YamlDotNet.Serialization.INodeDeserializer.Deserialize(IParser parser, Type expectedType, Func`3 nestedObjectDeserializer, Object& value)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.AliasValueDeserializer.DeserializeValue(IParser parser, Type expectedType, SerializerState state, IValueDeserializer nestedObjectDeserializer)
   at YamlDotNet.Serialization.ValueDeserializers.NodeValueDeserializer.<>c__DisplayClass3_0.<DeserializeValue>b__0(IParser r, Type t)
[...]

PR #728, opened 2021-07-22
Update KubernetesClient, minor related enhancements

TimHess:
- Update kubernetesclient #723
   - Only applies to (new) netstandard2.1 target
- reduce usage of temporary serviceproviders

### DIFF
--- a/src/Common/src/Common.Kubernetes/Steeltoe.Common.Kubernetes.csproj
+++ b/src/Common/src/Common.Kubernetes/Steeltoe.Common.Kubernetes.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>netstandard2.0</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
     <Description>Steeltoe common library for Kubernetes</Description>
     <PackageTags>Kubernetes</PackageTags>
   </PropertyGroup>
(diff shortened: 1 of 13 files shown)


src/Common/src/Common.Kubernetes/Steeltoe.Common.Kubernetes.csproj
```diff
@@ -1,14 +1,18 @@
...
+    <TargetFrameworks>netstandard2.0;netstandard2.1</TargetFrameworks>
     <Description>Steeltoe common library for Kubernetes</Description>
     <PackageTags>Kubernetes</PackageTags>
   </PropertyGroup>
 
   <Import Project="..\..\..\..\versions.props" />
   <Import Project="..\..\..\..\sharedproject.props" />
 
-  <ItemGroup>
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
```
TimHess, line 11: Intent is there should be no changes for netstandard2.0 since this is a patch release... We won't be testing the netstandard2.0 path (but shouldn't have anything to worry about as there's no change for that scenario) and we will drop this target in a future major/minor release

src/Common/test/Common.Kubernetes.Test/IServiceCollectionExtensionsTest.cs
```diff
@@ -61,7 +63,6 @@ public void GetKubernetesApplicationOptions_ReturnsAndAddsOptions()
             // assert
             Assert.NotNull(options);
             Assert.Single(appInfos);
-            Assert.Equal(options, appInfos.FirstOrDefault());
```
TimHess, line 64: The exact same instance isn't returned because we're not building a temporary ServiceProvider in `AddKubernetesApplicationInstanceInfo` anymore

kenswan (approved):

Issue #733, opened 2021-07-26
Excessive usage of BuildServiceProvider


TimHess:
Related to #503 but outside of management, there are multiple instances where BuildServiceProvider is called and maybe doesn't need to be

- Steeltoe.Common
  - [x] IServiceCollectionExtensions
- Configuration.Abstractions
  - [x] IServiceCollectionExtensions
- Configuration.CloudFoundryBase
  - [x] IServiceCollectionExtensions
- Discovery.ClientBase
  - [ ] #735
  - [x] ConfigurationDiscoveryClientExtensions
- Discovery.Consul
  - [x] ConsulDiscoveryClientExtension
- Discovery.Eureka
  - [x] EurekaDiscoveryClientExtension
- Logging.DynamicSerilogBase
  -  [ ] SerilogLoggingBuilderExtensions - consider in #718 
- Stream.BinderRabbitMQ
  - [x] RabbitExchangeQueueProvisioner
  - [ ] #739

PR #734, opened 2021-07-26
Reduce usage of BuildServiceProvider

TimHess:
#733

### DIFF
--- a/src/Discovery/src/ClientBase/DiscoveryServiceCollectionExtensions.cs
+++ b/src/Discovery/src/ClientBase/DiscoveryServiceCollectionExtensions.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
+using Steeltoe.Common;
 using Steeltoe.Common.Discovery;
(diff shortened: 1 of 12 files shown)


src/Discovery/src/ClientBase/DiscoveryServiceCollectionExtensions.cs
```diff
@@ -52,6 +53,7 @@ public static class DiscoveryServiceCollectionExtensions
         public static IServiceCollection AddDiscoveryClient(this IServiceCollection services, IConfiguration config, string serviceName = null, IDiscoveryLifecycle lifecycle = null)
         {
             Action<DiscoveryClientBuilder> builderAction = null;
+
             config ??= services.BuildServiceProvider().GetRequiredService<IConfiguration>();
```
TimHess, line 57:
```suggestion

            // TODO: don't BuildServiceProvider() here
            config ??= services.BuildServiceProvider().GetRequiredService<IConfiguration>();
```
Should we throw an exception if config is null? Significant work may be required to avoid this call otherwise.. Or maybe it's not as much of a concern since all relevant service bindings and config should already be available? Same applies to Line 182

kenswan: I'm also thinking the config should be available as well from `ConfigureAppConfiguration` at this point

kenswan (approved):

Issue #725, opened 2021-07-14
Actuator JsonSerializerOptions should be configurable


TimHess:
Allow configuration of these options https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/src/EndpointBase/Middleware/EndpointMiddleware.cs#L55

Follow up on #724

jkonicki:
Make sure to document the options

PR #745, opened 2021-08-17
Allow configuration of JsonSerializerOptions for Actuators

TimHess:
Allow configuring the `JsonSerializerOptions` used for Actuators #725
Add a `JsonConverter` that can bring back Epoch time for gitinfo from #724

### DIFF
--- a/src/Management/src/EndpointBase/Info/EpochSecondsDateTimeConverter.cs
+++ b/src/Management/src/EndpointBase/Info/EpochSecondsDateTimeConverter.cs
@@ -1,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Steeltoe.Management.Endpoint.Info
+{
+    public class EpochSecondsDateTimeConverter : JsonConverter<DateTime>
+    {
+        private static readonly DateTime _baseTime = new (1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) => DateTime.Parse(reader.GetString());
+
+        public override void Write(Utf8JsonWriter writer, DateTime value, JsonSerializerOptions options)
+        {
+            var utc = value.ToUniversalTime();
+            var valueToInsert = (utc.Ticks - _baseTime.Ticks) / 10000;
+            writer.WriteNumberValue(valueToInsert);
+        }
+    }
+}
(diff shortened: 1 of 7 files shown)


src/Management/src/EndpointBase/Middleware/EndpointMiddleware.cs
```diff
@@ -69,24 +69,30 @@ public virtual string Serialize(TResult result)
 
             return string.Empty;
         }
+
+        internal JsonSerializerOptions GetSerializerOptions(JsonSerializerOptions serializerOptions)
+        {
+            serializerOptions ??= new JsonSerializerOptions() { PropertyNamingPolicy = JsonNamingPolicy.CamelCase };
+            serializerOptions.IgnoreNullValues = true;
+            if (serializerOptions.Converters?.Any(c => c is HealthConverter) != true)
```
hananiel, line 77: Do you need another check for the metrics converter instead of checking only for HealthConverter.

hananiel (changes requested):
Please see comments ...

src/Management/src/EndpointBase/Info/EpochSecondsDateTimeConverter.cs
```diff
@@ -0,0 +1,24 @@
...
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Steeltoe.Management.Endpoint.Info
+{
+    public class EpochSecondsDateTimeConverter : JsonConverter<DateTime>
+    {
+        private static readonly DateTime _baseTime = new (1970, 1, 1, 0, 0, 0, DateTimeKind.Utc);
+
+        public override DateTime Read(ref Utf8JsonReader reader, Type typeToConvert, JsonSerializerOptions options) => DateTime.Parse(reader.GetString());
```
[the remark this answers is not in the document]
TimHess: I could be wrong, but I don't think this would actually get used (by us)? Would it be better to throw a `NotImplementedException`?

src/Management/src/EndpointBase/Middleware/EndpointMiddleware.cs
TimHess: We could add that, but they only way for it to happen is if a developer configured adding `HealthConverter` in their app config (see src/Management/test/EndpointCore.Test/CloudFoundry/EndpointMiddlewareTest.cs line 120 for an idea what that looks like)

hananiel: So If they add a HealthConverter, then they would lose the default MetricsResponseConverter ....

TimHess: That's true, but is there any conceivable reason somebody would do that?

hananiel: No but if they did, our code should be consistent, either dont check and always add, or check each one separately and add

src/Management/src/EndpointBase/Info/EpochSecondsDateTimeConverter.cs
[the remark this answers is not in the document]
hananiel: No not a notimplementedException, typically the converter is absent, the conversion simply doesn't happen. But if its there and input is formatted it will throw a formatexception which is fine in this case .

PR #751, opened 2021-08-30
Reduce config server-related test failures

TimHess:


azure-pipelines.yml
```diff
@@ -99,7 +99,7 @@ jobs:
       projects: src/Steeltoe.All.sln
       arguments: '--no-restore -c $(buildConfiguration) -v n /p:TreatWarningsAsErrors=True /p:VersionPrefix=$(VersionPrefix) /p:VersionSuffix=$(VersionSuffix)'
   - script: |
-      docker run -d --name configserver -p 8888:8888 steeltoeoss/config-server
+      docker run -d --name configserver -p 8888:8888 steeltoeoss/config-server --spring.cloud.config.server.git.default-label=main
```
TimHess, line 102:
```suggestion
      docker run -d --name configserver -p 8888:8888 steeltoeoss/config-server
```
Added this due to change in config server image, but that issue has been resolved. Should we leave the specification of default-label just as an example, or take it back out? @ccheetham

dtillman (approved):

Issue #761, opened 2021-09-07
Fire an event when a change is detected with registered apps/instances


TimHess:
Firing events here would make it easier to use the Eureka client with things like YARP:
https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Discovery/src/Eureka/AppInfo/Applications.cs#L185-L200

cc @macsux

PR #765, opened 2021-09-07
Add event for when Eureka registry is updated

macsux:
Adds support for #761


TimHess:
Given the location of the change it's not immediately clear to me how somebody would take advantage of this change, can you provide an example of what you expect that'd look like?

src/Discovery/src/Eureka/DiscoveryClient.cs
```diff
@@ -92,6 +92,9 @@ protected DiscoveryClient(ApplicationInfoManager appInfoManager, ILoggerFactory
             _startupLogger = logFactory?.CreateLogger("Startup." + this.GetType().FullName) ?? NullLogger.Instance;
         }
 
+        public event EventHandler<Applications> OnApplicationsChange;
+
```
TimHess, line 96:
```suggestion
```

macsux:
Something like this
```csharp
public void Configure(IApplicationBuilder app, IWebHostEnvironment env, IDiscoveryClient client)
        {
            var eureka = (DiscoveryClient) client;
            eureka.OnApplicationsChange += (sender, applications) =>
            {
                Console.WriteLine("====");
                foreach (var application in applications.GetRegisteredApplications())
                {
                    Console.WriteLine($"App: {application.Name}");
                    foreach (var instance in application.Instances)
                    {
                        Console.WriteLine($"  {instance.Actiontype} {instance.HostName}:{instance.Port}");
                    }
                }
            };
```

TimHess:
This implementation should be OK for now, but we'll probably want to revisit with [this issue](https://github.com/SteeltoeOSS/Steeltoe/issues/336) to align with Spring and things like [`ApplicationEvent`](https://github.com/spring-cloud/spring-cloud-commons/blob/main/spring-cloud-commons/src/main/java/org/springframework/cloud/client/discovery/event/HeartbeatEvent.java) 

Do you have time to add a unit test for this?

TimHess (approved):

PR #780, opened 2021-09-20
Add hosted service to replace the logger factory

### LINKED ISSUE #675: Initialization loggerFactory limitations
Currently some `IHostBuilder` extensions methods (e.g. `AddKubernetesConfiguration()`, `AddConfigServer()`) accepts an optional `ILoggerFactory`.

Because this code is evaluated before the DynamicSerilogProvider is created, the loggerFactory can only be configured using a static serilog configuration, i.e. its log level cannot be changed at runtime and cannot be configured from appsettings.

``` cs
//create loggerFactory from a static configuration (not ideal)
ILoggerFactory initLogFactory = new LoggerFactory().AddSerilog(new LoggerConfiguration().MinimumLevel.Error().WriteTo.Console().CreateLogger());

return Host.CreateDefaultBuilder(args)
    .AddDynamicSerilog()
    .AddKubernetesConfiguration(null, initLogFactory) //initLogFactory is not configured with DynamicSerilog
    .AddConfigServer(initLogFactory) //initLogFactory is not configured with DynamicSerilog
    .AddDiscoveryClient()
    .AddAllActuators()
    .ConfigureWebHostDefaults(webBuilder => webBuilder.UseStartup<Startup>());
```

I would like to be able to pass loggerFactory in a lazy manner, so that I can build it when the DynamicSerilog is available. If that is not possible maybe those extension method may avoid to receive a loggerFactory as parameter, andthey could take it using DI instead (is it possible?).

Use a static Serilog `LoggerConfiguration`.

hananiel:
resolves #675

### DIFF
--- a/src/Bootstrap/src/Autoconfig/WebHostBuilderExtensions.cs
+++ b/src/Bootstrap/src/Autoconfig/WebHostBuilderExtensions.cs
@@ -205,7 +205,7 @@ namespace Steeltoe.Bootstrap.Autoconfig
         private static void WireKubernetesConfiguration(this IWebHostBuilder hostBuilder) =>
             hostBuilder
                 .ConfigureAppConfiguration(cfg => cfg.AddKubernetes(loggerFactory: _loggerFactory))
-                .ConfigureServices(serviceCollection => serviceCollection.AddKubernetesApplicationInstanceInfo())
+                .ConfigureServices(serviceCollection => serviceCollection.AddKubernetesApplicationInstanceInfo().AddHostedService<ConfigServerHostedService>())
                 .Log(LogMessages.WireKubernetesConfiguration);
 
         [MethodImpl(MethodImplOptions.NoInlining)]
--- a/src/Configuration/src/KubernetesBase/KubernetesProviderBase.cs
+++ b/src/Configuration/src/KubernetesBase/KubernetesProviderBase.cs
@@ -6,6 +6,7 @@ using k8s;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using System;
+using System.Diagnostics;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -20,6 +21,8 @@ namespace Steeltoe.Extensions.Configuration.Kubernetes
         protected KubernetesConfigSourceSettings Settings { get; set; }
 
         protected CancellationToken CancellationToken { get; set; }
+
+        protected ILogger Logger => Settings.LoggerFactory?.CreateLogger(this.GetType());
 
         internal KubernetesProviderBase(IKubernetes kubernetes, KubernetesConfigSourceSettings settings, CancellationToken token = default)
         {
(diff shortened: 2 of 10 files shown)


jkonicki:
Backport - release/3.1

src/Bootstrap/src/Autoconfig/WebHostBuilderExtensions.cs
```diff
@@ -205,7 +205,7 @@ private static IWebHostBuilder Log(this IWebHostBuilder host, string message)
         private static void WireKubernetesConfiguration(this IWebHostBuilder hostBuilder) =>
             hostBuilder
                 .ConfigureAppConfiguration(cfg => cfg.AddKubernetes(loggerFactory: _loggerFactory))
-                .ConfigureServices(serviceCollection => serviceCollection.AddKubernetesApplicationInstanceInfo())
```
TimHess, line 208: This change is probably also needed in `HostBuilderExtenions.cs`

src/Configuration/src/KubernetesBase/KubernetesProviderBase.cs
```diff
@@ -38,6 +41,15 @@ internal KubernetesProviderBase(IKubernetes kubernetes, KubernetesConfigSourceSe
             CancellationToken = token;
         }
 
+        internal void ProvideRuntimeReplacements(ILoggerFactory loggerFactory)
+        {
+            if (loggerFactory is not null)
+            {
+                Settings.LoggerFactory = loggerFactory;
+                Logger?.LogTrace("Replacing Bootstrapped loggerFactory with actual factory");
```
TimHess, line 49: Shouldn't this method replace `Logger` with a new one from the new loggerFactory? I'm not clear on why we're passing the ILoggerFactory around instead of just replacing the loggers - If I'm reading this right, we start up with 2 (total) ILoggers, but when this replacement happens we are changing out a LoggerFactory and only doing anything for the loggers if they weren't previously set

src/Bootstrap/src/Autoconfig/WebHostBuilderExtensions.cs
hananiel: Yes I made the same change there as well ...

src/Configuration/src/KubernetesBase/KubernetesProviderBase.cs
hananiel: The Logger is getter (no setter). Having that be a property with a backed value was causing its value to be captured in the scope of the lambda, which was causing the old value to be used, even though Logger was later reassigned.

src/Bootstrap/src/Autoconfig/WebHostBuilderExtensions.cs
TimHess: I was referring to [this one](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Bootstrap/src/Autoconfig/HostBuilderExtensions.cs)

src/Configuration/src/KubernetesBase/KubernetesProviderBase.cs
TimHess: Does `Logger` get replaced at runtime if a LoggerFactory is provided for the initial startup?

src/Bootstrap/src/Autoconfig/WebHostBuilderExtensions.cs
hananiel: done.

src/Configuration/src/KubernetesBase/KubernetesProviderBase.cs
hananiel: Added a test to prove out ..

TimHess (approved):

Issue #764, opened 2021-09-07
Steeltoe needs to identify any problem areas when upgrading to .NET 6


jkonicki:
We need some more test runs of our code against .NET 6

Identify any issues and create the necessary tickets for them.

TimHess:
Add extensions to WebApplicationBuilder to match existing IHostBuilder and IWebHostBuilder options (not just config server #814)

Issue #756, opened 2021-09-02
Spike to investigate using GitHub Actions


jkonicki:
Will moving to GitHub and possibly splitting up the builds to improve build times.

PR #788, opened 2021-09-27
Pipeline updates, testing for .NET 6

TimHess:
- Update tests to run .NET 6 instead of 5 #764 
- Drop .NET 5 from packages that don't do anything different between netcoreapp3.1/net5
- New builds by component/feature with path triggers for faster responses (related to #756, but still in devops)
- #787


hananiel (approved):
LGTM;

PR #803, opened 2021-10-21
Bootstrap logger

macsux:
Adds bootstrap logger that can be statically accessed for access to logging infrastructure before service container is built. It supports "upgrades" as infrastructure is loaded, primarily via call to update it with IConfiguration and after container is constructed ILoggerFactory. Any existing loggers spawned from it are seamlessly upgraded. Refactored config server to take advantage of this implementation if explicit LoggerFactory is not specified.

@hananiel you should probably look at this one as it touches stuff you recently worked on


src/Common/src/Common/Logging/UpgradableBootstrapLoggerFactory.cs
```diff
@@ -0,0 +1,217 @@
...
+                    var innerLogger = _factoryInstance.CreateLogger(categoryName);
+                    logger = new BoostrapLoggerInst(innerLogger, categoryName);
+                    _loggers.Add(categoryName, logger);
+                }
+
+                return logger;
+            }
+        }
+
+        // private static readonly BootstrapLoggerFactoryInst _bootstrapFactory = new ();
```
TimHess, line 127: What's happening with these comments?

src/Common/src/Common/Logging/UpgradableBootstrapLoggerFactory.cs
```diff
@@ -0,0 +1,217 @@
...
+using System.Collections.Generic;
+
+namespace Steeltoe.Common.Logging
+{
+    /// <summary>
+    /// Allows early utilization of log infrastructure before log config is even read. Any providers spawned are instantly switched over to
+    /// real log providers as the application utilization progresses.
+    /// This class should only be used by components start are invoke before  logging infrastructure is build (prior to service container creation)
+    /// </summary>
+    internal class UpgradableBootstrapLoggerFactory : IBoostrapLoggerFactory
```
TimHess, line 17: Need to address S3881 or ignore with simple explanation

TimHess (changes requested):

hananiel (commented):
Looks fine, pending @TimHess's comments.

dtillman (approved):
LGTM

TimHess (approved):

PR #812, opened 2021-11-19
Support .NET 6, update OTel

TimHess:
* Update projects and pipeline to build or test .NET 6 as necessary
* Include .NET 6 fixes already in `main`
* Upgrade OTel to 1.2 beta to get around restriction of < 6.0.0 for Microsoft.Extensions.Logging

### DIFF
--- a/versions.props
+++ b/versions.props
@@ -91,7 +91,7 @@
     <MySqlEFCoreVersion>5.0.0</MySqlEFCoreVersion>
     <NpgsqlEFCoreVersion>5.0.0</NpgsqlEFCoreVersion>
     <OracleEFCoreVersion>5.21.1</OracleEFCoreVersion>
-    <PomeloEFCoreVersion>5.0.0-alpha.2</PomeloEFCoreVersion>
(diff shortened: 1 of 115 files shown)


TimHess:
This PR (and a 3.1.2 release) are going to hold for a bit longer in hopes that the Open Telemetry metrics spec can be finalized in short order and a GA of OTel 1.2.0 can be released

versions.props
```diff
@@ -91,4 +93,19 @@
     <OracleEFCoreVersion>5.21.1</OracleEFCoreVersion>
     <PomeloEFCoreVersion>5.0.0-alpha.2</PomeloEFCoreVersion>
   </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net6.0'">
+    <AspNetCoreVersion>6.0.0-rc.2.*</AspNetCoreVersion>
```
hananiel, line 97: Do these need to be rcs?

TimHess: Probably not anymore, we were being held back by EF Core packages for MySQL, Postgres and Oracle

hananiel (approved):

PR #822, opened 2022-01-19
Heapdump/Stacktrace changes - Use standard EventPipe API to do these

dtillman:
Removes dependency on CLRMD code


TimHess:
The failing tests do not appear to be related to this work. Certificate auth tests in particular are already addressed in #824 

@dtillman can you take a look at the [code coverage report](https://sonarcloud.io/component_measures?id=SteeltoeOSS_steeltoe&pullRequest=822&metric=new_coverage&view=list)? 33% seems pretty low, but maybe it isn't accurate?

TimHess:
@dtillman I don't see any changes to `Steeltoe.Management.Diagnostics` besides removal of the ref from EndpointBase.csproj - should we also be removing the project from source or marking it obsolete or taking any other action?

src/Management/src/EndpointBase/HeapDump/HeapDumper.cs
```diff
@@ -50,7 +66,30 @@ public string DumpHeap()
 
         internal string CreateFileName()
         {
-            return "minidump-" + DateTime.Now.ToString("yyyy-MM-dd-HH-mm-ss") + "-live" + ".dmp";
+            if (Environment.Version.Major == 3 || "gcdump".Equals(_options.HeapDumpType, StringComparison.OrdinalIgnoreCase))
+            {
+                return "gcdump-" + DateTime.Now.ToString("yyyy-MM-dd-HH-mm-ss") + "-live" + ".gcdump";
```
TimHess, line 71: Has this scenario been tested on a recent version of Apps Manager? I'm not 100% sure what the current state of file extension handling is here

dtillman: Tested, and as expected the Apps Manager renames the file incorrectly and VS2022 does not open the file correctly. If the file is renamed to include .gcdump extension, then all works as expected

TimHess (dismissed):

TimHess (approved):

PR #848, opened 2022-05-12
Make Steeltoe Metrics extensions play nice with Otel

hananiel:
Working draft for a way to make Otel dependency more explicit:

Now to add steeltoe actuators, you would need

```csharp
builder.AddAllActuators();
builder.Services.AddOpenTelemetryMetricsForSteeltoe((provider, builder) => { builder.AddAspNetCoreInstrumentation() customize})
```
if you want just wavefront exporter
```csharp
builder.AddWavefrontMetrics();
builder.Services.AddOpenTelemetryMetricsForSteeltoe((provider, builder) => { builder.AddAspNetCoreInstrumentation() customize})
```
or Use otel extensions
```csharp

builder.Services.AddOpenTelemetryMetrics(b =>
   {
        b.ConfigureSteeltoeMetrics();

       b.Configure((provider, deferredBuilder) =>
       {
           deferredBuilder.AddMeter("Another Meter")
           .AddHttpClientInstrumentation()
           .AddConsoleExporter()
              .Build();
       });
   });
```

### DIFF
--- a/src/Management/src/EndpointBase/Metrics/ServiceCollectionExtensions.cs
+++ b/src/Management/src/EndpointBase/Metrics/ServiceCollectionExtensions.cs
@@ -5,6 +5,7 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection.Extensions;
 using Microsoft.Extensions.Hosting;
+using OpenTelemetry;
 using OpenTelemetry.Metrics;
 using Steeltoe.Common.Diagnostics;
 using Steeltoe.Management;
@@ -15,6 +16,7 @@ using Steeltoe.Management.OpenTelemetry.Exporters;
 using Steeltoe.Management.OpenTelemetry.Metrics;
 using System;
 using System.Diagnostics;
+using System.Linq;
 
 namespace Microsoft.Extensions.DependencyInjection
 {
@@ -55,7 +57,6 @@ namespace Microsoft.Extensions.DependencyInjection
                 return new SteeltoeExporter(exporterOptions);
             }));
 
-            services.AddOpenTelemetryMetricsForSteeltoe();
             return services;
         }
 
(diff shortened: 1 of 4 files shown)


src/Management/src/EndpointBase/Metrics/ServiceCollectionExtensions.cs
```diff
@@ -55,7 +57,6 @@ public static IServiceCollection AddMetricsActuatorServices(this IServiceCollect
                 return new SteeltoeExporter(exporterOptions);
             }));
 
-            services.AddOpenTelemetryMetricsForSteeltoe();
```
TimHess, line 58: So we're going to move from configuring OTel by default to requiring explicit config (and double-checking that it wasn't added previously)... That is good for the original complaint, but (unless I'm missing something) doesn't cover the "please make it easy/automatic for me" use case. Could these methods also have some kind of "do it for me if it's MIA" option?

src/Management/src/OpenTelemetryBase/Exporters/Wavefront/WavefrontExporterOptions.cs
```diff
@@ -24,11 +24,11 @@ public WavefrontExporterOptions(IConfiguration config)
 
         public string ApiToken { get; set; }
 
-        public int Step { get; set; } = 30000; // milliseconds
+        public int Step { get; set; } = 30_000; // milliseconds
```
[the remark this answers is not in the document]
bart-vmware: https://stackoverflow.com/a/43476057

src/Management/src/EndpointBase/Metrics/ServiceCollectionExtensions.cs
hananiel: I'll add a AddMetricsActuatorServicesWithOpenTelemetry that will do it for you.

hananiel: The main method we have in this area for actuators is addallactuators and the surprise was that it prevented you from adding opentelemetry later (because it seems unrelated). So I dont think we have a good recourse.

src/Management/src/EndpointBase/Metrics/ServiceCollectionExtensions.cs
```diff
@@ -87,31 +91,60 @@ public static IServiceCollection AddPrometheusActuatorServices(this IServiceColl
                 var exporterOptions = new PullmetricsExporterOptions() { ScrapeResponseCacheDurationMilliseconds = options.ScrapeResponseCacheDurationMilliseconds };
                 return new SteeltoePrometheusExporter(exporterOptions);
             }));
-            services.AddOpenTelemetryMetricsForSteeltoe();
```
TimHess, line 90: Is this one still supposed to be getting removed?

src/Management/src/EndpointBase/Metrics/ServiceCollectionExtensions.cs
```diff
@@ -87,31 +91,60 @@ public static IServiceCollection AddPrometheusActuatorServices(this IServiceColl
...
 
             return services;
         }
 
-        public static IServiceCollection AddOpenTelemetryMetricsForSteeltoe(this IServiceCollection services, string name = null, string version = null)
+        /// <summary>
+        /// Helper method to configure opentelemetry metrics. Do not use in conjuction with Extension methods provided by Opentelemetry.
```
TimHess, line 99:
```suggestion
        /// Helper method to configure OpenTelemetry metrics. Do not use in conjunction with Extension methods provided by OpenTelemetry.
```

TimHess (approved):

PR #849, opened 2022-05-16
Fixed a few tests that didn't assert on the outcome

bart-vmware:
While looking at warnings from Resharper, I found some tests that didn't assert on the outcome.

After correcting these tests, some of them failed. I've adapted the target logic accordingly to make the tests succeed without changing their assumptions.

Note this is a breaking change. If taking a breaking change is not an option, we should adapt the tests instead to match the current behavior.

### DIFF
--- a/src/Common/src/Common/Util/MimeType.cs
+++ b/src/Common/src/Common/Util/MimeType.cs
@@ -144,7 +144,7 @@ namespace Steeltoe.Common.Util
             if (PARAM_CHARSET.Equals(attribute))
             {
                 value = Unquote(value);
-                Encoding.GetEncoding(value);
+                _ = Encoding.GetEncoding(value);
             }
             else if (!IsQuotedstring(value))
             {
--- a/src/Management/test/EndpointBase.Test/Metrics/Observer/HttpClientDesktopObserverTest.cs
+++ b/src/Management/test/EndpointBase.Test/Metrics/Observer/HttpClientDesktopObserverTest.cs
@@ -74,10 +74,11 @@ namespace Steeltoe.Management.Endpoint.Metrics.Observer.Test
 
             var req = GetHttpRequestMessage();
             var labels = observer.GetLabels(req, HttpStatusCode.InternalServerError).ToList();
-            labels.Contains(KeyValuePair.Create("clientName", (object)"localhost:5555"));
-            labels.Contains(KeyValuePair.Create("uri", (object)"/foo/bar"));
-            labels.Contains(KeyValuePair.Create("status", (object)"500"));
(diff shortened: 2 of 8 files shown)


src/Common/src/Common/Util/MimeType.cs
```diff
@@ -144,7 +144,7 @@ protected void CheckParameters(string attribute, string value)
             if (PARAM_CHARSET.Equals(attribute))
             {
                 value = Unquote(value);
-                Encoding.GetEncoding(value);
+                _ = Encoding.GetEncoding(value);
```
hananiel, line 147: Probably ignore this suggestion - I dont see value ...

hananiel (approved):
LGTM. See other comment.

src/Common/src/Common/Util/MimeType.cs
bart-vmware: This suppresses the R# warning that the return value of a pure function is unused, which is usually an indication of a mistake. In this case however, I believe the try/catch block is used to detect failure, whereas the actual returned encoding is irrelevant here.

Adding the `_` (discard) expresses that ignoring the return value is intentional. By being explicit about that, I believe this improves code readability: Future readers won't have to wonder whether this was intentional.

Based on the above, are you okay with keeping this?

hananiel: Ok sounds good.

src/Management/test/EndpointBase.Test/Metrics/Observer/HttpClientDesktopObserverTest.cs
```diff
@@ -74,10 +74,11 @@ public void GetTagContext_ReturnsExpected()
 
             var req = GetHttpRequestMessage();
             var labels = observer.GetLabels(req, HttpStatusCode.InternalServerError).ToList();
-            labels.Contains(KeyValuePair.Create("clientName", (object)"localhost:5555"));
-            labels.Contains(KeyValuePair.Create("uri", (object)"/foo/bar"));
-            labels.Contains(KeyValuePair.Create("status", (object)"500"));
-            labels.Contains(KeyValuePair.Create("method", (object)"GET"));
+
+            Assert.Contains(KeyValuePair.Create("clientName", (object)"localhost:5555"), labels);
+            Assert.Contains(KeyValuePair.Create("uri", (object)"/foo/bar"), labels);
```
hananiel, line 79: This is the correct assertion. And the code change is correct.

hananiel:
> Note this is a breaking change. If taking a breaking change is not an option, we should adapt the tests instead to match the current behavior.
This fixes the uri to be compatible with the way spring-boot apps report uri. So shouldnt be a problem. Go ahead and merge.

PR #850, opened 2022-05-16
Fixed: Incorrect stack trace on thrown exception

bart-vmware:
The conversion method should return the exception and let the caller throw, so the stack trace is preserved.


dtillman (approved):
Thanks

PR #854, opened 2022-05-24
Code quality cleanup

bart-vmware:
This PR addresses the low-hanging fruit from Resharper warnings and suggestions. I've carefully reviewed every change to ensure it's correct and does not introduce breaking changes.

I recommend reviewing per commit.

Update: After switching to file-scoped namespaces, I had to update the SonarCube analyzer as well. After that, I started fixing Sonar warnings, but decided to take those out and create a separate PR for them.


ccheetham (dismissed):
lgtm

hananiel (dismissed):
It's all good - may be we should all use resharper. I do think resharper make VS a tad bit slow - havent used it in about a year or so.  We can discuss later.

src/CircuitBreaker/src/HystrixBase/HystrixBaseOptions.cs
```diff
@@ -45,9 +45,9 @@ protected virtual long GetLong(string prefix, string key, string property, long
         protected virtual string GetString(string prefix, string key, string property, string globalDefault, string instanceDefaultFromCode)
         {
             var result = globalDefault;
-            result = (_dynamic != null) ? _dynamic.GetString($"{prefix}:default:{property}", result) : result; // dynamic global default
```
TimHess: While the parentheses in these expressions are all technically redundant, some of them (IMO) improve readability, such as these in ternary expressions. Maybe the better option for those cases would be to write them out in the longer form, but that is probably a separate discussion to be had

bart-vmware: Thanks for expressing this.

StyleCop, Roslyn, as well as Resharper/Rider provide automated handling of adding/removing parentheses in expressions to improve readability, of which Resharper/Rider provides the most flexible configuration:

image

image

image

Above are the default settings, which make total sense to me: `a || b && c` is definitely confusing without parentheses, as well as bitwise operations, because their precedence rules are not well-known for most C# developers. Some folks believe that parenthesis around arithmetic terms improve readability as well, for example: `a + (b * c)`, which I find odd because we've all learned those precedence rules at elementary school.

Note that even Resharper/Rider does not support putting parentheses around terms in ternary expressions. I'm not aware of any analyzer for that, so I don't think there's anything we can do to enforce/normalize that, except to turn off automated rule-based analysis and handle it manually. Frankly I've never seen or heard the concern of parenthesis around ternary terms being raised before.

But once we configure a maximum line length, Resharper breaks ternary expressions as below, which you may like better:
```c#
var result = SomeLongExpressionThatExceedsMaximumLineLength
    ? ifTrueExpression
    : ifFalseExpression
```

TimHess (approved):
LGTM

PR #862, opened 2022-06-03
RabbitMQ Test Fixes

dtillman:
This PR includes two changes to Unit tests that improve reliability of those tests.   
Also includes a fix which was causing a process to hang during RabbitMQ shutdown.


dtillman:
This PR should be backported to 3.2.1

TimHess (approved):

PR #871, opened 2022-06-19
Sonar rules cleanup

bart-vmware:
## Description

{A detailed description}

Fixes #{bug number}

## Quality checklist

- [ ] Your code complies with our [Coding Style](https://github.com/SteeltoeOSS/.github/blob/main/contributing-docs/contributing-code-style.md).
- [ ] You've updated unit and/or integration tests for your change, where applicable.
- [ ] You've updated documentation for your change, where applicable.
      If your change affects other repositories, such as [Documentation](https://github.com/SteeltoeOSS/Documentation), [Samples](https://github.com/SteeltoeOSS/Samples) and/or [MainSite](https://github.com/SteeltoeOSS/MainSite), add linked PRs here.
- [ ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.
- [ ] You've added required license files and/or file headers (explaining where the code came from with proper attribution), where code is copied from StackOverflow, a blog, or OSS.

PR #873, opened 2022-06-27
Drop netstandard and update package versions

bart-vmware:
## Description

This PR changes the target frameworks to .NET 6 and .NET Core 3.1 and updates NuGet dependencies to the latest versions, except for the next ones which are incompatible:
- [KubernetesClient](https://www.nuget.org/packages/KubernetesClient/) (KubernetesClientVersion)
- [RabbitMQ.Client](https://www.nuget.org/packages/RabbitMQ.Client) (RabbitClientVersion)

All package versions are now centralized in `versions.props`, which has been split up into exposed and non-exposed dependencies.

See the individual commits for details.

~**NOTE:** This PR temporarily targets the **sonar-cleanup** branch, to make reviewing the diff easier. Once #871 has been merged, I'll update the target branch of this PR to main.~

## Quality checklist

- [ ] N/A: You've updated documentation for your change, where applicable.
- [ ] N/A: There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.
- [ ] N/A: You've added required license files and/or file headers (explaining where the code came from with proper attribution), where code is copied from StackOverflow, a blog, or OSS.


src/Security/src/DataProtection.CredHubBase/Credentials/Utilities/SetRequestJsonConverter.cs
```diff
@@ -21,7 +21,9 @@ public override void Write(Utf8JsonWriter writer, CredentialSetRequest value, Js
         writer.WriteStartObject("value");
         foreach (var prop in value.Value.GetType().GetProperties())
         {
+#pragma warning disable SYSLIB0020 // Type or member is obsolete
```
TimHess, line 24: What is obsolete here?

versions.props
```diff
@@ -2,103 +2,82 @@
...
-    <MySqlConnectorVersion>0.49.3</MySqlConnectorVersion>
-    <MySqlV8>8.0.23</MySqlV8>
-    <NpgsqlVersion>4.0.0</NpgsqlVersion>
-    <SqlClientVersion>4.8.1</SqlClientVersion>
-    <StackExchangeVersion>2.0.593</StackExchangeVersion>
+    <!--
+      Exposed dependencies, observable by Steeltoe consumers.
+      Be cautious with updates:
+      - Wildcards are only allowed when starting work on the next major version, until its release.
+      - Before release of the next major version, pin these down and never change them between minor/patch releases.
```
TimHess, line 48: This is a great comment and I like this strategy

src/Management/src/TracingBase/TracingBaseServiceCollectionExtensions.cs
```diff
@@ -86,7 +86,10 @@ public static IServiceCollection AddDistributedTracing(this IServiceCollection s
 
                 if (traceOpts.PropagationType.Equals("B3", StringComparison.InvariantCultureIgnoreCase))
                 {
+                    // TODO: Investigate alternatives and remove suppression.
+#pragma warning disable CS0618 // Type or member is obsolete
```
TimHess, line 90: FYI @hananiel, looks like we need to look at B3 headers and see where things are going

TimHess (approved):
LGTM, kind of surprised by some of the small silly errors fixed here (like the extra backup .csproj file)

versions.props
bart-vmware: Glad to hear. I also thought about renaming the first category to `*TestVersion` to make the distinction more explicit. That would make it less likely to accidentally get them mixed up. For example, assume an existing lib in the future needs to reference `HealthChecksTestVersion`, the naming should signal to add `HealthChecksVersion` instead of reuse `HealthChecksTestVersion`. But its still somewhat brittle.

Another approach would be to have test/lib versions in separate files, then reference the lib-versions from project-props and both from test-props.

src/Management/src/OpenTelemetryBase/Exporters/Prometheus/PrometheusSerializer.cs
```diff
@@ -26,25 +23,9 @@ internal static partial class PrometheusSerializer
...
-        var result = value.TryFormat(span, out var cchWritten, "G", CultureInfo.InvariantCulture);
-        Debug.Assert(result, $"{nameof(result)} should be true.");
-
-        for (int i = 0; i < cchWritten; i++)
-        {
-            buffer[cursor++] = unchecked((byte)span[i]);
-        }
-#else
             cursor = WriteAsciiStringNoEscape(buffer, cursor, value.ToString(CultureInfo.InvariantCulture));
-#endif
```
bart-vmware, line 47: Looking at this again, I suspect I may have taken the wrong turn here (same in `PullmetricsCollectionManager`). Although technically correct (because the project targeted `netstandard20` initially, making the `true`-blocks effectively dead code), now it seems that the `false` block, which optimizes for `Span<T>` usage, is the preferred path to keep.

TimHess: @hananiel can you take a look at this? it looks like we should be using the code under `#if NETCOREAPP3_1_OR_GREATER`, but that code may not have been getting used before since the project was targeting netstandard 2.0

bart-vmware: Resolving this conversation to unblock for PR merge. We can revert the change later in a separate PR if needed.

PR #883, opened 2022-07-07
Update Healthcheck actuator to be compatible with ALV 

hananiel:
## Description

Make Health check endpoint compatible with latest version of SpringBoot endpoint to be compatible with App Live View on TAP. 

Fixes #{bug number}

## Quality checklist


src/Common/src/Abstractions/Steeltoe.Common.Abstractions.csproj
```diff
@@ -11,5 +11,6 @@
 
   <ItemGroup>
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="$(ExtensionsVersion)" />
+    <PackageReference Include="System.Text.Json" Version="6.0.4" />
```
TimHess, line 14: I'm not sure about bringing System.Text.Json and the Json attributes into this package... the abstraction is supposed to be transport agnostic. For example, would this affect potential to add support for using Spring Boot Admin over gRPC?

TimHess:
This PR will be part of 4.0, we'll need to backport to `release/3.2` for inclusion in 3.2.1

src/Common/src/Abstractions/Steeltoe.Common.Abstractions.csproj
bart-vmware: Note that #873 introduces a centralized version placeholder, so please use `$(SystemVersion)` once it has been merged.

hananiel: > I'm not sure about bringing System.Text.Json and the Json attributes into this package... the abstraction is supposed to be transport agnostic. For example, would this affect potential to add support for using Spring Boot Admin over gRPC?



> I'm not sure about bringing System.Text.Json and the Json attributes into this package... the abstraction is supposed to be transport agnostic. For example, would this affect potential to add support for using Spring Boot Admin over gRPC?

In 4.0, this is going to have to be completely redesigned. By exposing types that library users will need to implement health checks (ex HealthCheckResult) and statically typing it to the response of HealthEndpoint, AND using the same object for serialization, we are exposing too much to safely make changes in a backwards compatible way. It is hard to make a change without a breaking change. Considering the situation we are in my suggestion is take this dependency for now. When we do implement gRPC IMO the best thing would be to have separate packages for gRPC & Http.

bart-vmware:
@hananiel This PR has merge conflicts. Can you update?

PR #884, opened 2022-07-13
Make HealthCheck compatible with ALV

### LINKED ISSUE #877: Make HealthEndpoint compatible with latest SpringBoot version
The latest versions of Spring Boot actuators have a newer schema application/vnd.spring-boot.actuator.v3+json which Steeltoe needs to support. 
New Functionality Needs to be backwards compatible and hidden via feature toggle exposed by configuration.

hananiel:
Resolves #877


src/Management/src/EndpointBase/Health/HealthConverter.cs
```diff
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+using Steeltoe.Common.HealthChecks;
```
TimHess, line 5:
```suggestion
using Steeltoe.Common.HealthChecks
```

src/Management/src/EndpointBase/Health/HealthConverterV3.cs
```diff
@@ -0,0 +1,46 @@
...
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Text.Json;
+using System.Text.Json.Serialization;
+
+namespace Steeltoe.Management.Endpoint.Health
+{
+    public class HealthConverterV3 : JsonConverter<HealthEndpointResponse>
```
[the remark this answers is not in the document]
hananiel: We support json serializer settings in IManagementOptions that allow you set it. I was looking for initial feedback before adding unit tests and committing to this path.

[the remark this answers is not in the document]
TimHess: OK, looking forward to some tests that help clarify how it all fits together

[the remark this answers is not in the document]
hananiel: I added a test for V3

src/Management/test/EndpointCore.Test/Health/EndpointMiddlewareTest.cs
```diff
@@ -149,6 +149,32 @@ public async Task HealthActuator_ReturnsDetails()
             Assert.Contains("diskSpace", health["details"].ToString());
         }
 
+        [Fact]
+        public async Task HealthActuatorV3_ReturnsDetails()
+        {
+            var settings = new Dictionary<string, string>(appSettings) { { "management:endpoints:customjsonconverters:0", typeof(HealthConverterV3).FullName} };
```
TimHess, line 155: So if I want my app to use the v3 health response, I need to use app settings (as expected) - is this what it looks like in pure json?
```json
{
  "Management": {
    "Endpoints": {
      "CustomJsonConverters": [
        "Steeltoe.Management.Endpoint.Health.HealthConverterV3"
      ]
    }
  }
}
```

TimHess: Thinking about this from the perspective of needing additional response changes very soon (mappings at least, right?) do we want to add a shorthand property for something like "preferred response version" that can default to v2 here but be v3 in Steeltoe 4.0 so that each converter wouldn't need to be specified? Maybe excessive, but it would be nicer to have one small change than need to copy/paste `n` converters in here if it's not a huge deal

hananiel: I am of the thought that we best do that in  v4 that doesn't need backwards compatibility. In 3.2, adding any additional properties to this interface (IManagementOptions) is a breaking change. It's a major refactor to get this option into where in the code its needed - lots of public API that would need changed - so this is what I finally came up with.

TimHess: Ok on doing it as-is and making it easier in v4 - is the json I typed out accurate?

hananiel: Yes the json is correct.

hananiel:
> Are any additional changes required for `HealthRegistrationsAggregator` to work with the v3 response format?
No additional changes for healthRegistrationsAggregator to work. No matter the aggregator the response is a `HealthCheckResult`

TimHess (dismissed):
LGTM

TimHess (approved):

Issue #827, opened 2022-02-18
Consider an IEndpointRouteBuilderExtensions.MapAllActuators that takes no parameters


hananiel:
Currently MapAllActuators forces you to pass an Action for ConventionBuilder or null. See [SO issue](https://stackoverflow.com/questions/71163854/why-is-endpoints-mapallactuators-failing)

stamahto:
I would also appreciate it 👍

bart-vmware:
This was implemented in #1155.

bart-vmware:
It's in the 4.x line, which hasn't been released yet.

PR #889, opened 2022-07-20
drop .NET Core 3.1

### LINKED ISSUE #874: RFC: Steeltoe 4.0 support for .NET Core 3.1 - Ends July 15th
## Summary
As you may be aware, Microsoft plans to [end support for .NET Core 3.1 on December 13, 2022](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core). Development on Steeltoe 4.0 has recently begun and may not ship before this date, so we are considering proactively dropping support for this release. We plan to finalize a decision by July 15th, please share any comments before then.

No changes are being proposed for other Steeltoe release lines at this time.

## Why make this move?
In order to reduce the maintenance burden on contributors, Steeltoe [follows the life cycle of .NET runtimes and Microsoft’s support policies](https://github.com/SteeltoeOSS/Steeltoe/wiki/Steeltoe-Support-Versions#net-runtime-support). Additionally, once Microsoft stops performing security patches for a given runtime, it is in all of our best interest to move to a newer runtime.

## Why now?
Steeltoe 4.0 is a great time to make this move.  We will be performing significant code modernization efforts that will require breaking changes. Changing target runtimes during this release cycle allows us to take advantage of new features and functionality in the runtime right now. Additionally, this change allows users of Steeltoe to continue to develop against the latest and most feature-rich releases from .NET.

## What can you do?
We would love to hear from you!  If you have any questions or concerns about this change, please leave your comments below.  
We welcome you to bring your opinions in issues, contribute code, request features, and join the discussions.  Feel free to bring up any Steeltoe related questions or requests to our [Steeltoe Slack Channel](https://slack.steeltoe.io) or join the [Community Call](https://steeltoe.io/community) (the next opportunity is July 12th).

Thank you,

The Steeltoe Team

TimHess:
## Description

This PR removes .NET Core 3.1

Fixes #874 

## Quality checklist


src/Configuration/test/SpringBootCore.Test/SpringBootHostBuilderExtensionsTest.cs
```diff
@@ -95,4 +96,45 @@ public void GenericHostConfiguresIConfiguration_CmdLine()
         Assert.NotNull(config["spring:cloud:stream:bindings:input:group"]);
         Assert.Equal("testGroup", config["spring:cloud:stream:bindings:input:group"]);
     }
+
+    [Fact]
+    public void ConfigureSpringBoot_ThrowsIfNulls_Net60()
```
bart-vmware, line 101: Perhaps drop the `_Net60` from the test name?

src/Connectors/test/Connector.EFCore.Test/EntityFrameworkCoreTypeLocatorTest.cs
```diff
@@ -60,4 +60,30 @@ public void Options_Found_In_Devart_Assembly()
...
+    {
+        // arrange ~ narrow the assembly list to one specific nuget package
+        var types = EntityFrameworkCoreTypeLocator.MySqlEntityAssemblies;
+        EntityFrameworkCoreTypeLocator.MySqlEntityAssemblies = new[] { "MySql.EntityFrameworkCore" };
+
+        var type = EntityFrameworkCoreTypeLocator.MySqlDbContextOptionsType;
+
+        Assert.NotNull(type);
+        EntityFrameworkCoreTypeLocator.MySqlEntityAssemblies = types;
+    }
```
bart-vmware, line 88: This test used to be netcore31 only. I inferred that we switched to a different third-party package for net60+, but looking at this, my assumption may be incorrect. Is this something that should work on net60 as well?

src/Connectors/test/Connector.EFCore.Test/MySqlDbContextOptionsExtensionsTest.cs
```diff
@@ -3,17 +3,17 @@
...
 using Steeltoe.Extensions.Configuration.CloudFoundry;
 using Xunit;
-#if NET6_0_OR_GREATER
 using MySqlConnectionAlias = MySqlConnector.MySqlConnection;
-#else
-using MySqlConnectionAlias = MySql.Data.MySqlClient.MySqlConnection;
-#endif
+
+// using MySqlConnectionAlias = MySql.Data.MySqlClient.MySqlConnection
+
```
bart-vmware, line 16: Can this be removed?

bart-vmware: nit: `MySqlConnectionAlias` can be inlined now.

src/Management/src/EndpointCore/Extensions/ServiceCollectionExtensions.netcore31.cs
```diff
@@ -1,34 +0,0 @@
...
-
-public static class ServiceCollectionExtensions
-{
-    /// <summary>
-    /// Registers mapping for an endpoint. This gets used when calling <see cref="ActuatorRouteBuilderExtensions.MapAllActuators(AspNetCore.Routing.IEndpointRouteBuilder, MediaTypeVersion)"/>.
-    /// </summary>
-    /// <typeparam name="TEndpoint">The type of endpoint.</typeparam>
-    /// <param name="services">The service collection.</param>
-    /// <returns>A reference to the service collection.</returns>
-    public static IServiceCollection AddActuatorEndpointMapping<TEndpoint>(this IServiceCollection services)
```
TimHess, line 20: Is it a bad idea to be _removing_ these thing that are .net core 3.1 only? Could/should we handle this situation differently?

src/Management/src/TracingCore/TracingHostBuilderExtensions.cs
```diff
@@ -26,4 +27,15 @@ public static IHostBuilder AddDistributedTracingAspNetCore(this IHostBuilder hos
     /// <param name="action">Customize the <see cref="TracerProviderBuilder" />.</param>
     public static IWebHostBuilder AddDistributedTracingAspNetCore(this IWebHostBuilder hostBuilder, Action<TracerProviderBuilder> action = null)
         => hostBuilder.ConfigureServices((_, services) => services.AddDistributedTracingAspNetCore(action));
+
+    /// <summary>
+    /// Configure distributed tracing via OpenTelemetry with HttpClient and ASP.NET Core Instrumentation along with (optionally) Zipkin and Wavefront exporting.
+    /// </summary>
+    /// <param name="hostBuilder">Your HostBuilder.</param>
+    /// <param name="action">Customize the <see cref="TracerProviderBuilder" />.</param>
+    public static WebApplicationBuilder AddDistributedTracingAspNetCore(this WebApplicationBuilder hostBuilder,  Action<TracerProviderBuilder> action = null)
```
bart-vmware, line 36: Maybe drop the `Core` in `AspNetCore` from these method names?

bart-vmware (commented):
Done with review pass. Nice work!
A nice side effect is this probably cuts our build and testrun times in half!

src/Management/src/EndpointCore/Extensions/ServiceCollectionExtensions.netcore31.cs
bart-vmware: What do you have in mind that possibly makes removing them a bad idea?

TimHess: I'm just thinking about the case where somebody is updating Steeltoe and their target framework at or around the same time... maybe it's not an issue since 4.0 only targets .net 6 and they wouldn't be able to. Wondering if there's a way we could realistically [Obsolete] the code with a hint where to go next

build/templates/component-build.yaml
```diff
@@ -15,10 +15,6 @@ jobs:
   pool:
     vmImage: ${{parameters.OS}}-latest
   steps:
-  - task: UseDotNet@2
-    displayName: Install .NET Core 3.1
-    inputs:
-      version: 3.1.x
```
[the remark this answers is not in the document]
TimHess: Not intentional, I just missed it! I considered dropping any `Install` tasks, but figured that it doesn't add much time and we'll need it back for .NET 7 soon enough anyway

src/Connectors/test/Connector.EFCore.Test/EntityFrameworkCoreTypeLocatorTest.cs
TimHess: IIRC there are at least 2 packages that can be used and they worked in the same scenarios. We were just using framework to be able to test both of those packages

src/Management/src/TracingCore/TracingHostBuilderExtensions.cs
TimHess: Hm. Maybe, but I think I'd rather tackle it during upcoming package combining or API review efforts

src/Connectors/test/Connector.EFCore.Test/EntityFrameworkCoreTypeLocatorTest.cs
bart-vmware: Makes sense, we should keep it then.

src/Connectors/test/Connector.EFCore.Test/Steeltoe.Connector.EFCore.Test.csproj
```diff
@@ -17,10 +17,11 @@
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="$(EFCoreTestVersion)" />
     <PackageReference Include="Oracle.EntityFrameworkCore" Version="$(EFCoreOracleTestVersion)" />
   </ItemGroup>
-
+  <!-- Come back to this in the future with another strategy
   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
     <PackageReference Include="MySql.EntityFrameworkCore" Version="$(EFCoreTestVersion)" />
   </ItemGroup>
+  -->
```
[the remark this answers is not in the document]
TimHess: I think all of the EFCore related questions/comments are ultimately the same issue: a strategy for testing with more than one compatible NuGet package

src/Management/src/TracingCore/TracingHostBuilderExtensions.cs
bart-vmware: Fine by me. There are probably more occurrences, I haven't checked for that. Just thought I'd point it out.

src/Management/src/EndpointCore/Extensions/ServiceCollectionExtensions.netcore31.cs
bart-vmware: I wouldn't worry about that, this happens all the time. But it would be good to mention in the release notes, though. That's the first place I look after updating and finding that the code no longer compiles. If we list breaking changes in a document, then we don't need to keep compiling/testing/etc the old code.

To be honest, I expected a different concern: a user is updating to .NET 6, but not adopting the new `Program.cs` without `Startup.cs`. So still using the old host builder, which is a scenario still supported by ASP.NET. If we do we want to support that scenario, we need to keep the `IServiceCollection` extension methods, as well as the ones on host builder.

TimHess: I _think_ that's already handled, but I don't know offhand - @hananiel do you have a thought here?

bart-vmware: Well if the code was already in a `#if`, scoping it down to .NET Core, then we haven't supported that scenario until now, therefore I think we don't need to start supporting it now. But maybe this varies on a case-by-case basis. Sounds like something to take into account during public API review.

TimHess: Oh.. re-reading this and looking at the other file, I don't think this comment belonged here at all. This particular example is a total non-issue for any of the concerns we've discussed

src/Connectors/test/Connector.EFCore.Test/Steeltoe.Connector.EFCore.Test.csproj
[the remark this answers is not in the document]
bart-vmware: I don't understand. If our goal is to test on both packages, what's preventing us from doing that? Is it not just a matter of referencing both NuGets from the test project and run against both of them, possibly using a base class?

src/Management/src/EndpointCore/Extensions/ServiceCollectionExtensions.netcore31.cs
TimHess: We should revisit this conversation in https://github.com/SteeltoeOSS/Steeltoe/issues/827

src/Connectors/test/Connector.EFCore.Test/Steeltoe.Connector.EFCore.Test.csproj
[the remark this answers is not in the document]
TimHess: Found an alternative approach that might have also revealed an unexpected signature change in `MySql.EntityFrameworkCore`. We should probably test the existing behavior with the Steeltoe MySql Connector in a .NET 6 app using EF Core 6.0 and `MySql.EntityFrameworkCore` to confirm the current status

src/Connectors/test/Connector.EFCore.Test/MySqlDbContextOptionsExtensionsTest.cs
```diff
@@ -2,22 +2,20 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
-using System;
-using System.Collections.Generic;
 using Microsoft.EntityFrameworkCore;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
+using Steeltoe.Connector.EFCore;
+using MySqlDataConnectionAlias = MySql.Data.MySqlClient.MySqlConnection;
```
bart-vmware, line 9: I find it tricky to track what's what, as the namespaces of these external types aren't very helpful. Suggestion: use two aliases to make the distinction clear at their usages:
```c#
using OfficialMySqlConnection = MySql.Data.MySqlClient.MySqlConnection;
using PomeloMySqlConnection = MySqlConnector.MySqlConnection;
```
This would result in a build error when unqualified `MySqlConnection` is used, because there's no namespace import for it (which is a good thing).

src/Connectors/test/Connector.EFCore.Test/MySqlDbContextOptionsExtensionsTest.cs
```diff
@@ -80,9 +78,13 @@ public void UseMySql_ThrowsIfServiceNameNull()
...
     }
 
-    [Fact]
-    public void AddDbContext_NoVCAPs_AddsDbContext_WithMySqlConnection()
+    [Theory]
+    [InlineData(new[]{"Pomelo.EntityFrameworkCore.MySql"}, typeof(MySqlConnector.MySqlConnection))]
+    [InlineData(new[]{"MySql.EntityFrameworkCore", "MySql.Data.EntityFrameworkCore"}, typeof(MySqlDataConnectionAlias))]
+    public void AddDbContext_NoVCAPs_AddsDbContext_WithMySqlConnection(string[] efAssemblies, Type mySqlConnection)
     {
+        SetTypeLocators(efAssemblies, mySqlConnection);
```
bart-vmware, line 86: When this test fails, it influences the outcome of other tests because it fails to reset. To fix that, at the minimum it should be wrapped in a try-catch, but there's a more elegant solution:
```c#
private sealed class AlternateTypeLocatorScope : IDisposable
{
    private readonly string[] _mySqlEntityAssembliesBackup;
    private readonly string[] _mySqlAssembliesBackup;

    public AlternateTypeLocatorScope(string[] mySqlEntityAssemblies, Type mySqlConnectionType)
    {
        _mySqlEntityAssembliesBackup = EntityFrameworkCoreTypeLocator.MySqlEntityAssemblies;
        EntityFrameworkCoreTypeLocator.MySqlEntityAssemblies = mySqlEntityAssemblies;

        _mySqlAssembliesBackup = MySqlTypeLocator.Assemblies;
        MySqlTypeLocator.Assemblies = new[] { mySqlConnectionType.Assembly.FullName?.Split(',')[0] };
    }

    public void Dispose()
    {
        EntityFrameworkCoreTypeLocator.MySqlEntityAssemblies = _mySqlEntityAssembliesBackup;
        MySqlTypeLocator.Assemblies = _mySqlAssembliesBackup;
    }
}
```
Which enables the tests to use:
```c#
using var scope = new AlternateTypeLocatorScope(efAssemblies, mySqlConnection);
```

src/Connectors/test/Connector.EFCore.Test/MySqlDbContextOptionsExtensionsTest.cs
```diff
@@ -80,9 +78,13 @@ public void UseMySql_ThrowsIfServiceNameNull()
         Assert.Contains(nameof(serviceName), ex4.Message);
     }
 
-    [Fact]
-    public void AddDbContext_NoVCAPs_AddsDbContext_WithMySqlConnection()
+    [Theory]
+    [InlineData(new[]{"Pomelo.EntityFrameworkCore.MySql"}, typeof(MySqlConnector.MySqlConnection))]
+    [InlineData(new[]{"MySql.EntityFrameworkCore", "MySql.Data.EntityFrameworkCore"}, typeof(MySqlDataConnectionAlias))]
+    public void AddDbContext_NoVCAPs_AddsDbContext_WithMySqlConnection(string[] efAssemblies, Type mySqlConnection)
```
bart-vmware, line 84: `Type mySqlConnection` parameter: Its common practice to suffix identifier names that represent a `Type`, `MethodInfo`, `PropertyInfo` etc, to distinguish them from instances. So I suggest to rename this to `Type mySqlConnectionType`, which makes the line `Assert.True(con.GetType() == mySqlConnection);` not look like a bug.

src/Connectors/test/Connector.EFCore.Test/Steeltoe.Connector.EFCore.Test.csproj
```diff
@@ -17,12 +17,8 @@
     <PackageReference Include="Npgsql.EntityFrameworkCore.PostgreSQL" Version="$(EFCoreTestVersion)" />
     <PackageReference Include="Oracle.EntityFrameworkCore" Version="$(EFCoreOracleTestVersion)" />
   </ItemGroup>
-  <!-- Come back to this in the future with another strategy
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup>
```
bart-vmware, line 20: There's no more need to have separate `ItemGroup`s containing `PackageReference`s now, they can be merged into one.

src/Connectors/src/Connector.EFCore/MySqlDbContextOptionsExtensions.cs
```diff
@@ -187,10 +188,20 @@ private static DbContextOptionsBuilder DoUseMySql(DbContextOptionsBuilder builde
...
+            {
+                useMethod = FindUseSqlMethod(extensionType, new[] { typeof(DbContextOptionsBuilder), typeof(string), typeof(Action<DbContextOptionsBuilder>) });
+                parameters = new[] { builder, connection, mySqlOptionsAction };
+            }
+            else
+            {
+                // If the server version wasn't passed in, see if we need to use the EF Core lib to autodetect it (this is the part that creates an extra connection)
+                serverVersion ??= ReflectionHelpers.FindMethod(EntityFrameworkCoreTypeLocator.MySqlVersionType, "AutoDetect", new[] { typeof(string) }).Invoke(null, new object[] { connection });
+                parameters = new[] { builder, connection, serverVersion, mySqlOptionsAction };
+            }
```
bart-vmware, line 204: I'm not confident asserting the correctness of this, because I don't sufficiently know what's going on here. I trust you did the right thing, but if you're in need of an extra pair of eyes, please ask someone on the team that has a better understanding of this.

src/Connectors/src/Connector.EFCore/MySqlDbContextOptionsExtensions.cs
```diff
@@ -216,13 +227,12 @@ private static MethodInfo FindUseSqlMethod(Type type, Type[] parameterTypes)
...
+        foreach (var ci in declaredMethods.Where(method => method.Name.Equals("UseMySQL", StringComparison.InvariantCultureIgnoreCase)))
         {
             var parameters = ci.GetParameters();
-            if ((parameters.Length == 3 || parameters.Length == parameterTypes.Length) &&
-                ci.Name.Equals("UseMySQL", StringComparison.InvariantCultureIgnoreCase) &&
-                parameters[0].ParameterType.Equals(parameterTypes[0]) &&
-                parameters[1].ParameterType.Equals(parameterTypes[1]) &&
+            if (parameters.Length == parameterTypes.Length &&
+                parameters[0].ParameterType == parameterTypes[0] &&
+                parameters[1].ParameterType == parameterTypes[1] &&
```
bart-vmware, line 235: This should also check that the number of parameters is >= 2, to avoid a crash when this method is called with just one parameter.

TimHess: I hear what you're saying, but this should be caught at either dev-time or in the CI build. This method is private, only intended to work with 2 specific scenarios. Ideally we're making absolutely minimal changes in here before we get to the major refactor (which also needs to be part of 4.0)

src/Connectors/src/Connector.EFCore/MySqlDbContextOptionsExtensions.cs
TimHess: I got around to testing this with a sample app.. the code is cleaner now and works with 6.0 versions of both packages we're intending to support

hananiel (approved):
LGTM

src/Connectors/src/Connector.EFCore/MySqlDbContextOptionsExtensions.cs
bart-vmware: I agree with making minimal changes to existing code, because that's not the scope of this PR. However the diff shows you actually _removed_ the existing length check. The other 2 `FindUseSqlMethod` methods in the solution (untouched by this PR) contain that check as well, so for consistency I think we should keep it.

src/Connectors/src/Connector.EFCore/MySqlDbContextOptionsExtensions.cs
bart-vmware: That's great! Closing this one.

src/Connectors/src/Connector.EFCore/MySqlDbContextOptionsExtensions.cs
TimHess: Are you suggesting this:
`(parameters.Length == 3 || parameters.Length == parameterTypes.Length)` 
this:
`(parameters.Length >= 2 && parameters.Length == parameterTypes.Length)`
Or something different?

bart-vmware: Yes

bart-vmware: I realized the old code had `||` instead of `&&`, so it didn't validate properly either. This makes my argumentation invalid, so I'm okay with keeping it as-is.

bart-vmware (approved):

Issue #973, opened 2022-08-02
Consolidate Steeltoe Packages Names


TimHess:
Over the last several releases of .NET (Core), Microsoft has moved towards ASP.NET Core and several key .NET feature areas (like Configuration, Dependency Injection and Logging) being included in the runtime, removing the NuGet references that we were trying to avoid pulling in unless necessary. As such, there's no longer a solid reason for the Base/Core suffix naming convention that Steeltoe has used since version 2.0. 

Steeltoe 4.0 will look to consolidate packages in order to streamline package naming and dependency trees.

PR #974, opened 2022-08-03
Add global CLSCompliant/ComVisible attributes and unify .props

### LINKED ISSUE #943: Address S3990/S3992: Assemblies should be marked as CLSCompliant/ComVisible
CLSCompliantAttribute and ComVisibleAttribute are [no longer relevant](https://github.com/dotnet/extensions/issues/3147), so we should add:
```c#
[assembly: CLSCompliant(false)]
[assembly: ComVisible(false)]
```
in a centralized location and remove them from the individual `AssemblyInfo.cs` files, and keep these rules turned off.

bart-vmware:
## Description

Project files used to always import `versions.props`, along with either `sharedproject.props` or `sharedtest.props`, depending on the project type. This PR changes project files to always only import `shared.props`. This new `shared.props` imports the other files, depending on conventions. This new file also contains what was common between `sharedproject.props` and `sharedtest.props`. Additional commits contain some more cleanups.

After these changes, it was easy to add assembly-level attributes for CLSCompliant/ComVisible.

Fixes #943.

## Quality checklist


shared.props
```diff
@@ -0,0 +1,37 @@
...
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\Steeltoe.Debug.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+
+  <PropertyGroup Condition="'$(Configuration)' == 'Release'">
+    <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <CodeAnalysisRuleSet>$(MSBuildThisFileDirectory)\Steeltoe.Release.ruleset</CodeAnalysisRuleSet>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <AdditionalFiles Include="$(MSBuildThisFileDirectory)\stylecop.json" Visible="false" />
```
TimHess, line 17: I think this file previously also had `<CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>` attached to it, because the default value prevented incremental builds from working (for example, if you made a change in Management.EndpointCore, all projects it depends on would rebuild first even if nothing changed) ~ unless something has changed in the build tooling, the loss of this value could result in wasted compute time

shared.props
```diff
@@ -0,0 +1,37 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <AssemblyAttribute Include="System.CLSCompliantAttribute">
```
TimHess, line 21: How many issues does it create if this is set to `true`? Not saying we need to change it, just curious

src/Configuration/test/ConfigServer.Test/Info.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
-namespace Steeltoe.Extensions.Configuration.ConfigServer.ITest;
+namespace Steeltoe.Extensions.Configuration.ConfigServer.Test;
```
TimHess, line 5: Changing the name of this project seems likely to cause unresolvable conflict with #973. I'd rather spell out `IntegrationTest` than drop the `I` completely

TimHess (changes requested):
ConfigServer.ITest should not be renamed to ConfigServer.Test

src/Configuration/test/ConfigServer.Test/Info.cs
bart-vmware: I assumed the `I` was a typo, didn't realize it was referring to integration tests. Renaming to *.IntegrationTest violates the documented naming convention:

> The unit tests for a package should be put in a project with the same name and append .Test to the name.

Alternatively we could rename to *.Integration.Test. Which do you prefer?

shared.props
bart-vmware: When turned on, it results in 1109 warnings.

shared.props
bart-vmware: As documented [here](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/EnableConfiguration.md), there's no need to copy `stylecop.json` to the output directory. This makes sense to me, because project items of type `AdditionalFile` were specifically designed to be consumable by roslyn analyzers. Roslyn analyzers only execute at compile time, so there's no point in copying them to the output directory to be consumed at runtime.

Looking at history, it used to be `<CopyToOutputDirectory>Always</CopyToOutputDirectory>`, which basically means the parent project changes always, resulting in an (unneeded) rebuild.

shared.props
TimHess: OK, that seems be enough to not entertain the notion of turning on right now

src/Configuration/test/ConfigServer.Test/Info.cs
TimHess: Let's go with .Integration.Test

TimHess (approved):
LGTM

PR #978, opened 2022-08-05
Fix type and member naming

### LINKED ISSUE #909: Address S100: Methods and properties should be named in PascalCase
Address existing violations of [S100: Methods and properties should be named in PascalCase](https://rules.sonarsource.com/csharp/RSPEC-100) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

- Fix the violation by renaming the affected member
- Suppress the violation in code using `#pragma warning disable/restore`, in case it represents a well-known exception (for example: OSX)

bart-vmware:
## Description

Fixes #909.

## Quality checklist

### DIFF
--- a/src/Common/src/Common.Expression/Spring/Standard/Tokenizer.cs
+++ b/src/Common/src/Common.Expression/Spring/Standard/Tokenizer.cs
@@ -688,4 +688,3 @@ internal sealed class Tokenizer
         throw new InternalParseException(new SpelParseException(_expressionString, start, msg, inserts));
     }
 }
-#pragma warning restore S125 // Sections of code should not be commented out
--- a/src/Common/test/Common.Expression.Test/Spring/PerformanceTests.cs
+++ b/src/Common/test/Common.Expression.Test/Spring/PerformanceTests.cs
@@ -135,4 +135,3 @@ public class PerformanceTests
         }
     }
 }
-#pragma warning restore xUnit1004 // Test methods should not be skipped
--- a/src/Common/test/Common.Expression.Test/Spring/SpelReproTests.cs
+++ b/src/Common/test/Common.Expression.Test/Spring/SpelReproTests.cs
@@ -11,6 +11,9 @@ using Steeltoe.Common.Expression.Internal.Spring.Support;
 using Steeltoe.Common.Util;
 using Xunit;
 
+// ReSharper disable InconsistentNaming
+#pragma warning disable S100 // Methods and properties should be named in PascalCase
+
 namespace Steeltoe.Common.Expression.Internal.Spring;
 
 public class SpelReproTests : AbstractExpressionTests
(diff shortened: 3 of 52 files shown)


src/Common/src/Common.Expression/Spring/Standard/Tokenizer.cs
```diff
@@ -688,4 +688,3 @@ private void RaiseParseException(int start, SpelMessage msg, params object[] ins
         throw new InternalParseException(new SpelParseException(_expressionString, start, msg, inserts));
     }
 }
-#pragma warning restore S125 // Sections of code should not be commented out
```
TimHess, line 691: What about the line (6) that starts the suppression?

src/Common/test/Common.Expression.Test/Spring/SpelReproTests.cs
```diff
@@ -11,6 +11,11 @@
 using Steeltoe.Common.Util;
 using Xunit;
 
+// ReSharper disable InconsistentNaming
+#pragma warning disable S100 // Methods and properties should be named in PascalCase
+#pragma warning disable SA1307 // Accessible fields should begin with upper-case letter
+#pragma warning disable IDE1006 // Naming Styles
```
TimHess, line 17: Should these be turned back on later in the file?

src/Common/test/Common.Expression.Test/Spring/PerformanceTests.cs
```diff
@@ -135,4 +135,3 @@ public void TestPerformanceOfMethodAccess()
         }
     }
 }
-#pragma warning restore xUnit1004 // Test methods should not be skipped
```
TimHess, line 138: What about the line that starts the suppression?

src/Common/test/Common.Expression.Test/Spring/SpelReproTests.cs
bart-vmware: No, all preprocessor defines automatically go out of scope at the end of the file they were declared in.

src/Common/test/Common.Expression.Test/Spring/PerformanceTests.cs
bart-vmware: The suppression is still needed, but there's just no point in restoring at the end of the file.

TimHess (approved):

src/Common/src/Common.Expression/Spring/Standard/Tokenizer.cs
bart-vmware: I suspect I've answered this in the other responses. Did you mean something else?

TimHess: Just another instance of the same question. Feels like having something open and not close, but if there's no side affect then I'm not worried about it

bart-vmware: I get your point. With preprocessor directives its quite common to not 'close' them, in contrast to most other things. For example, consider `#define USE_TRACING` at the top of a file, which makes it easy to turn on/off during debug. If we were to 'close' that one, we'd have to undefine USE_TRACING at the end of the file, which nobody ever does.

PR #982, opened 2022-08-05
Add/remove Async suffix from method names (fixes S4261)

### LINKED ISSUE #952: Address S4261: Methods should be named according to their synchronicities
Address existing violations of [S4261: Methods should be named according to their synchronicities](https://rules.sonarsource.com/csharp/RSPEC-4261) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

- Add or remove the "Async" suffix

bart-vmware:
Fixes #952.

## Quality checklist

### DIFF
--- a/src/CircuitBreaker/test/HystrixBase.Test/HystrixCommandTest.cs
+++ b/src/CircuitBreaker/test/HystrixBase.Test/HystrixCommandTest.cs
@@ -849,7 +849,7 @@ public class HystrixCommandTest : CommonHystrixCommandTests<TestHystrixCommand<i
             f = command1.ExecuteAsync(); // Running
             Time.Wait(50); // Let first start
             f2 = command2.ExecuteAsync(); // In Queue
-            await command3.ExecuteAsync(); // Start, queue rejected
+            await command3.ExecuteAsync(); // StartAsync, queue rejected
             Assert.True(false, "we shouldn't get here");
         }
         catch (Exception e)
--- a/src/CircuitBreaker/test/HystrixBase.Test/HystrixCommandTest.cs
+++ b/src/CircuitBreaker/test/HystrixBase.Test/HystrixCommandTest.cs
@@ -849,7 +849,7 @@ public class HystrixCommandTest : CommonHystrixCommandTests<TestHystrixCommand<i
             f = command1.ExecuteAsync(); // Running
             Time.Wait(50); // Let first start
             f2 = command2.ExecuteAsync(); // In Queue
-            await command3.ExecuteAsync(); // StartAsync, queue rejected
+            await command3.ExecuteAsync(); // Start, queue rejected
             Assert.True(false, "we shouldn't get here");
         }
         catch (Exception e)
--- a/src/Common/test/Common.Http.Test/HttpClientHelperTest.cs
+++ b/src/Common/test/Common.Http.Test/HttpClientHelperTest.cs
@@ -162,9 +162,9 @@ public class HttpClientHelperTest
     [Fact]
     public void GetAccessToken_ThrowsNulls()
     {
-        Assert.ThrowsAsync<ArgumentException>(() => HttpClientHelper.GetAccessToken(string.Empty, null, null));
-        Assert.ThrowsAsync<ArgumentException>(() => HttpClientHelper.GetAccessToken("https://foo/bar", null, null));
-        Assert.ThrowsAsync<ArgumentException>(() => HttpClientHelper.GetAccessToken("https://foo/bar", "clientid", null));
+        Assert.ThrowsAsync<ArgumentException>(() => HttpClientHelper.GetAccessTokenAsync(string.Empty, null, null));
+        Assert.ThrowsAsync<ArgumentException>(() => HttpClientHelper.GetAccessTokenAsync("https://foo/bar", null, null));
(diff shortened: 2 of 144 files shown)


src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
```diff
@@ -925,13 +925,13 @@ protected internal string FetchAccessToken()
...
 
-        return HttpClientHelper.GetAccessToken(settings.AccessTokenUri, settings.ClientId, settings.ClientSecret, settings.Timeout,
+        return HttpClientHelper.GetAccessTokenAsync(settings.AccessTokenUri, settings.ClientId, settings.ClientSecret, settings.Timeout,
             settings.ValidateCertificates, httpClient, logger).GetAwaiter().GetResult();
     }
 
     // fire and forget
 #pragma warning disable S3168 // "async" methods should not return "void"
-    protected internal async void RefreshVaultTokenAsync(object state)
+    protected internal async void RefreshVaultToken(object state)
```
TimHess, line 934: While it doesn't return `Task`, shouldn't it retain `async` anyway given that it is async?

TimHess: * This is not the only instance, but I'll refrain from commenting on any others until we resolve here

src/CircuitBreaker/test/HystrixBase.Test/HystrixCommandTest.cs
```diff
@@ -849,7 +849,7 @@ public async Task TestRejectedThreadWithNoFallback()
             f = command1.ExecuteAsync(); // Running
             Time.Wait(50); // Let first start
             f2 = command2.ExecuteAsync(); // In Queue
-            await command3.ExecuteAsync(); // Start, queue rejected
+            await command3.ExecuteAsync(); // StartAsync, queue rejected
```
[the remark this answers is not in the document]
bart-vmware: Nice catch! I must have missed that one while reviewing proposed changes in comments as part of rename. I'll revert this one.

src/Messaging/src/RabbitMQ/Core/BatchingRabbitTemplate.cs
```diff
@@ -93,7 +93,7 @@ public override void Send(string exchange, string routingKey, IMessage message,
                 {
                     _cancellationTokenSource = new CancellationTokenSource();
                     TimeSpan delay = next.Value - DateTime.Now;
-                    _scheduledTask = Task.Run(() => ReleaseBatches(delay, _cancellationTokenSource.Token), _cancellationTokenSource.Token);
+                    _scheduledTask = Task.Run(() => ReleaseBatchesAsync(delay, _cancellationTokenSource.Token), _cancellationTokenSource.Token);
```
TimHess, line 96: Should there also be an `await` on this line? cc @dtillman @hananiel

src/Common/test/Common.Http.Test/HttpClientHelperTest.cs
```diff
@@ -162,9 +162,9 @@ public void GetRequestMessage_CreatesCorrectMessage_WithBasicAuth()
     [Fact]
     public void GetAccessToken_ThrowsNulls()
```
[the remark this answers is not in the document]
bart-vmware: That's hard to say for sure. Assuming the exception is thrown before the first await point, there's no need to actually await the call. The same concept applies to `yield returns`. So if the intent of this test is to throw early, then it is correct. But the test is named quite poorly, so we can't know.

```c#
var task = HttpClientHelper.GetAccessToken(...);  // throw here? (seems most correct to me)
await task; // or here?
```

src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
bart-vmware: I've removed the Async suffix in all three suggested cases, for the next reason. The purpose of the `async` keyword in the method declaration is only for the method body itself. It is unobservable by callers, they see just `void` or `Task`. So from a callers perspective, this method has nothing to do with asynchronous execution, which is why adding the prefix would be wrong.

src/Messaging/src/RabbitMQ/Core/BatchingRabbitTemplate.cs
bart-vmware: Yes, I suspect it should. The `ReleaseBatchesAsync` method contains await points. So any exception thrown after the first await point goes unnoticed. Maybe even crash the process, although I'm not entirely sure about that as it depends on the execution context.

TimHess: This concern will be revisited in another issue that I'll create momentarily

TimHess (approved):
These changes are an improvement that have highlighted some problems that will be fixed in another PR

src/Messaging/src/RabbitMQ/Core/BatchingRabbitTemplate.cs
bart-vmware: Resolving, this is trakced at #983.

PR #987, opened 2022-08-12
Correct release ruleset metadata

ccheetham:
Updated ruleset metadara to match the semantics suggested by its filename


bart-vmware (approved):
Thanks!

Issue #829, opened 2022-03-03
InvalidOperationException for a configuration mistake is it too much?


sintetico82:
https://github.com/SteeltoeOSS/Steeltoe/blob/8e9686587844f86b40135cbe73d6d193f0c4bcb5/src/Discovery/src/Consul/ConsulPostConfigurer.cs#L24

In my opinion, avoid to start an application in a container only for a miss configuration for service discovery is too much. Why don't just write a critical error log?

TimHess:
Sorry for the super delayed response, I thought I replied to this already. The intent here was a fast failure on what was seen as something that could only be a misconfiguration that is likely critical to the success of the application.

You're not wrong, it is a bit extreme. Do you have a scenario where the app is likely to run successfully with this configuration?

sintetico82:
Sorry for the delayed answer too.

I catch this in error in the development environment. I think in some multiple purpose scenarios, where some services can be also optional (for developing reasons, for the available chain, or other factors).

In my opinion, let the use of framework more flexible could be a better option, but of course, it's a questionable opinion.

thompson-tomo:
As a thought, could we extend the configure options with an enabled flag which if false doesn't result in the exception. I note the logic is in static class & the Eureka implementation is very similar except it has the enabled flag.

TimHess:
Thank you for that thought - I should note that @bart-vmware is in the middle of a comprehensive API review and some potentially significant re-work of the entire service discovery component (ahead of working with [Aspire](#1254)), so he would be in a better position to comment here, and it would be prudent to check with him before attempting any changes to discovery in `main`.

bart-vmware:
We've had some internal discussion and believe it's best to follow ASP.NET conventions here, which is implementing [IValidateOptions](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/options?view=aspnetcore-8.0#ivalidateoptionstoptions-and-ivalidatableobject), which throws `OptionsValidationException` when the configuration is invalid. This helps to catch mistakes early, instead of silently ignoring what's configured (assuming not everyone is watching the logs all the time).

PR #989, opened 2022-08-16
Sonar various

### LINKED ISSUE #912: Address S1067: Expressions should not be too complex
Address existing violations of [S1067: Expressions should not be too complex](https://rules.sonarsource.com/csharp/RSPEC-1067) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

  - Most are easy to fix by comparing against a frozen dictionary or extract method

### LINKED ISSUE #926: Address S2292: Trivial properties should be auto-implemented
Address existing violations of [S2292: Trivial properties should be auto-implemented](https://rules.sonarsource.com/csharp/RSPEC-2292) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

### LINKED ISSUE #927: Address S2302: "nameof" should be used
Address existing violations of [S2302: "nameof" should be used](https://rules.sonarsource.com/csharp/RSPEC-2302) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

Note: This rule results in many false positives, so consider to use Rider/Resharper analysis instead to fix existing problems and leave this rule disabled.

Related to #940.

### LINKED ISSUE #940: Address S3928: Parameter names used into ArgumentException constructors should match an existing one
Address existing violations of [S3928: Parameter names used into ArgumentException constructors should match an existing one](https://rules.sonarsource.com/csharp/RSPEC-3928) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

Related to #927.

### LINKED ISSUE #948: Address S4039: Interface methods should be callable by derived types
Address existing violations of [S4039: Interface methods should be callable by derived types](https://rules.sonarsource.com/csharp/RSPEC-4039) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

To address the violations (currently one), choose from the following on a case-by-case basis:

bart-vmware:
## Description

Addressed Sonar rules:
- Fixed S1067: Expressions should not be too complex
- Fixed S2292: Trivial properties should be auto-implemented
- Reviewed S2302: "nameof" should be used
  - Fixed a small set of violations, but many false positives remain, so keeping this rule turned off
- Fixed S3928: Parameter names used into `ArgumentException` constructors should match an existing one
- Fixed S4039: Interface methods should be callable by derived types

Other:
- Introduced argument guards and converted existing usages of Argument*Exception
- Renamed parameters of type `System.Type` to contain 'Type' suffix
- Use structured log messages, remove unneeded `ToString()` calls, ensure exceptions are propagated and fix parameter order
- Remove line break between property get/set and event add/remove
  - StyleCop used to require this, while the Resharper formatter has a bug that fails to remove them
- Rename files to match type name, remove unneeded Resharper exclusion, cleanup `AssemblyInfo.cs` files

Fixes #912
Fixes #926
Fixes #927
Fixes #940
Fixes #948

## Quality checklist


src/Discovery/src/Eureka/EurekaPostConfigurer.cs
```diff
@@ -191,6 +182,37 @@ public static void UpdateConfiguration(IConfiguration config, EurekaServiceInfo
...
+        {
+            return;
+        }
+
+        if (!clientOptions.ShouldRegisterWithEureka && !clientOptions.ShouldFetchRegistry)
+        {
+            return;
+        }
+
+        throw new InvalidOperationException(
```
TimHess, line 212: Possibly outside the scope of this PR, but if we can log instead of throwing it _might_ be preferable https://github.com/SteeltoeOSS/Steeltoe/issues/829

src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
```diff
@@ -46,10 +46,22 @@ static bool Query(Type t)
...
-                    t.Name.StartsWith("HeapDump");
+                bool isThreadDump = type.Name.StartsWith("ThreadDump", StringComparison.Ordinal);
+                bool isHeapDump = type.Name.StartsWith("HeapDump", StringComparison.Ordinal);
+
+                if (!(isThreadDump || isHeapDump))
+                {
+                    return true;
+                }
+
+                if (isThreadDump && Platform.IsWindows)
```
TimHess, line 64: I think this is incorrect... ThreadDump should work on any platform since #822 (and some cleanup in #848). Also looks like the docs never got updated to indicate this changed

src/CircuitBreaker/src/Abstractions/Properties/AssemblyInfo.cs
```diff
@@ -4,28 +4,4 @@
...
-// Version information for an assembly consists of the following four values:
-//
-//      Major Version
-//      Minor Version
-//      Build Number
-//      Revision
-//
-// [assembly: AssemblyVersion("1.0.0.0")]
-// [assembly: AssemblyFileVersion("1.0.0.0")]
 [assembly: InternalsVisibleTo("Steeltoe.CircuitBreaker.HystrixBase.Test")]
```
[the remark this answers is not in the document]
bart-vmware: Thanks. I took the suggestion, then realized the entire file should be deleted instead, so I force-pushed that.

src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
bart-vmware: Glad you're bringing this up. The result looked odd to me, so I double-checked to ensure the refactoring didn't change the outcome. What should it be instead? And which documentation are you referring to?

src/Discovery/src/Eureka/EurekaPostConfigurer.cs
bart-vmware: The linked issue is about `ConsulPostConfigurer` (called from `ValidateConsulOptions()`), while this comment is on `EurekaPostConfigurer` (called from `UpdateConfiguration()`). Does the same concern apply here?

For both, we could add `bool AreOptionsValid()`, which enables callers to check upfront. Then the callers can be expanded with a `bool throwOnError` parameter. These are just suggestions without me fully understanding the context, so my thinking is this is probably better first discussed in the linked issue, followed by a separate PR (and possibly backported).

src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
TimHess: I guess this line would just be `if (isThreadDump)`, but that makes the whole method look a bit weird since SupportedOnPlatform is basically always true now

src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
```diff
@@ -46,10 +46,22 @@ static bool Query(Type t)
...
+                {
+                    return true;
+                }
```
bart-vmware, line 67:
```suggestion
                return isThreadDump || isHeapDump;
```

bart-vmware: Is that what it should be?

bart-vmware: Or remove `SupportedOnPlatform` entirely?

src/Discovery/src/Eureka/EurekaPostConfigurer.cs
TimHess: Yes the same concern applies. yeah, let's postpone changes for now

src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
TimHess: I think we can delete this whole thing, the result should always be `true`

src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
```diff
@@ -46,10 +46,22 @@ static bool Query(Type t)
...
+
+                return isHeapDump;
             }
 
             List<Type> types = Assembly.Load("Steeltoe.Management.Endpoint").GetTypes().Where(Query).Where(SupportedOnPlatform).ToList();
```
TimHess, line 72:
```suggestion
            List<Type> types = Assembly.Load("Steeltoe.Management.Endpoint").GetTypes().Where(Query).ToList();
```

bart-vmware: Thanks, I've already pushed a commit to address this (and keep the formatter happy).

TimHess (approved):
LGTM

Issue #897, opened 2022-07-27
Public API Surface Area Review


TimHess:
Steeltoe has a broad surface area of public APIs, probably broader than necessary which can hamper project progress. This epic will track efforts to reduce what is (unnecessarily) public for Steeltoe 4.0.

bart-vmware:
Considerations
- Review namespaces, consider [pubternal](https://www.reddit.com/r/dotnet/comments/jwzgl9/word_of_the_day_pubternal/) use
  - Don't put extensions methods in System namespaces. While this makes them easy to discover, sooner or later they are going to clash with other libraries, requiring users to write non-intuitive code to workaround the conflict
- Remove obsolete types and members
- Consider making public types internal, use `InternalsVisibleTo` to access them from tests
- Seal public types, unless they are designed for inheritance
- Avoid static classes, they are hard to test. Exceptions: extension method containers or classes to wrap a set of immutable constants
- A class or interface should have a single purpose
  - A class or interface should have a single purpose within the system it functions in. In general, a class either represents a primitive type like an email or ISBN number, an abstraction of some business concept, a plain data structure, or is responsible for orchestrating the interaction between other classes. It is never a combination of those. This rule is widely known as the [Single Responsibility Principle](https://8thlight.com/blog/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html), one of the S.O.L.I.D. principles
  - Similarly, a property, method or local function should do only one thing
- Consider adding doc-comments to public types and members, but avoid [undocumentation](https://blog.codinghorror.com/avoiding-undocumentation/)
- Public types should protect the consistency of their internal state: validate inputs and use guards like `AssertNotDisposed();`
- Use a method instead of a property when:
  - The work is more expensive than setting a field value
  - It represents a conversion such as the `Object.ToString` method
  - It returns a different result each time it is called, even if the arguments didn’t change. For example, the `NewGuid` method returns a different value each time it is called
  - The operation causes a side effect such as changing some internal state not directly related to the property (which violates the [Command Query Separation](http://martinfowler.com/bliki/CommandQuerySeparation.html) principle). Populating an internal cache or implementing [lazy-loading](http://www.martinfowler.com/eaaCatalog/lazyLoad.html) is a good exception
- Suffix identifier names that represent a `Type`, `MethodInfo`, `PropertyInfo` etc, to distinguish them from instances. For example: rename `MethodInfo connect` to `MethodInfo connectMethod`
[...]

PR #990, opened 2022-08-16
Enable S3956, address existing violations with changes and suppressions

### LINKED ISSUE #941: Address S3956: "Generic.List" instances should not be part of public APIs
Address existing violations of [S3956: "Generic.List" instances should not be part of public APIs](https://rules.sonarsource.com/csharp/RSPEC-3956) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

  - Prefer collection interfaces such as `ICollection<>` and `IDictionary<,>` over arrays, list and dictionaries for inputs and outputs
    - Be hesitant with incoming `IEnumerable<>` parameters: the implementation may change over time, causing the need for multiple enumerations, which may result in undesired side effects
    - Consider to return unchangeable collections, such as `IReadOnlyCollection<>`, `IReadOnlyDictionary<,>` and `IImmutableList<T>`

Related to #916 and #946.

TimHess:
## Description

Enable S3956, address existing violations with changes and suppressions.

Fixes #941

**PLEASE NOTE** Changes in this PR may be better handled in the context of #897. As a result of this line of thought, this PR may sit for a while, be closed or split into separate pieces and that's OK

## Quality checklist

### DIFF
--- a/src/CircuitBreaker/src/Hystrix.MetricsStream/RabbitMetricsStreamPublisher.cs
+++ b/src/CircuitBreaker/src/Hystrix.MetricsStream/RabbitMetricsStreamPublisher.cs
@@ -67,7 +67,7 @@ public class RabbitMetricsStreamPublisher : HystrixMetricsStreamPublisher
         }
     }
 
-    protected override void OnNext(List<string> jsonList)
+    protected override void OnNext(IReadOnlyCollection<string> jsonList)
     {
         if (Channel != null)
         {
--- a/src/CircuitBreaker/src/Hystrix/AbstractCommand.cs
+++ b/src/CircuitBreaker/src/Hystrix/AbstractCommand.cs
@@ -96,7 +96,7 @@ public abstract class AbstractCommand<TResult> : AbstractCommandBase, IHystrixIn
 
     public ExecutionResult.EventCounts EventCounts => CommandResult.Eventcounts;
 
-    public List<HystrixEventType> ExecutionEvents => CommandResult.OrderedList;
+    public IReadOnlyList<HystrixEventType> ExecutionEvents => CommandResult.OrderedList;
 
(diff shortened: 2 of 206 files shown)


src/CircuitBreaker/src/Hystrix/AbstractCommand.cs
```diff
@@ -97,7 +97,7 @@ protected virtual ExecutionResult CommandResult
 
     public ExecutionResult.EventCounts EventCounts => CommandResult.Eventcounts;
 
-    public List<HystrixEventType> ExecutionEvents => CommandResult.OrderedList;
+    public IEnumerable<HystrixEventType> ExecutionEvents => CommandResult.OrderedList;
```
bart-vmware, line 100: This looks quite odd, given the property name indicates an ordered list instead of a (potentially lazily executed) sequence. `IReadOnlyList<>` seems more appropriate. The advantage it that it allows callers to index into the collection, obtain its count etc, while still protecting against mutation. If mutation is something that should be possible, `IList<>` would be better.

TimHess: I think you're right, `IEnumerable` is likely wrong and is definitely confusing. `IReadOnlyList` should work and makes more sense

src/CircuitBreaker/src/Hystrix.MetricsStream/RabbitMetricsStreamPublisher.cs
```diff
@@ -67,7 +67,7 @@ protected override bool EnsureConnection()
         }
     }
 
-    protected override void OnNext(List<string> jsonList)
+    protected override void OnNext(IList<string> jsonList)
```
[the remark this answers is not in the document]
TimHess: It looks like `IReadOnlyCollection` works here and in related places, I'll push the change

PR #996, opened 2022-08-18
Use AttributeUsageAttribute (S3993) and seal attributes (S4060)

### LINKED ISSUE #944: Address S3993: Custom attributes should be marked with "System.AttributeUsageAttribute"
Address existing violations of [S3993: Custom attributes should be marked with "System.AttributeUsageAttribute"](https://rules.sonarsource.com/csharp/RSPEC-3993) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

### LINKED ISSUE #951: Address S4060: Non-abstract attributes should be sealed
Address existing violations of [S4060: Non-abstract attributes should be sealed](https://rules.sonarsource.com/csharp/RSPEC-4060) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

TimHess:
## Description

Use AttributeUsageAttribute on custom attributes so appropriate usage of the attributes is defined. Also mark non-abstract attributes as sealed

Fixes #944 
Fixes #951

## Quality checklist

### DIFF
--- a/src/Messaging/src/RabbitMQ/Attributes/DeclareQueueBaseAttribute.cs
+++ b/src/Messaging/src/RabbitMQ/Attributes/DeclareQueueBaseAttribute.cs
@@ -4,6 +4,7 @@
 
 namespace Steeltoe.Messaging.RabbitMQ.Attributes;
 
+[AttributeUsage(AttributeTargets.Class)]
 public abstract class DeclareQueueBaseAttribute : Attribute
 {
     public virtual string Durable { get; set; } = string.Empty;
--- a/src/Messaging/src/RabbitMQ/Attributes/DeclareQueueBaseAttribute.cs
+++ b/src/Messaging/src/RabbitMQ/Attributes/DeclareQueueBaseAttribute.cs
@@ -4,7 +4,7 @@
 
 namespace Steeltoe.Messaging.RabbitMQ.Attributes;
 
-[AttributeUsage(AttributeTargets.Class)]
+[AttributeUsage(AttributeTargets.Class, Inherited = false)]
 public abstract class DeclareQueueBaseAttribute : Attribute
 {
     public virtual string Durable { get; set; } = string.Empty;
--- a/Steeltoe.Debug.ruleset
+++ b/Steeltoe.Debug.ruleset
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <RuleSet Name="Steeltoe Debug" Description="Code quality rules for Steeltoe debug builds" ToolsVersion="17.0">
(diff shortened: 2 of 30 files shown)


bart-vmware:
Can you take #951 in this PR too? It seems closely related.

src/Messaging/src/RabbitMQ/Attributes/DeclareQueueBaseAttribute.cs
```diff
@@ -4,6 +4,7 @@
 
 namespace Steeltoe.Messaging.RabbitMQ.Attributes;
 
+[AttributeUsage(AttributeTargets.Class)]
 public abstract class DeclareQueueBaseAttribute : Attribute
```
bart-vmware, line 8: All derived types are annotated with:
```c#
[AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface | AttributeTargets.Method, AllowMultiple = true)]
```
Is this difference intentional?

TimHess: no, I missed that. I'll add those too

TimHess: Actually looking at this again... Given that it is a BaseAttribute I would think that it should only be used on classes

bart-vmware: why?

TimHess: I'll start by saying I'm not an expert on how this is supposed to be used, but looking at how we're currently using it, it's only in classes. Given that it's `abstract` it can't be used directly, so how or why would you use it any other way?

bart-vmware: This is explained at https://docs.microsoft.com/en-us/dotnet/api/system.attributeusageattribute.inherited?view=net-6.0.
Because `ValidOn` defaults to `All` (which is always wrong in our case), the best we can do is set it to `Class` because there is no `None`. But more importantly, we should set `Inherited` to `false`, which basically ensures that the `AttributeUsage` here is ignored by derived classes.

src/Messaging/src/RabbitMQ/Attributes/DeclareQueueBaseAttribute.cs
```diff
@@ -4,6 +4,7 @@
 
 namespace Steeltoe.Messaging.RabbitMQ.Attributes;
 
+[AttributeUsage(AttributeTargets.Class)]
```
TimHess, line 7:
```suggestion
[AttributeUsage(AttributeTargets.Class, Inherited = false)]
```

bart-vmware (approved):

PR #999, opened 2022-08-19
Reduce binary serialization

bart-vmware:
## Description

This PR removes `[Serializable]` from exceptions (most exceptions didn't have it), followed by more cleanup of exception classes.

~Also removes utf-7 encoding, which is obsolete because it is insecure.~ Moved this change to #1004.

## Quality checklist

### DIFF
--- a/src/Common/src/Common.Expression/ParseException.cs
+++ b/src/Common/src/Common.Expression/ParseException.cs
@@ -6,18 +6,18 @@ namespace Steeltoe.Common.Expression.Internal;
 
 public class ParseException : ExpressionException
 {
+    public ParseException(int position, string message)
+        : base(position, message)
+    {
+    }
+
     public ParseException(string expressionString, int position, string message)
         : base(expressionString, position, message)
     {
     }
 
-    public ParseException(int position, string message, Exception cause)
-        : base(position, message, cause)
-    {
-    }
-
-    public ParseException(int position, string message)
-        : base(position, message)
+    public ParseException(int position, string message, Exception innerException)
+        : base(position, message, innerException)
     {
     }
 }
--- a/src/Common/src/Common.Expression/ParseException.cs
+++ b/src/Common/src/Common.Expression/ParseException.cs
@@ -11,13 +11,13 @@ public class ParseException : ExpressionException
     {
     }
 
-    public ParseException(string expressionString, int position, string message)
-        : base(expressionString, position, message)
+    public ParseException(int position, string message, Exception innerException)
+        : base(position, message, innerException)
     {
     }
 
-    public ParseException(int position, string message, Exception innerException)
-        : base(position, message, innerException)
+    public ParseException(string expressionString, int position, string message)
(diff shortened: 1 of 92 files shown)


src/Common/src/Common.Expression/ParseException.cs
```diff
@@ -6,18 +6,18 @@ namespace Steeltoe.Common.Expression.Internal;
...
-    public ParseException(string expressionString, int position, string message)
-        : base(expressionString, position, message)
+    public ParseException(int position, string message)
+        : base(position, message)
     {
     }
 
-    public ParseException(int position, string message, Exception cause)
-        : base(position, message, cause)
+    public ParseException(string expressionString, int position, string message)
```
TimHess, line 14: I see the ordering changes in these classes, most of them before this one have seemed like clear readability improvements, but this one seems less orderly to me... How is the ordering being determined?

src/Messaging/src/RabbitMQ/Support/ShutdownSignalException.cs
```diff
@@ -8,18 +8,6 @@ namespace Steeltoe.Messaging.RabbitMQ.Support;
 
 public class ShutdownSignalException : Exception
 {
-    public ushort ClassId => Args.ClassId;
```
TimHess, line 11: Were these not in use by anything?

TimHess (commented):
This generally looks fine to me, but I don't have context on why Utf7 support was in here in the first place, so I'd like to confirm with @dtillman before removing it

src/Messaging/src/RabbitMQ/Support/ShutdownSignalException.cs
bart-vmware: There weren't used. Their values were already accessible by the `Args` argument directly (which isn't being checked for `null`, so the properties may throw), so I thought I'd remove these. This releaves us from having to remember to update these properties in case the structure of `RC.ShutdownEventArgs` changes, which is in an external library.

src/Common/src/Common.Expression/ParseException.cs
bart-vmware: It was a manual effort. `((int position, string message))` is the shortest, so should come first. But for the remaining two, I don't think one order is better over the other. This was a judgement call, but I'd be happy to swap them if you prefer.

bart-vmware:
> This generally looks fine to me, but I don't have context on why Utf7 support was in here in the first place, so I'd like to confirm with @dtillman before removing it

I've put back UTF-7 encoding, to unblock merging this PR. Removing UTF-7 is now in #1004.

src/Common/src/Common.Expression/ParseException.cs
TimHess: I don't think I have a strong preference, but I'd be inclined to sort by the first parameter. up to you

bart-vmware: I've swapped them.

PR #1000, opened 2022-08-19
Fix conditional cibuild: always build on global config change

bart-vmware:
## Description

It turns out that a PR containing a change to a global file such as `/Steeltoe.Debug.ruleset` does not trigger builds for all the components. This PR fixes that.

From https://docs.microsoft.com/en-us/azure/devops/pipelines/repos/azure-repos-git?view=azure-devops&tabs=yaml&WT.mc_id=AZ-MVP-5003781#paths:

> When you specify paths, you must explicitly specify branches to trigger on. You can't trigger a pipeline with only a path filter; you must also have a branch filter, and the changed files that match the path filter must be from a branch that matches the branch filter.
> 
> Wilds cards are supported for path filters. For instance, you can include all paths that match src/app/**/myapp*. You can use wild card characters (**, *, or ?) when specifying path filters.
> 
> - Paths are always specified relative to the root of the repository.
> - If you don't set path filters, then the root folder of the repo is implicitly included by default.
> - If you exclude a path, you cannot also include it unless you qualify it to a deeper folder. For example if you exclude /tools then you could include /tools/trigger-runs-on-these
> - The order of path filters doesn't matter.
> - Paths in Git are case-sensitive. Be sure to use the same case as the real folders.
> - You cannot use [variables](https://docs.microsoft.com/en-us/azure/devops/pipelines/process/variables?view=azure-devops) in paths, as variables are evaluated at runtime (after the trigger has fired).

The first paragraph is not true. We don't have a branch filter, yet path filters for components are working fine.

The culprit is in the third bullet. To be able to include individual files in `/`, you cannot have an exclude filter on `/` itself. So I changed the exclude filter from `/` into `/src` and `/build`.

I've tested the correctness of these changes in #1002 (see commit history).

## Quality checklist

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -6,7 +6,6 @@ trigger:
     exclude:
     - README.md
     - build/*
-    - project-docs/*
     - roadmaps/*
 
 jobs:
--- a/build/circuitbreaker.yml
+++ b/build/circuitbreaker.yml
@@ -1,12 +1,13 @@
 pr:
   paths:
     exclude:
-    - /
+    - build
+    - src
     include:
-    - /.editorconfig
-    - /stylecop.json
-    - /*.props
-    - /*.ruleset
+    - .editorconfig
+    - stylecop.json
+    - '*.props'
+    - '*.ruleset'
     - build/templates
     - build/circuitbreaker.yml
     - src/CircuitBreaker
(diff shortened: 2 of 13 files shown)


ccheetham (dismissed):
lgtm

ccheetham (approved):
lgtm

PR #1003, opened 2022-08-19
Remove explicit line endings setting from .editorconfig

bart-vmware:
## Description

I recently added `end_of_line = lf` to `.editorconfig` for completeness, but we found that this sometimes results in the following pop-up in Visual Studio on Windows:

image

The reason is that the default git-windows installation converts `\n` to `\r\n` on fetch. Which is okay, because it converts back when pushing to the server.

While the problem goes away by unchecking "Always show this dialog", I'm removing this to avoid the barrier for new contributors. Modern editors can handle different line endings fine, so this is mostly a concern from the past.

Though I've seen failing test assertions on multi-line strings on different OSes, but that's easy to work around.

## Quality checklist

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -8,7 +8,6 @@ indent_style = space
 # (Please don't specify an indent_size here; that has too many unintended consequences.)
 charset = utf-8
 trim_trailing_whitespace = true
-end_of_line = lf
 insert_final_newline = true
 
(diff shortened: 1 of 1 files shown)


ccheetham (approved):
lgtm

ccheetham (approved):
lgtm

PR #1005, opened 2022-08-24
Rename 'config' to 'configuration'

### LINKED ISSUE #998: Decide on the usage of .ConfigureAwait(false) in Steeltoe code
Today the Steeltoe codebase contains limited usage of `.ConfigureAwait(false)`. I'm wondering if that was a deliberate decision, and if so, what the rationale was for sometimes using it and sometimes omitting it. Should we revisit this, now that we're not targeting .NET Framework and NetStandard anymore?

There's quite some contradicting guidance around the topic.
- The blog at https://itnext.io/a-deep-dive-into-configureawait-65f52b9605c2 states that there's no more need for it when targeting only .NET Core:
  > At this moment, one might think that in .NET Core you won’t need to spread `ConfigureAwait(false)` all over your code. Almost!
  > This is almost true, it is still recommended the utilization of `ConfigureAwait(false)` for libraries as a fallback if those libraries are used within a legacy framework. But for most of the cases yes, in `.NET Core` you can drop the ConfigureAwait(false) usage.
- EF Core [recently](https://devblogs.microsoft.com/dotnet/announcing-entity-framework-core-6-0-preview-2/#preserve-synchronization-context-in-savechangesasync) switched to adding `.ConfigureAwait(false)` almost everywhere, despite requiring .NET Core.
- The ASP.NET Core repo [removed](https://github.com/dotnet/aspnetcore/issues/31132) `.ConfigureAwait(false)` from all projects that don't target NetStandard.
- Excerpt from the [ConfigureAwait FAQ](https://devblogs.microsoft.com/dotnet/configureawait-faq/) by Stephen Toub:
  > I’ve heard ConfigureAwait(false) is no longer necessary in .NET Core. True?
  >
  > False. It’s needed when running on .NET Core for exactly the same reasons it’s needed when running on .NET Framework. Nothing’s changed in that regard.

  > What has changed, however, is whether certain environments publish their own SynchronizationContext. In particular, whereas the classic ASP.NET on .NET Framework has [its own SynchronizationContext](https://github.com/microsoft/referencesource/blob/3b1eaf5203992df69de44c783a3eda37d3d4cd10/System.Web/AspNetSynchronizationContextBase.cs), in contrast ASP.NET Core does not. That means that code running in an ASP.NET Core app by default won’t see a custom SynchronizationContext, which lessens the need for ConfigureAwait(false) running in such an environment.

  > It doesn’t mean, however, that there will never be a custom SynchronizationContext or TaskScheduler present. If some user code (or other library code your app is using) sets a custom context and calls your code, or invokes your code in a Task scheduled to a custom TaskScheduler, then even in ASP.NET Core your awaits may see a non-default context or scheduler that would lead you to want to use ConfigureAwait(false). Of course, in such situations, if you avoid synchronously blocking (which you should avoid doing in web apps regardless) and if you don’t mind the small performance overheads in such limited occurrences, you can probably get away without using ConfigureAwait(false).

bart-vmware:
## Description
Various refactorings, see commit messages for details.

Fixes #998.

## Quality checklist


src/Connectors/src/Connector/CosmosDb/CosmosDbTypeLocator.cs
```diff
@@ -3,8 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 using Steeltoe.Common.Reflection;
+using Steeltoe.Connector;
 
-namespace Steeltoe.Connector.CosmosDb;
+namespace Steeltoe.Bootstrap.AutoConfiguration.TypeLocators;
```
TimHess, line 8: My initial reaction to this change is that I don't agree with it. Bootstrap.Autoconfiguration is a completely optional layer on top of Connectors, and these classes are a core piece of Connectors. I could see centralizing all Assembly/Type metadata somewhere in Common maybe, but I'm not sure if that's worth the effort since these aren't likely to be used anywhere else. Outside the scope of this PR, but we might even be better served by marking these `internal`

TimHess: I forgot that we need to use these in AutoConfiguration... so that could explain why this change might make sense from a certain point of view, but I'm not sure I agree with that though

bart-vmware: I didn't realize these files were linked into the project from somewhere else. I've reverted the namespace changes. The types aren't being used from the Bootstrap.Autoconfiguration project, so I'm going to try to remove the linking and see if things still work.

src/Connectors/src/Connector/MongoDb/MongoDbTypeLocator.cs
```diff
@@ -5,7 +5,7 @@
 using System.Reflection;
 using Steeltoe.Common.Reflection;
 
-namespace Steeltoe.Connector.MongoDb;
+namespace Steeltoe.Connector.CosmosDb;
```
TimHess, line 8: I was going to add a suggestion to change this back to `MongoDb`, but it would impact other files too,  and it looks like all of the TypeLocators got moved into `Steeltoe.Connector.CosmosDb` regardless of what they're for

src/Discovery/src/Eureka/EurekaDiscoveryClientExtension.cs
```diff
@@ -51,100 +51,103 @@ public bool IsConfigured(IConfiguration configuration, IServiceInfo serviceInfo
...
-                var config = serviceProvider.GetRequiredService<IConfiguration>();
-                var appInfo = serviceProvider.GetRequiredService<IApplicationInstanceInfo>();
-                var inetOptions = config.GetSection(InetOptions.Prefix).Get<InetOptions>();
-                options.NetUtils = new InetUtils(inetOptions);
-                options.ApplyNetUtils();
-                const string endpointAssembly = "Steeltoe.Management.Endpoint";
-
-                if (ReflectionHelpers.IsAssemblyLoaded(endpointAssembly))
+            .Configure<IConfiguration>((options, configuration) =>
+                configuration.GetSection(EurekaInstanceOptions.EurekaInstanceConfigurationPrefix).Bind(options)).PostConfigure<IServiceProvider>(
```
TimHess, line 72: Was the splitting of `.PostConfigure` from it's param intentional? The diff is a bit hard to read, seems like there must have been a whitespace change that went with it

TimHess (changes requested):

src/Discovery/src/Eureka/EurekaDiscoveryClientExtension.cs
bart-vmware: I'm not sure I understand the question. This is what my diff tool shows when discarding leading whitespace:

image

I did the renames, then ran the auto-reformat, which changed indents in this case, but not the meaning of the code as far as I'm aware. By extracting nested logic in a separate method and using formatter directives, this can be cleaned up to the following:
```c#
// @formatter:wrap_chained_method_calls chop_always
// @formatter:keep_existing_linebreaks true

services
    .AddOptions<EurekaInstanceOptions>()
    .Configure<IConfiguration>((options, configuration) => configuration
        .GetSection(EurekaInstanceOptions.EurekaInstanceConfigurationPrefix)
        .Bind(options))
    .PostConfigure<IServiceProvider>((options, serviceProvider) => NewMethod(serviceProvider, options));

// @formatter:keep_existing_linebreaks restore
// @formatter:wrap_chained_method_calls restore
```

I can make that change, but what should "NewMethod" be named?

src/Connectors/src/Connector/CosmosDb/CosmosDbTypeLocator.cs
bart-vmware: Tried to remove the linking and tested in an empty API project with management endpoints, using the NuGets produced from this PR. It throws a `FileNotFoundException` at startup, referring to the Steeltoe.Connector assembly. Then reverted the changes in this PR and repeated the test, which succeeded.

Why this matters is still a mystery to me, but somehow the linking appears to be essential, despite all unit tests succeeding without them.

src/Discovery/src/Eureka/EurekaDiscoveryClientExtension.cs
TimHess: This is what I saw initially
image

After telling github to hide whitespace changes it's easier to read. The line break changes don't make it easier for _me_ to read to read the code, but it's not that big of a deal if that's what auto-format is doing

TimHess (approved):
LGTM

PR #1007, opened 2022-08-31
Messaging web app extensions

hananiel:
## Description

Add Extensions methods to allow RabbitMQHost to work with `WebApplicationBuilder`.


src/Messaging/src/RabbitMQ/Host/RabbitMQHost.cs
```diff
@@ -29,30 +28,35 @@ public static IHostBuilder CreateDefaultBuilder(string[] args)
         return new RabbitMQHostBuilder(Microsoft.Extensions.Hosting.Host.CreateDefaultBuilder(args));
     }
 
+    public static WebApplicationBuilder CreateWebApplicationBuilder()
```
TimHess, line 31: Can you please add some XML comments to all the public methods?

src/Messaging/src/RabbitMQ/Host/RabbitMQHostExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+using Steeltoe.Messaging.RabbitMQ.Configuration;
+using Steeltoe.Messaging.RabbitMQ.Extensions;
+
+namespace Steeltoe.Messaging.RabbitMQ.Host;
+
+public static class RabbitMQHostExtensions
+{
+    public static void ConfigureRabbitServices(this IServiceCollection services, IConfiguration configuration)
+    {
+        //IConfigurationSection rabbitConfigSection = configuration.GetSection(RabbitOptions.Prefix);
```
TimHess, line 18:
```suggestion
        // IConfigurationSection rabbitConfigSection = configuration.GetSection(RabbitOptions.Prefix);
```

src/Messaging/src/RabbitMQ/Host/RabbitMQHostExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+using Steeltoe.Messaging.RabbitMQ.Extensions;
+
+namespace Steeltoe.Messaging.RabbitMQ.Host;
+
+public static class RabbitMQHostExtensions
+{
+    public static void ConfigureRabbitServices(this IServiceCollection services, IConfiguration configuration)
+    {
+        // IConfigurationSection rabbitConfigSection = configuration.GetSection(RabbitOptions.Prefix);
+        //services.Configure<RabbitOptions>(rabbitConfigSection);
```
TimHess, line 19:
```suggestion
        // services.Configure<RabbitOptions>(rabbitConfigSection);
```

src/Messaging/test/RabbitMQ.Test/Host/RabbitMQHostTest.cs
```diff
@@ -97,4 +98,22 @@ public void ShouldWorkWithRabbitMQConnection()
 
         Assert.NotNull(connectionFactory);
     }
+
+    [Fact]
+    [Trait("Category", "SkipOnMacOS")]
+    [Trait("Category", "SkipOnLinux")]
```
TimHess, line 104: Why is this Windows-only?

src/Messaging/test/RabbitMQ.Test/Host/RabbitMQHostWebApplicationTest.cs
```diff
@@ -0,0 +1,135 @@
...
+    public void WebAppHostShouldAddRabbitOptionsConfiguration()
+    {
+        WebApplicationBuilder builder = RabbitMQHost.CreateWebApplicationBuilder();
+
+        var appSettings = new Dictionary<string, string>
+        {
+            [$"{RabbitOptions.Prefix}:host"] = "ThisIsATest",
+            [$"{RabbitOptions.Prefix}:port"] = "1234",
+            [$"{RabbitOptions.Prefix}:username"] = "TestUser",
+            [$"{RabbitOptions.Prefix}:password"] = "TestPassword"
```
TimHess, line 69: If I understand correctly, this test fails because:

- `RabbitOptions.Prefix` is `spring:rabbitmq`, but Connectors use `rabbitmq:client:` or `rabbit:client`
- Port is never set [here](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Messaging/src/RabbitMQ/Extensions/RabbitServicesExtensions.cs#L238)
- `host` is referred to as `server` in Connectors
- The hostname is converted to lowercase when parsed as a Uri somewhere in connectors

src/Messaging/test/RabbitMQ.Test/Host/RabbitMQHostWebApplicationTest.cs
```diff
@@ -0,0 +1,135 @@
...
+        };
+
+        builder.Configuration.AddInMemoryCollection(appSettings);
```
TimHess, line 72:
```suggestion
        var appSettings = new Dictionary<string, string>
        {
            ["rabbitmq:client:server"] = "ThisIsATest",
            ["rabbitmq:client:port"] = "1234",
            ["rabbitmq:client:username"] = "TestUser",
            ["rabbitmq:client:password"] = "TestPassword"
            //[$"{RabbitOptions.Prefix}:host"] = "ThisIsATest",
            //[$"{RabbitOptions.Prefix}:port"] = "1234",
            //[$"{RabbitOptions.Prefix}:username"] = "TestUser",
            //[$"{RabbitOptions.Prefix}:password"] = "TestPassword"
        };
        WebApplicationBuilder builder = RabbitMQHost.CreateWebApplicationBuilder(configure: configuration => configuration.AddInMemoryCollection(appSettings));

        using WebApplication webApp = builder.Build();
        RabbitOptions rabbitOptions = webApp.Services.GetService<IOptions<RabbitOptions>>()?.Value;

        Assert.NotNull(rabbitOptions);
        Assert.Equal("thisisatest", rabbitOptions.Host);
        Assert.Equal(1234, rabbitOptions.Port);
        Assert.Equal("TestUser", rabbitOptions.Username);
        Assert.Equal("TestPassword", rabbitOptions.Password);
```
Not sure if this is exactly what we're after, but this test does pass for me

TimHess: The change on L73 is notable... without it, the dictionary is not passed to Connectors

hananiel: And then using a RabbitMQHost will not be similar to using generic host, because adding the configuration after (as this test was verifying before modification) no longer works. So it seems we can have one or other but not both. This is a leaky abstraction and I am wondering if we should either move away from it altogether or only have it work for some of the use cases (without connectors). The latter is what I am leaning towards.

hananiel:
After considering the implementation of WebApplicationBuilder, and the existing Messaging and Stream specific hosts, it would be best not to add an overload to provide a WebApplicationBuilder. The best way to add this functionality would be using the RabbitMQHost and StreamHost and configuring the WebHost or to use the minimal WebApplicationBuilder extensions and add Messaging/Stream related services using one of the existing steeltoe provided extension methods.

PR #1021, opened 2022-09-28
Handle preinit events

### LINKED ISSUE #985: EventCounterListener throws initialization errors
Ocassionally the `OnEventWritten` gets fired before constructor and throws errors. Need to investigate if related to build failures for example: 
https://dev.azure.com/SteeltoeOSS/Steeltoe/_build/results?buildId=9718&view=ms.vss-test-web.build-test-results-tab&runId=1465457&resultId=100245&paneView=debug

The fix is to wait for initialization before attempting processing and also add a thread safe collection to store until initialization.

hananiel:
resolves #985

### DIFF
--- a/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
+++ b/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
@@ -39,7 +39,7 @@ public class EventCounterListener : EventListener
         _logger = logger;
         _isInitialized = true;
 
-        // ProcessPreInitEventSources();
+        ProcessPreInitEventSources();
     }
 
     /// <summary>
@@ -81,12 +81,22 @@ public class EventCounterListener : EventListener
             throw new ArgumentNullException(nameof(eventSource));
         }
 
-        if (!_isInitialized)
-        {
-           _eventSources.Add(eventSource);
-        }
-
         if (_eventSourceName.Equals(eventSource.Name, StringComparison.OrdinalIgnoreCase))
+        {
+            if (!_isInitialized)
+            {
+                _eventSources.Add(eventSource);
+            }
+            else
+            {
+                SafeEnableEvents(eventSource);
+            }
+        }
+    }
+
+    private void ProcessPreInitEventSources()
+    {
+        foreach (var eventSource in _eventSources)
         {
             SafeEnableEvents(eventSource);
         }
--- a/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
+++ b/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
@@ -23,7 +23,7 @@ public class EventCounterListener : EventListener
     private readonly IMetricsObserverOptions _options;
     private readonly bool _isInitialized = false;
 
-    private readonly Dictionary<string, string> _refreshInterval = new () { { "EventCounterIntervalSec", "1" } }; // TODO: Make it configurable
+    private readonly Dictionary<string, string> _refreshInterval = new () { { "EventCounterIntervalSec", "1" } };
 
     private ConcurrentDictionary<string, ObservableGauge<double>> _doubleMeasureMetrics = new ();
     private ConcurrentDictionary<string, ObservableGauge<long>> _longMeasureMetrics = new ();
--- a/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
+++ b/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
@@ -38,7 +38,8 @@ public class EventCounterListener : EventListener
         _options = options ?? throw new ArgumentNullException(nameof(options));
         _logger = logger;
         _isInitialized = true;
-        ProcessPreInitEventSources();
+
+        // ProcessPreInitEventSources();
     }
 
     /// <summary>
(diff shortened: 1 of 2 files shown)


src/Management/test/KubernetesCore.Test/HostBuilderExtensionsTest.cs
```diff
@@ -21,6 +21,7 @@ namespace Steeltoe.Management.Kubernetes.Test;
 public class HostBuilderExtensionsTest
 {
     [Fact]
+    [Trait("Category", "SkipOnLinux")]
```
TimHess, line 24: Are you sure skipping is the best course of action here? Have you confirmed that your changes are not related to these failing?

hananiel: No that was for debugging. Now the bug is fixed and the skipping removed.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
```diff
@@ -21,17 +21,25 @@ public class EventCounterListener : EventListener
     private readonly string _eventSourceName = "System.Runtime";
     private readonly string _eventName = "EventCounters";
     private readonly IMetricsObserverOptions _options;
+    private readonly bool _isInitialized = false;
+
+    private readonly Dictionary<string, string> _refreshInterval = new () { { "EventCounterIntervalSec", "1" } }; // TODO: Make it configurable
```
bart-vmware, line 26: Should this TODO be addressed in this PR? If not, can you remove it from the code and create an issue for it?

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
```diff
@@ -70,18 +83,37 @@ protected override void OnEventSourceCreated(EventSource eventSource)
...
+
+    private void SafeEnableEvents(EventSource eventSource)
+    {
+        try
+        {
+            EnableEvents(eventSource, EventLevel.Verbose, EventKeywords.All, _refreshInterval);
+        }
+        catch (Exception ex)
+        {
+            _logger?.LogError(ex.Message, ex);
```
bart-vmware, line 113:
```suggestion
            _logger?.LogError($"Failed to enable events: {ex.Message}", ex);
```

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
```diff
@@ -21,17 +21,25 @@ public class EventCounterListener : EventListener
     private readonly string _eventSourceName = "System.Runtime";
     private readonly string _eventName = "EventCounters";
     private readonly IMetricsObserverOptions _options;
+    private readonly bool _isInitialized = false;
```
bart-vmware, line 24: I think this can be removed. It is always set to `true` from the constructor, and there are no other constructors, so it will never be `false`.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
hananiel: I think its outside the scope of this change - Ill create an issue.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
hananiel: There is a wierd case of of timing because of how event listener subscribers work. So this code is needed because the event is fired before the constructor is run.

bart-vmware: Can you elaborate on that? As far as I'm aware, its impossible in C# to change the value of a readonly field.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
bart-vmware: Thanks for the update. Next time, can you let me resolve the conversation I initiated, so I can verify the change?

hananiel: I accepted your change and I believe it automatically resolved it.

bart-vmware: Oh, I didn't know it worked that way. In that case, it's fine.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
hananiel: So the timing as can be observed from this class, 

1.  the fields are initialized 
2. OnEventWritten(EventWrittenEventArgs eventData) gets invoked. ( at this time , is initialized is false, and the events are added to a bag) 
3. Constructor runs (isInitialized is true) - bag is processed
4. OnEventWritten fires again and this time isInitialized is true, and bag is not updated and the event is processed.

bart-vmware: We discussed this during a meeting. I suspect the base class constructor invokes `OnEventWritten`, at which point the derived constructor hasn't run. Either way, the code works as intended, so I'm closing this.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
hananiel: Can we resolve this conversation?

TimHess (approved):

TimHess:
@hananiel does this need to be brought in 4.0 too?

PR #1027, opened 2022-10-04
Update version so the NuGet build of main produces 4.0.x instead 4.0.0

bart-vmware:
## Description

This attempts to fix producing duplicate NuGet packages in dev feed when `main` is built. At the time we're going to publish v4.0.0 to NuGet.org, we should revert this change.

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,7 +1,7 @@
 {
   "$schema":
     "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "4.0.0-alpha",
+  "version": "4.0-alpha",
   "publicReleaseRefSpec": [
     "^refs/heads/main$",
     "^refs/heads/release/\\d+(?:\\.\\d+)?$"


version.json
```diff
@@ -1,7 +1,7 @@
 {
   "$schema":
     "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "4.0.0-alpha",
+  "version": "4.0-alpha",
```
TimHess, line 4:
```suggestion
  "version": "4.0.0-alpha.{height}",
```
Will something like this work?  Found [here](https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/versionJson.md), I'm not sure yet what additional options are available

bart-vmware: I don't think that's going to work. `"4.0-alpha"` implicitly means `"4.0.{height}-alpha"` We need the variation at the patch level in major.minor.patch for NuGet packages to version on subsequent commits. Also, the height should be sorted on before nested sorting on `alpha`, `beta`, `rc` and so on occurs.

When the final version of v4 is released, we're going to drop the `alpha` part, which results in `4.0.0.{height}` based on your suggestion. That produces duplicate NuGet packages again.

TimHess: Because these are CI builds and not true public releases, I think we need the variation after the dash instead of in the major.minor.patch. I think it will be confusing and problematic for anybody using the dev feed if our CI builds go 4.0.2-alpha, 4.0.10-alpha, 4.0.50-alpha, etc. before we change to 4.0.0

TimHess: I think we need to be producing pre-release numbers with dot notation... see [docs](https://learn.microsoft.com/en-us/nuget/create-packages/prerelease-packages) for details
> NuGet 4.3.0+ supports [Semantic Versioning v2.0.0](https://semver.org/spec/v2.0.0.html), which supports pre-release numbers with dot notation, as in 1.0.1-build.23. Dot notation is not supported with NuGet versions before 4.3.0. In earlier versions of NuGet, you could use a form like 1.0.1-build23 but this was always considered a pre-release version.

bart-vmware: That still won't solve duplicate non-pre packages. See https://github.com/dotnet/Nerdbank.GitVersioning/blob/main/doc/public_vs_stable.md for details on how versioning works with NBGV and the distinction between public and stable versions.

Perhaps we should just go with their scheme and drop the idea that the final 4.0 needs to be v4.0.0. This would imply that a released stable version like v4.0.87653 means: v4.0 with some kind of ever-incrementing build ID. That makes it easy to infer where an alpha/beta version fits in between stable versions.

TimHess: I think we should change `publicReleaseRefSpec` to only produce public releases when a tag is present. Given that we always tag releases before shipping them to nuget.org it seems to align with our process

TimHess: Nevermind, let's just let Nerdbank do the work the way it is built to do it

PR #1032, opened 2022-10-19
Allow actuators to be exposed on an additional (optionally ssl) port

### LINKED ISSUE #369: Actuator Endpoints should listen on a different port than application port
Support listening to actuators on a separate port than Application port.

hananiel:
Resolves #369 

To expose management on a different port a user has to:
1. Use any extension method to `AddXActuator` or `AddAllActuators`
2. Set "management:endpoints:port"=<port#> for ex. 9090
3. Additionally you can force ssl by setting : "management:endpoints:sslenabled"=true; 
4. AddAllActuators and AddXActuator no longer need to call UseCloudHosting in any cloud environment.

I added some internal extensions to Steeltoe.Common.Hosting to allow the Steeltoe.Management.Endpoint package to add additional Urls. So simply adding (or leaving) the management:endpoints:port setting  appsettings.json, should not expose more ports/urls.  

`AddAllActuators` and `AddXActuator` no longer need to call `UseCloudHosting` in any cloud environment, ex TAS, TAP. There was an issue previously with this extension when running locally, that it would override the settings in launchsettings.json, but adding the environment ASPNETCORE_URLs seems to resolve this issue.


src/Common/src/Common.Hosting/HostBuilderExtensions.cs
```diff
@@ -13,7 +13,9 @@ public static class HostBuilderExtensions
     public const string DefaultUrl = "http://*:8080";
 
     /// <summary>
-    /// Configure the application to listen on port(s) provided by the environment at runtime. Defaults to port 8080.
+    /// Configure the application to listen on port(s) provided by the environment at runtime. Specifically it adds ports and/or urls from the following
+    /// environment variables: PORT, SERVERPORT, ASPNETCORE_URLS and in Kubernetes $"{appName}_SERVICE_PORT_HTTP") where appName is the prefix of HOSTNAME.
+    /// It also opens the application on the managementPort is expected to be used in conjuction with the ManagementPortMiddlware. Defaults to port 8080.
```
bart-vmware, line 18:
```suggestion
    /// It also opens the application on the management port, which is expected to be used in conjunction with <see cref="ManagementPortMiddleware" />. Defaults to port 8080.
```

bart-vmware (commented):
Looks good to me, aside from minor misspellings in docs.

src/Common/src/Common.Hosting/HostBuilderExtensions.cs
hananiel: The cref is to a type thats not available in this project. Other than that thanks - I will fix it as suggested.

src/Management/src/Endpoint/ManagementWebApplicationBuilderExtensions.cs
```diff
@@ -32,6 +33,7 @@ public static class ManagementWebApplicationBuilderExtensions
     /// </param>
     public static WebApplicationBuilder AddDbMigrationsActuator(this WebApplicationBuilder applicationBuilder)
     {
+        applicationBuilder.UseCloudHosting(ManagementWebHostBuilderExtensions.ConfigureManagementUrls);
```
TimHess, line 36: As you've previously commented, the methods in this file are a bit brute-force/repetetive... would it make sense to add a private method that calls `UseCloudHosting`, `ActivateActuatorEndpoints` (and any other future common calls) and returns the applicationBuilder as a short-term solution to reduce some of the bulk?

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
```diff
@@ -315,4 +322,39 @@ public static IWebHostBuilder AddWavefrontMetrics(this IWebHostBuilder hostBuild
     {
         return hostBuilder.ConfigureServices((_, collection) => collection.AddWavefrontMetrics());
     }
+
+    internal static void ConfigureManagementUrls(this IWebHostBuilder webHostBuilder, List<string> urls)
+    {
+        string managementPort = webHostBuilder.GetSetting(ManagementPortKey);
```
TimHess, line 328: Are there scenarios where this method of getting these settings do work? When would this work as opposed to looking for an appsettings.json file directly?

src/Management/test/Endpoint.Test/ManagementEndpointServedOnDifferentPort.cs
```diff
@@ -0,0 +1,161 @@
...
+using Steeltoe.Common.Hosting;
+using Xunit;
+
+namespace Steeltoe.Management.Endpoint.Test;
+
+public class ManagementEndpointServedOnDifferentPort
+{
+    public ManagementEndpointServedOnDifferentPort()
+    {
+        ClearEnvVars();
```
TimHess, line 22: Does this need to be called at the start and end of every test in this class? We could [use IDisposable](https://xunit.net/docs/shared-context) to clean up instead of calling it directly each time

src/Management/test/Endpoint.Test/Refresh/EndpointMiddlewareTest.cs
```diff
@@ -60,6 +60,8 @@ public async Task RefreshActuator_ReturnsExpectedData()
         string ancEnv = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
         Environment.SetEnvironmentVariable("ASPNETCORE_ENVIRONMENT", null);
 
+        Environment.SetEnvironmentVariable("ASPNETCORE_URLS", null);
```
TimHess, line 63: Is this needed due to other tests leaking?

src/Management/test/Endpoint.Test/Env/EndpointMiddlewareTest.cs
```diff
@@ -64,6 +64,8 @@ public async Task EnvActuator_ReturnsExpectedData()
         string originalEnv = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
         Environment.SetEnvironmentVariable("ASPNETCORE_ENVIRONMENT", null);
 
+        Environment.SetEnvironmentVariable("ASPNETCORE_URLS", null);
```
TimHess, line 67: Is this needed due to other tests leaking?

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
hananiel: When using the WebApplication Builder and also when running tests,  these settings are  available. Specifically when testing, it keeps us from having to add an appsettings.json.
We need the appsettings.json when using the generic hostbuilder.

src/Management/test/Endpoint.Test/Refresh/EndpointMiddlewareTest.cs
hananiel: Yes.

src/Management/test/Endpoint.Test/ManagementEndpointServedOnDifferentPort.cs
hananiel: Its mostly with the start - I sometimes added this at the end, but only some tests are sensitive to this and I added it to the either start or in this case constructor.

src/Management/test/Endpoint.Test/Env/EndpointMiddlewareTest.cs
hananiel: Yes - buts not just the ones in the ManagementPortTest class, its all the actuators - too much change to handle by implementing IDisposable etc everywhere - cleaner to just do this in the tests that are sensitive.

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
TimHess: This particular situation has not changed, @dtillman what are your thoughts? If `builder.GetSetting` worked reliably then I'd be fine with it, but what happens next is a fallback that manually looks for `appsettings.json` which is not going to be able to build the same view of the app's config as a developer might expect. 

This discussion item has become disconnected from the file, which is [here](https://github.com/SteeltoeOSS/Steeltoe/pull/1032/files#diff-f0f43afe1e468a0f4913c7ac4ee966246fa6691fd5b11e51731d5b7a42c3491bR328)

dtillman: I'll dig into this next .. finishing up the code for the last RabbitMQ tutorial

hananiel:
Redoing via #1046 and another future PR

PR #1036, opened 2022-10-25
Health check updates

hananiel:
Resolves issues with HealthCheck output and adds pingContributor.

### DIFF
--- a/src/Management/src/EndpointBase/Health/HealthConverterV3.cs
+++ b/src/Management/src/EndpointBase/Health/HealthConverterV3.cs
@@ -41,7 +41,7 @@ public class HealthConverterV3 : JsonConverter<HealthEndpointResponse>
                 writer.WriteEndObject();
             }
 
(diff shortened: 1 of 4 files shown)


src/Management/src/EndpointBase/Health/HealthConverterV3.cs
```diff
@@ -38,6 +40,12 @@ public override void Write(Utf8JsonWriter writer, HealthEndpointResponse value,
 
                 writer.WriteEndObject();
             }
+
+            if (health.Groups != null && health.Groups.Count() > 0)
```
TimHess, line 44:
```suggestion
            if (health.Groups?.Any() == true)
```
Typed in the browser, might want to double-check I got it right before commiting... `.Any` being the main change.  The addition of this property seems like something that should have a test or an assertion too

dtillman (dismissed):
LGTM

TimHess (approved):

Issue #1043, opened 2022-11-08
Config Server Logging is not working consistently


TimHess:
## Describe the bug
Log entries are not generated when `.AddConfigServer(loggerFactory)` is used. The LoggerFactory param is not passed around everywhere it needs to go

## Steps to reproduce 
This code should result in logs relating to config server interactions hitting the console:
```csharp
        public static void Main(string[] args)
        {
            BuildWebHost(args).Run();
        }

        public static IWebHost BuildWebHost(string[] args)
        {
            var loggerFactory = LoggerFactory.Create(loggingBuilder => loggingBuilder
                .SetMinimumLevel(LogLevel.Trace)
                .AddConsole());

            return WebHost.CreateDefaultBuilder(args)
                .AddConfigServer(loggerFactory)
                .UseStartup<Startup>()
                .Build();
        }
```

## Expected behavior
With a bad config server entry, logs like this should be visible:
```text
info: Steeltoe.Extensions.Configuration.ConfigServer.ConfigServerConfigurationProvider[0]
      Fetching config from server at: http://localhost:8888/badPath
info: Steeltoe.Extensions.Configuration.ConfigServer.ConfigServerConfigurationProvider[0]
      Config Server returned status NotFound invoking path http%3A%2F%2Flocalhost%3A8888%2FbadPath%2Ffoo%2FDevelopment
warn: Steeltoe.Extensions.Configuration.ConfigServer.ConfigServerConfigurationProvider[0]
      Could not locate PropertySource:
```
## Environment (please complete the following information):
 - Steeltoe Version 3.2.1
 - Platform: local
 - OS: any
 - .NET Version: any supported

Issue #1031, opened 2022-10-12
Config Server Client does not use Token setting from code


TimHess:
From [SteeltoeOSS slack conversation](https://steeltoeteam.slack.com/archives/C0ENXS0M8/p1665587855497699)

"I am trying to move the SCCS provider config to code only, vs using appSettings.config like this:"
```
var configServerSettings = new ConfigServerClientSettings(){
  Environment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT"),
  Label = Environment.GetEnvironmentVariable("SCCS_LABEL"),
  Token = "a valid vault token",
  Uri = Environment.GetEnvironmentVariable("SCCS_URI"),
  FailFast = true,
  RetryEnabled = true,
  Name = "App Name"
};

builder.Configuration.AddConfigServer(configServerSettings);
```
"However, it seems when doing it like this, the vault token header is not being passed or set based on the SCCS logs."

For v4, fix [this line](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Configuration/src/ConfigServer/ConfigurationSettingsHelper.cs#L41), for v3 fix [this line](https://github.com/SteeltoeOSS/Steeltoe/blob/release/3.2/src/Configuration/src/ConfigServerBase/ConfigurationSettingsHelper.cs#L52). In both cases it should be as simple as looking at the settings object for a value (like all the other lines in that method)

PR #1044, opened 2022-11-08
Config Server Token, Logging Fixes

TimHess:
## Description

Fixes for these minor bugs in 3.2.1: #1043, #1031 

Changes need to be brought forward into 4.0 as well

## Quality checklist

### DIFF
--- a/src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
+++ b/src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
@@ -4,8 +4,6 @@
 
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
-using Microsoft.Extensions.Logging.Abstractions;
-using Microsoft.Extensions.Primitives;
 using Steeltoe.Common.Discovery;
 using Steeltoe.Common.Http;
 using Steeltoe.Common.Logging;
@@ -307,7 +305,7 @@ public class ConfigServerConfigurationProvider : ConfigurationProvider, IConfigu
                 if (env != null)
                 {
                     _logger.LogInformation(
-                        "Located environment: {name}, {profiles}, {label}, {version}, {state}", env.Name, env.Profiles, env.Label, env.Version, env.State);
+                        "Located environment name: {name}, profiles: {profiles}, labels: {label}, version: {version}, state: {state}", env.Name, env.Profiles, env.Label, env.Version, env.State);
                     if (updateDictionary)
                     {
                         var data = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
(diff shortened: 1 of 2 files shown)


src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
```diff
@@ -588,7 +589,7 @@ protected internal async Task<ConfigEnvironment> RemoteLoadAsync(string[] reques
                     if (response.StatusCode >= HttpStatusCode.BadRequest)
                     {
                         // HttpClientErrorException
-                        throw new HttpRequestException(message);
+                        throw new HttpRequestException(string.Format(statusMessage, response.StatusCode, WebUtility.UrlEncode(requestUri)));
```
bart-vmware, line 592: If you'd inline `statusMessage`, analyzers validate that the placeholders match the parameters. In the case of `LogInformation`, using `{Status}` and `{Path}` instead of `{0}` and `{1}` enables named parameters when using JSON logging. The `throw` line could use string interpolation instead, which is more efficient in .NET 6.

src/Configuration/src/ConfigServerBase/ConfigurationSettingsHelper.cs
```diff
@@ -122,9 +122,9 @@ private static bool GetEnabled(IConfigurationSection clientConfigsection, bool d
         return clientConfigsection.GetValue("enabled", def);
     }
 
-    private static string GetToken(IConfigurationSection clientConfigsection)
+    private static string GetToken(IConfigurationSection clientConfigsection, string def)
```
bart-vmware, line 125: I'm assuming that `def` means `defaultValue`?

src/Configuration/src/ConfigServerBase/ConfigServerConfigurationProvider.cs
```diff
@@ -645,22 +648,20 @@ protected internal virtual async Task<ConfigEnvironment> RemoteLoadAsync(string
         try
         {
             using var response = await _httpClient.SendAsync(request).ConfigureAwait(false);
+
+            // Log status
```
bart-vmware, line 652: I'm not a fan of these comments that describe what the code does. Would you consider deleting them?

src/Configuration/src/ConfigServerBase/ConfigurationSettingsHelper.cs
TimHess: Yeah. Not the best, but I was just matching the methods around it and figured we'd make sure it all looks better when bringing the change to 4.0

bart-vmware: I'm okay with that.

bart-vmware (approved):
LGTM

PR #1046, opened 2022-11-09
Refactor use cloud hosting

hananiel:
This Change 
1. Removes support for local ports (helps with complexity) - there are already so many ways to set local ports, and we are primarily adding new interpretations of env variables through  UseCloudHosting.
2. Remove Generic HostBuilder Extensions - while a generic hostbuilder can have a webhost it is not required. 
3. Make the method stateful so that subsequent invocations (for example from an actuator setup call can add a management port without conflicting or wiping out current state) - This additional functionality will come in a separate PR.


src/Common/src/Common.Hosting/HostBuilderExtensions.cs
```diff
@@ -104,28 +62,22 @@ private static IWebHostBuilder BindToPorts(this IWebHostBuilder webHostBuilder,
...
+        return webHostBuilder.BindToPorts(urls);
+    }
+
+    private static IWebHostBuilder BindToPorts(this IWebHostBuilder webHostBuilder, List<string> urls)
+    {
+        string currentSetting = webHostBuilder.GetSetting(WebHostDefaults.ServerUrlsKey);
+        List<string> currentUrls = currentSetting?.Split(';').ToList() ?? new List<string>();
+        currentUrls.AddRange(urls);
+        IEnumerable<string> distinctUrls = currentUrls.Distinct();
+        return webHostBuilder.UseSetting(WebHostDefaults.ServerUrlsKey, string.Join(";", distinctUrls));
```
bart-vmware, line 80: Not sure if it matters much here, but if using a `HashSet<>` instead of `List<>`, there's no need to call `.Distinct()` which copies the collection.

hananiel: Bart, yes that makes sense - I made that change.

PR #1047, opened 2022-11-10
Address S2955: Comparing unconstrained generic values against null

### LINKED ISSUE #933: Address S2955: Generic parameters not constrained to reference types should not be compared to "null"
Address existing violations of [S2955: Generic parameters not constrained to reference types should not be compared to "null"](https://rules.sonarsource.com/csharp/RSPEC-2955) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

bart-vmware:
## Description

For some cases, I was able to add a `class` constraint to the generic type parameter to make the warning go away.
For the remaining ones, it resulted in errors, so for now, I've used a hack (`is null` instead of `== null`), which Sonar does not detect. Please advise on a case-by-case basis whether it should a) compare against `default` or b) keep checking against `null` or c) remove the check altogether.

Fixes #933

## Quality checklist


src/CircuitBreaker/src/Hystrix/Collapser/RequestBatch.cs
```diff
@@ -74,7 +74,7 @@ public int Size
                     collapsedRequest.CompletionSource = tcs;
 
                     CollapsedRequest<TRequestResponse, TRequestArgument> existing =
-                        arg == null ? GetOrAddNullArg(collapsedRequest) : _argumentMap.GetOrAdd(arg, collapsedRequest);
+                        arg is null ? GetOrAddNullArg(collapsedRequest) : _argumentMap.GetOrAdd(arg, collapsedRequest);
```
bart-vmware, line 77: @dtillman `arg` is of unconstrained generic type `TRequestArgument`, so comparing to `null` doesn't make sense when it's a value type (resulting in S2955 warning). How do you think this should be solved? I've tried adding a `class` constraint, but that results in errors.

src/Common/src/Common.Expression/Internal/Spring/Ast/Indexer.cs
```diff
@@ -286,7 +286,7 @@ private T ConvertValue<T>(ITypeConverter converter, object value)
         Type targetType = typeof(T);
         var result = (T)converter.ConvertValue(value, value == null ? typeof(object) : value.GetType(), targetType);
 
-        if (result == null)
+        if (result is null)
```
bart-vmware, line 289: @dtillman Same question here. Allowing value types seems intentional.

src/Messaging/src/RabbitMQ/Core/RabbitTemplate.cs
```diff
@@ -2266,7 +2266,7 @@ private void LogReceived(IMessage message)
 
         TReply reply = receiveAndReplyCallback(messageAsTReceive);
 
-        if (reply != null)
+        if (reply is not null)
```
bart-vmware, line 2269: @hananiel `reply` is of unconstrained generic type `TReply`, so comparing to null doesn't make sense when it's a value type (resulting in S2955 warning). How do you think this should be solved? I've tried adding a class constraint, but that results in errors.

bart-vmware: @dtillman

dtillman (approved):

PR #1049, opened 2022-11-11
Fix S4487: Unread "private" fields should be removed

### LINKED ISSUE #954: Address S4487: Unread "private" fields should be removed
Address existing violations of [S4487: Unread "private" fields should be removed](https://rules.sonarsource.com/csharp/RSPEC-4487) in the codebase and set severity to `Info` in `Steeltoe.Debug.ruleset` and to `Warning` in `Steeltoe.Release.ruleset`.

Note: The difference in severities between Debug/Release is to avoid distraction or hindrance while writing/debugging code.

- Remove the field and its assignments
- Add code that reads the field

bart-vmware:
## Description

This PR removes private fields that are unused or only assigned but never read.

If during review, you find any cases where the field should not be removed, please indicate how to update existing code to read from the field and use its value.

Fixes #954.

### DIFF
--- a/shared-package.props
+++ b/shared-package.props
@@ -53,6 +53,6 @@
 
   <PropertyGroup Condition="!$(MSBuildProjectName.StartsWith('Steeltoe.Configuration'))">
     <!-- Narrow the condition above as we're completing more public API reviews -->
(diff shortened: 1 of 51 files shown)


bart-vmware:
As @hananiel pointed out elsewhere, it may be necessary to keep fields of reference types that aren't read from, just to extend their lifetime and prevent them from being garbage-collected. That's a legitimate case to suppress warnings.

dtillman (dismissed):

TimHess (approved):

PR #1050, opened 2022-11-14
Fix: Allow multiple IDynamicMessageProcessor instances to be registered

bart-vmware:
## Quality checklist


hananiel:
@bart-vmware Can you add some more context here - are you doing this to match the dependency retreival side. I realize it supports multiple processors but we never use it, but not clear whether this is needed or not. If Do we need to add a test for any breaking functionality?

bart-vmware:
This change was suggested by @TimHess. The consuming code injects `IEnumerable<IDynamicMessageProcessor>` and iterates over them when building the log message. This change allows Steeltoe consumers to [add](https://stackoverflow.com/questions/67611918/what-is-the-precise-difference-between-tryaddenumerableservicedescriptor-and-o) additional message processors, which we assumed to be the intended behavior.

dtillman (approved):

PR #1053, opened 2022-11-14
Address the use of URIs in relation to Sonar rules

### LINKED ISSUE #945: Address S3997/S4005: Call Uri-based overloads instead of string-based
Address existing violations of the following, related rules:
- [S3997: String URI overloads should call "System.Uri" overloads](https://rules.sonarsource.com/csharp/RSPEC-3997)
- [S4005: "System.Uri" arguments should be used instead of strings](https://rules.sonarsource.com/csharp/RSPEC-4005)

in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

  - When the method calls into a .NET method that has an overload for Uri, call that instead and update the method signature to take an Uri instead of a string. Consider adding an overload that takes a string, which calls this method

Note: This issue mitigates security risks, though it requires more thorough investigation on the best path forward. Should we enable (some of) the rules or are there too many false positives, leading to numerous suppressions?

bart-vmware:
## Description

As it turns out, Steeltoe accepts URLs that are discarded by `System.Uri`, for example:
```
mongodb://a9s-brk-usr-e74b9538:a9sb8b69cc@mongodb-0.node.dc1.a9s-mongodb-consul:27017,mongodb-1.node.dc1.a9s-mongodb-consul:27017,mongodb-2.node.dc1.a9s-mongodb-consul:27017/d5584e9?replicaSet=rs0
```

which fails with:
```
System.UriFormatException: Invalid URI: Invalid port specified.
```

That makes it dangerous to activate the next rules, which is why I've kept them deactivated:
- [S3994: URI Parameters should not be strings](https://rules.sonarsource.com/csharp/RSPEC-3994)
- [S3995: URI return values should not be strings](https://rules.sonarsource.com/csharp/RSPEC-3995)
- [S3996: URI properties should not be strings](https://rules.sonarsource.com/csharp/RSPEC-3996)

This PR fixes the following rules:
- [S3997: String URI overloads should call "System.Uri" overloads](https://rules.sonarsource.com/csharp/RSPEC-3997)
- [S4005: "System.Uri" arguments should be used instead of strings](https://rules.sonarsource.com/csharp/RSPEC-4005)

In addition, it renames namespaces that contain "Task" to "Tasks" to avoid a clash with `System.Threading.Tasks.Task` and it reduces the use of ` task.GetAwaiter().GetResult()` in tests.

Fixes #945

## Quality checklist


TimHess:
Changes generally look good to me, but I don't think it should close #945 though since it only addresses a portion of it.

Are there violations of S3994-3996 outside of Connectors that could be fixed?

TimHess (approved):
LGTM

bart-vmware:
@Tim as we discussed, I've created https://github.com/SteeltoeOSS/Steeltoe/issues/1069 to separately address S3994-3996.

PR #1054, opened 2022-11-14
Metrics updates

hananiel:
## Description
Adds metrics for clr.cpu.count 
Adds an AllowList Option to only add metrics needed, as opposed to listing all unneeded metrics
for example: 
```json
 "Metrics":{                                              
   "Observer":{                                 
     "EventCounterEvents": true,                
     "IncludedMetrics": ["cpu-usage"]           
     }                                            
  }                                              
```
## Quality checklist

### DIFF
--- a/src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
+++ b/src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
@@ -20,6 +20,15 @@ public interface IMetricsObserverOptions
     string EgressIgnorePattern { get; }
 
     /// <summary>
+    /// Gets Allow list of metrics that should not be captured
+    /// </summary>
+    /// <remarks>
+    ///     Currently only applies to System.Runtime metrics captured by <see cref="EventCounterListener"/><para />
+    ///     See this list for values to choose from: <see href="https://docs.microsoft.com/dotnet/core/diagnostics/available-counters#systemruntime-counters" />
+    /// </remarks>
+    List<string> IncludedMetrics { get; }
+
+    /// <summary>
     /// Gets a list of metrics that should not be captured
     /// </summary>
     /// <remarks>
--- a/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
+++ b/src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
@@ -12,6 +12,7 @@ using System.Collections.Generic;
 using System.Diagnostics.Metrics;
 using System.Diagnostics.Tracing;
 using System.Globalization;
+using System.Linq;
 
 namespace Steeltoe.Management.Endpoint.Metrics.Observer;
 
@@ -121,23 +122,20 @@ public class EventCounterListener : EventListener
         long? longValue = null;
         var counterName = string.Empty;
         var labelSet = new List<KeyValuePair<string, object>>();
-        var excludedMetric = false;
+
         string counterDisplayUnit = null, counterDisplayName = null;
         foreach (var payload in eventPayload)
         {
-            if (excludedMetric)
-            {
-                break;
-            }
-
             var key = payload.Key;
             switch (key)
             {
                 case var kn when key.Equals("Name", StringComparison.OrdinalIgnoreCase):
                     counterName = payload.Value.ToString();
-                    if (_options.ExcludedMetrics.Contains(counterName))
+                    if ((_options.IncludedMetrics.Any()
+                        && !_options.IncludedMetrics.Contains(counterName))
+                        || _options.ExcludedMetrics.Contains(counterName))
                     {
-                        excludedMetric = true;
+                        return;
                     }
 
                     break;
--- a/src/Management/src/EndpointBase/Metrics/MetricsObserverOptions.cs
+++ b/src/Management/src/EndpointBase/Metrics/MetricsObserverOptions.cs
@@ -66,5 +66,8 @@ public class MetricsObserverOptions : IMetricsObserverOptions
     public bool HystrixEvents { get; set; } = false;
 
     /// <inheritdoc/>
+    public List<string> IncludedMetrics { get; set; } = new List<string>();
+
+    /// <inheritdoc/>
     public List<string> ExcludedMetrics { get; set; } = new List<string>();
 }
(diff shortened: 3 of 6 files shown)


src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
```diff
@@ -19,6 +19,15 @@ public interface IMetricsObserverOptions
     /// </summary>
     string EgressIgnorePattern { get; }
 
+    /// <summary>
+    /// Gets Allow list of metrics that should be captured
+    /// </summary>
+    /// <remarks>
+    ///     Currently only applies to System.Runtime metrics captured by <see cref="EventCounterListener"/><para />
+    ///     See this list for values to choose from: <see href="https://docs.microsoft.com/dotnet/core/diagnostics/available-counters#systemruntime-counters" />
+    /// </remarks>
```
bart-vmware, line 28: I suspect this is going to fail in the main branch because of [SA1629](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1629.md).

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
```diff
@@ -121,23 +122,20 @@ private void ExtractAndRecordMetric(string eventSourceName, IDictionary<string,
...
-
             var key = payload.Key;
             switch (key)
             {
                 case var kn when key.Equals("Name", StringComparison.OrdinalIgnoreCase):
                     counterName = payload.Value.ToString();
-                    if (_options.ExcludedMetrics.Contains(counterName))
+                    if ((_options.IncludedMetrics.Any()
+                        && !_options.IncludedMetrics.Contains(counterName))
+                        || _options.ExcludedMetrics.Contains(counterName))
```
bart-vmware, line 136: When an entry exists in both include/exclude lists, it will be included. Is that intentional, or should the exclude list win over inclusions? Either way, it would be helpful to document which list has precedence.

hananiel: Yes, Include takes precedence in mind - since its more exclusive.

src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
TimHess: For summary, remarks or both? Want to add it as a suggestion?

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
TimHess: https://github.com/SteeltoeOSS/Documentation/issues/288

TimHess (approved):
looks OK to me, wouldn't be a bad idea to address the sonar issue now but isn't _required_ in this branch

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
bart-vmware: I didn't mean to create external documentation. Just a simple line like "In case of conflict, see [ref IncludedMetrics] takes precedence." on the `ExcludedMetrics` property should be sufficient.

bart-vmware:
> looks OK to me, wouldn't be a bad idea to address the sonar issue now but isn't _required_ in this branch

Just because no warning is produced? In that case, I'm afraid I have to disagree. In that same reasoning, it would be okay to introduce race conditions. Writing proper English is something we should do regardless of tooling.

I don't expect this PR to fix all existing docs, but we should get it right from the start for new ones. There's no point in postponing. The effort is minimal and it needs to be done anyway. Postponing only makes it harder to merge to `main` at a later time.

src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
bart-vmware: From the link I posted:
> A violation of this rule occurs when part of the documentation does not end with a period. 

> How to fix violations
> To fix a violation of this rule, ensure that all sections of the documentation end with a period.

src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
```diff
@@ -19,6 +19,15 @@ public interface IMetricsObserverOptions
     /// </summary>
     string EgressIgnorePattern { get; }
 
+    /// <summary>
+    /// Gets Allow list of metrics that should be captured
+    /// </summary>
+    /// <remarks>
+    ///     Currently only applies to System.Runtime metrics captured by <see cref="EventCounterListener"/><para />
+    ///     See this list for values to choose from: <see href="https://docs.microsoft.com/dotnet/core/diagnostics/available-counters#systemruntime-counters" />
```
TimHess, line 27:
```suggestion
    /// Gets Allow list of metrics that should be captured.
    /// </summary>
    /// <remarks>
    ///     Currently only applies to System.Runtime metrics captured by <see cref="EventCounterListener"/><para />
    ///     See this list for values to choose from: <see href="https://docs.microsoft.com/dotnet/core/diagnostics/available-counters#systemruntime-counters" />.
```

TimHess:
> > looks OK to me, wouldn't be a bad idea to address the sonar issue now but isn't _required_ in this branch
> 
> Just because no warning is produced? In that case, I'm afraid I have to disagree. In that same reasoning, it would be okay to introduce race conditions. Writing proper English is something we should do regardless of tooling.
> 
> I don't expect this PR to fix all existing docs, but we should get it right from the start for new ones. There's no point in postponing. The effort is minimal and it needs to be done anyway. Postponing only makes it harder to merge to `main` at a later time.

I get your point, and in that case my response would be different, but in this particular case we're talking about two periods that probably won't even be noticed by anybody. In the future, for trivial changes like this I think it would be nice to include the suggested change up front (when possible) for a quicker resolution

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
TimHess: Seems fair to put it both places. This new property definitely does need to be in the docs regardless of how it works

src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
hananiel: None of the other elements have this - wouldn't be consistent. We can address resolving this in a separate PR.

src/Management/src/EndpointBase/Metrics/Observer/EventCounterListener.cs
hananiel: self documenting code

src/Management/src/EndpointBase/Metrics/IMetricsObserverOptions.cs
hananiel: I will fix this in main branch.

PR #1060, opened 2022-11-17
Change default to 60 second timeout for config server client

### LINKED ISSUE #1058: Config Server Provider Config Setting value for Timeout
We currently have a lot of apps running on Tanzu Application Service using the Steeltoe Config Server Provider.  We've noticed at times some apps are failing to retrieve config values from a bound Spring Cloud Config Server service instance.  We finally determined that the bound SC config server service instance was taking too long to respond to the .NET app and the Steeltoe Config Server Provider was dropping the connection.  We determined the Config Server Provider default **timeout** is set to [6000 ms](https://docs.steeltoe.io/api/v3/configuration/config-server-provider.html#configure-settings), which seems low and kind of an arbitrary number.  Should this have been 60 seconds (60000 ms), which aligns with the default timeout for the Java Spring Cloud Client?   To address the issue, all of our application owners had to override the default in their **appsettings.json** file to something like:

`"spring": {
  "application": {
    "name": "my-app"
  },
  "cloud": {
    "config": {
      "timeout": 30000
    }
  }
},`

Can the default **timeout** value be set to 60000 to align with the Java Spring Cloud Client default timeout?

None

## Additional context
None

dtillman:
## Description

The default timeout for config server client was 6 seconds.  This change updates the timeout to 60 seconds to be in line with the Java Spring implementation.

Fixes #1058


bart-vmware (approved):
LGTM

PR #1064, opened 2022-11-29
Logging fixes and enhancements

bart-vmware:
## Description

Logging
- Reuse existing console formatters.
- Fixed: scopes do not work when enabled through configuration.
- Fixed: turn off colors when on CloudFoundry.

To clarify: before this PR, `DynamicConsoleLoggerProvider` would instantiate `ConsoleLoggerProvider`, passing in options but not formatters. This resulted in the default formatters still being instantiated under the covers, but ignoring configuration from appsettings.json. The effect is that it wasn't possible to turn on scopes or turn off console colors. The fix is to obtain a `ConsoleLoggerProvider` from the D/I container instead, which calls the proper constructor that includes the `formatters` parameter.

The logic to control and test colors used to be wrapped in a `#if ... #endif` because `ConsoleLoggerOptions.DisableColors` was deprecated after .NET Core 3.1 in favor of `SimpleConsoleFormatterOptions.ColorBehavior`. When Steeltoe dropped support for .NET Core 3.1, the related code was deleted. This PR restores that functionality.

## Quality checklist


src/Logging/src/Abstractions/DynamicLoggerProviderBase.cs
```diff
@@ -333,4 +333,12 @@ private IEnumerable<string> GetKeyPrefixes(string name)
 
         return null;
     }
+
+    public void SetScopeProvider(IExternalScopeProvider scopeProvider)
```
TimHess, line 337: How or when is this used?

src/Management/test/Endpoint.Test/Loggers/TestLogProvider.cs
```diff
@@ -35,4 +35,8 @@ public void SetLogLevel(string category, LogLevel? level)
         Category = category;
         Level = level ?? LogLevel.None;
     }
+
+    public void SetScopeProvider(IExternalScopeProvider scopeProvider)
```
TimHess, line 39: Is this supposed to be empty?

src/Logging/src/Abstractions/DynamicLoggerProviderBase.cs
bart-vmware: For scopes to work, `ConsoleLogger` needs a scope provider [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleLogger.cs#L79). It is then subsequently passed on to the formatter [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/ConsoleLogger.cs#L56), which renders the scope values. `LoggerFactory` creates the scope provider and exposes it to logger providers  [here](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging/src/LoggerFactory.cs#L197-L202).

TimHess: Oh! I missed the connection with `ISupportExternalScope`

src/Logging/src/Abstractions/DynamicLoggerProviderBase.cs
```diff
@@ -333,4 +333,12 @@ private IEnumerable<string> GetKeyPrefixes(string name)
 
         return null;
     }
+
```
hananiel, line 336:
```suggestion
/// <inheritdoc cref="ISupportExternalScope"/>
```

hananiel: Since it's public it helps to have some docs even though its only ever called by framework.

src/Logging/src/Abstractions/MessageProcessingLogger.cs
```diff
@@ -46,31 +46,32 @@ public bool IsEnabled(LogLevel logLevel)
...
+        if (processors != null)
         {
-            foreach (IDynamicMessageProcessor processor in MessageProcessors)
+            foreach (IDynamicMessageProcessor processor in processors)
             {
                 message = processor.Process(message);
             }
         }
 
-        if (!string.IsNullOrEmpty(message) || exception != null)
```
hananiel, line 66: This  null/empty message or exception was removed- was this intentional?

src/Logging/src/DynamicLogger/ConsoleLoggerOptionsSetup.cs
```diff
@@ -1,17 +0,0 @@
...
-// The .NET Foundation licenses this file to you under the Apache 2.0 License.
-// See the LICENSE file in the project root for more information.
-
-using Microsoft.Extensions.Logging.Configuration;
-using Microsoft.Extensions.Logging.Console;
-using Microsoft.Extensions.Options;
-
-namespace Steeltoe.Logging.DynamicLogger;
-
-internal sealed class ConsoleLoggerOptionsSetup : ConfigureFromConfigurationOptions<ConsoleLoggerOptions>
```
hananiel, line 11: I don't know if this impacts how we are using this but not all of `ConsoleLoggerOptions`[link](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.console.consoleloggeroptions?view=dotnet-plat-ext-7.0) has been deprecated - just most of it. Do the remaining two have a bearing on how our dynamic logger is  configured?

src/Logging/src/DynamicLogger/DynamicLoggingBuilder.cs
```diff
@@ -24,34 +24,69 @@ public static ILoggingBuilder AddDynamicConsole(this ILoggingBuilder builder)
...
+        // Remove the original ConsoleLoggerProvider registration as ILoggerProvider to prevent duplicate logging.
+        ServiceDescriptor descriptor = services.FirstOrDefault(descriptor => descriptor.ImplementationType == typeof(ConsoleLoggerProvider));
+
+        if (descriptor != null)
+        {
+            services.Remove(descriptor);
+        }
+
+        // Yet we need an instance from the container to construct DynamicConsoleLoggerProvider, so register without interface.
+        services.AddSingleton<ConsoleLoggerProvider>();
```
hananiel, line 66: 👍

src/Logging/src/DynamicLogger/DynamicLoggingBuilder.cs
```diff
@@ -24,34 +24,69 @@ public static ILoggingBuilder AddDynamicConsole(this ILoggingBuilder builder)
...
             builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<ILoggerProvider, DynamicConsoleLoggerProvider>());
-            builder.Services.TryAddEnumerable(ServiceDescriptor.Singleton<IConfigureOptions<ConsoleLoggerOptions>, ConsoleLoggerOptionsSetup>());
-
-            builder.Services.TryAddEnumerable(ServiceDescriptor
-                .Singleton<IOptionsChangeTokenSource<ConsoleLoggerOptions>,
-                    LoggerProviderOptionsChangeTokenSource<ConsoleLoggerOptions, ConsoleLoggerProvider>>());
-
             builder.Services.AddSingleton(p => p.GetServices<ILoggerProvider>().OfType<IDynamicLoggerProvider>().SingleOrDefault());
+
+            DisableConsoleColorsOnCloudFoundry(builder);
```
hananiel, line 36: Are we sure it is actually still needed (for example see: https://github.com/dotnet/runtime/issues/37421) b. Is this needed in K8s/TAP and if so it might be named differently.

hananiel (changes requested):
Do  we need some tests to actually verify scoped logging is working? For example all our `Logging:IncludeScopes` is false and I believe that is no longer the correct path for that configuration.

src/Logging/src/DynamicLogger/DynamicLoggingBuilder.cs
TimHess: I think @dtillman mentioned on a call today that it is still needed and is also needed on TAP, so we should probably rename this and switch to `Platform.IsCloudHosted`

src/Logging/src/Abstractions/MessageProcessingLogger.cs
bart-vmware: Yes. The logic has moved into the formatters, for example: https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.Logging.Console/src/SimpleConsoleFormatter.cs#L50-L53

src/Logging/src/DynamicLogger/ConsoleLoggerOptionsSetup.cs
bart-vmware: [MaxQueueLength](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.console.consoleloggeroptions.maxqueuelength?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-console-consoleloggeroptions-maxqueuelength) and [QueueFullMode](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.console.consoleloggeroptions.queuefullmode?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-console-consoleloggeroptions-queuefullmode) were added in .NET 7, [LogToStandardErrorThreshold](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.console.consoleloggeroptions.logtostandarderrorthreshold?view=dotnet-plat-ext-7.0#microsoft-extensions-logging-console-consoleloggeroptions-logtostandarderrorthreshold) was introduced in .NET Core 3.

These settings are not relevant for dynamic logging, however they are consumed by its wrapped logger. In that case, I don't think we need to do anything to have the built-in console logger consume these settings.

src/Logging/src/Abstractions/DynamicLoggerProviderBase.cs
bart-vmware: Agreed. Turns out just adding it without `cref` works as well:

image

src/Common/test/Common.TestResources/TestHelpers.cs
```diff
@@ -86,7 +87,7 @@ public static ILoggerFactory GetLoggerFactory()
     {
         IServiceCollection serviceCollection = new ServiceCollection();
         serviceCollection.AddLogging(builder => builder.SetMinimumLevel(LogLevel.Trace));
-        serviceCollection.AddLogging(builder => builder.AddConsole());
```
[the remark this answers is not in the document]
bart-vmware: Thanks, I've removed the helper.

src/Logging/src/DynamicLogger/ConsoleLoggerOptionsSetup.cs
bart-vmware: I've manually tested that setting `Logging:Console:LogToStandardErrorThreshold = "Information"` in appsettings.json works when dynamic logging is used. Similarly to scopes, this is pretty hard to unit-test.

bart-vmware: Added a test that uses the obsolete boolean `DisableColors`, so this should be covered now.

bart-vmware:
> Do we need some tests to actually verify scoped logging is working? For example all our Logging:IncludeScopes is false and I believe that is no longer the correct path for that configuration.

You're right, `Logging:IncludeScopes` does not work. Although obsolete, `Logging:Console:IncludeScopes` still works.

The modern way is to use `Logging:Console:FormatterOptions:IncludeScopes`, BUT that only works when _also_ specifying a formatter explicitly by either:
- adding `Logging:Console:FormatterName="Simple|Systemd|Json"` in configuration
- calling `services.AddLogging(options => options.Add[Simple|Systemd|Json]Console(formatterOptions => formatterOptions.IncludeScopes = true))` instead of the default `.AddConsole()`

I've added configuration-based tests for both, and updated all other tests to use the obsolete `Logging:Console:IncludeScopes` instead of `Logging:IncludeScopes`. That way, those tests keep working whether any of the three formatters is specified or none at all.

PR #1065, opened 2022-11-30
Management updates

### LINKED ISSUE #369: Actuator Endpoints should listen on a different port than application port

hananiel:
## Description

Expose management on additional port 
Fixes #369

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
+++ b/src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
@@ -4,7 +4,9 @@
 
 using Microsoft.AspNetCore.Builder;
 using Microsoft.AspNetCore.Hosting;
+using Microsoft.Extensions.Configuration;
 using Steeltoe.Common.HealthChecks;
+using Steeltoe.Common.Hosting;
 using Steeltoe.Logging.DynamicLogger;
 using Steeltoe.Management.Endpoint.CloudFoundry;
 using Steeltoe.Management.Endpoint.DbMigrations;
@@ -25,6 +27,9 @@ namespace Steeltoe.Management.Endpoint;
 
 public static class ManagementWebHostBuilderExtensions
 {
+    public const string ManagementPortKey = "management:endpoints:port";
+    public const string ManagementSSLKey = "management:endpoints:sslenabled";
+
     /// <summary>
     /// Adds the Database Migrations actuator to the application.
     /// </summary>
@@ -33,7 +38,7 @@ public static class ManagementWebHostBuilderExtensions
     /// </param>
     public static IWebHostBuilder AddDbMigrationsActuator(this IWebHostBuilder hostBuilder)
     {
-        return hostBuilder.ConfigureServices((context, collection) =>
+        return hostBuilder.AddManagementPort().ConfigureServices((context, collection) =>
         {
             collection.AddDbMigrationsActuator(context.Configuration);
             collection.ActivateActuatorEndpoints();
@@ -48,7 +53,7 @@ public static class ManagementWebHostBuilderExtensions
     /// </param>
     public static IWebHostBuilder AddEnvActuator(this IWebHostBuilder hostBuilder)
     {
-        return hostBuilder.ConfigureServices((context, collection) =>
+        return hostBuilder.AddManagementPort().ConfigureServices((context, collection) =>
         {
             collection.AddEnvActuator(context.Configuration);
             collection.ActivateActuatorEndpoints();
@@ -63,7 +68,7 @@ public static class ManagementWebHostBuilderExtensions
     /// </param>
     public static IWebHostBuilder AddHealthActuator(this IWebHostBuilder hostBuilder)
     {
-        return hostBuilder.ConfigureServices((context, collection) =>
+        return hostBuilder.AddManagementPort().ConfigureServices((context, collection) =>
         {
             collection.AddHealthActuator(context.Configuration);
             collection.ActivateActuatorEndpoints();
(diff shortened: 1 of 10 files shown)


src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
```diff
@@ -313,6 +318,66 @@ public static IWebHostBuilder AddCloudFoundryActuator(this IWebHostBuilder hostB
...
+    }
+
+    internal static void GetManagementUrl(this IWebHostBuilder webHostBuilder, out int? httpPort, out int? httpsPort)
+    {
+        string portSetting = webHostBuilder.GetSetting(ManagementPortKey);
+        string sslSetting = webHostBuilder.GetSetting(ManagementSSLKey);
+
+        httpPort = httpsPort = null;
+
+        if (string.IsNullOrEmpty(portSetting))
```
TimHess, line 331: From reviewing this PR, I don't see a test or known scenario when `GetConfiguration()` should be used. I recall you mentioned there are scenarios where `GetSetting` on ln 326-327 doesn't return settings that have been specified, but I don't recall when that was - should we have a test or at least some commentary to indicate when that is expected to happen?

dtillman (approved):

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
hananiel: So using the generic host (as all our old examples do) will need to build from appsettings because of the time at which this setting is needed. 
Although appsettings.json is always present in WebHost or WebApplicationBuilder usage, it is not available in a testing scenario and needs to be handled for all existing unit tests that use AddXActuator code. 
So about the unit test - its more of an integration test to test that there is an appsettings.json and it is formatted correctly and we would be basically testing framework code thats doing this:
`new ConfigurationBuilder().AddJsonFile("appsettings.json").Build();`  

I am open to suggestions to do this better though.

src/Management/src/Endpoint/ManagementPort/ManagementPortMiddleware.cs
```diff
@@ -0,0 +1,53 @@
...
+        {
+            await _next(context);
+        }
+    }
+
+    private Task ReturnErrorAsync(HttpContext context, string managementPort)
+    {
+        string errorMessage = $"Access denied to {context.Request.Path} on port {context.Request.Host.Port} since Management Port is set to {managementPort} ";
+        _logger?.LogError("ManagementMiddleWare Error: {0}", errorMessage);
+        context.Response.Headers.Add("Content-Type", "application/json;charset=UTF-8");
```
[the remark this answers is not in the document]
hananiel: Actually the default behaviour in every other case is to just send 404 with no response header and no body. I thought this would be helpful and now that you mention it, I can't decode whether to change the response to json or change header to plaintext or remove both entirely. Open to suggestions. FWIW spring boot rest app is sending an error json but that behaviour is quite different from dotnet rest api app.

dtillman (approved):

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
TimHess: Maybe adjusting the name `GetConfiguration` to something like `GetConfigurationFallback` or somehow indicate with the method name when it's expected to happen? If I look at this again in 6 months, I don't think I'll remember this detail from reading the code. 

If we absolutely must have this method, maybe it should match the default configuration expectations for WebHost? Like add Env vars and `appsettings.{Environment}.json` too?

TimHess (changes requested):

src/Management/src/Endpoint/ManagementPort/ErrorResponse.cs
```diff
@@ -0,0 +1,20 @@
+using System.Text.Json.Serialization;
```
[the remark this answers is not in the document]
hananiel: Done ...

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
hananiel: Done ...

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
```diff
@@ -372,12 +371,8 @@ private static IConfiguration GetConfigurationFallback()
 
         try
         {
-            var environment = Environment.GetEnvironmentVariable("ASPNETCORE_ENVIRONMENT");
-            config = new ConfigurationBuilder()
```
hananiel: The left side seems better to me dont you think? Can we fix it to allow multi line and not long streams of code?

TimHess: I would guess this was an over-correction for the blank line on 379. We have plenty of examples of multi-line statements like this (just without extra lines in between)

TimHess (approved):

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
bart-vmware: This is a known limitation. The Resharper formatter cannot distinguish between Fluent API syntax and regular chained method calls, such as: `GetItems().Any()`. As explained [here](https://github.com/bkoelman/CSharpGuidelinesAnalyzer/issues/129#issuecomment-1111028535), this is a tough problem to tackle for any code analyzer.

However, it can be overridden on a case-by-case basis as below:

```c#
// @formatter:wrap_chained_method_calls chop_always
// @formatter:keep_existing_linebreaks true

            config = new ConfigurationBuilder()
                .AddJsonFile("appsettings.json")
                .AddJsonFile($"appsettings.{environment}.json", optional: true)
                .Build();

// @formatter:keep_existing_linebreaks restore
// @formatter:wrap_chained_method_calls restore
```

Issue #1076, opened 2022-12-14
Connectors fail locally when connection string does not end with a semicolon


bart-vmware:
See the PostgreSQL sample at https://github.com/SteeltoeOSS/Samples/blob/3ffaa73a8ab66d1c1adc9a5cf1d3abedce3a7949/Connectors/src/PostgreEFCore/appsettings.Development.json#L13, which is unable to connect locally using [docker](https://github.com/SteeltoeOSS/Samples/blob/main/CommonTasks.md#postgresql):

```
PostgresException: 28P01: password authentication failed for user "steeltoe"
```

Adding a semicolon to the end of the connection string fixes the problem.

Traced this down to https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Connectors/src/Connector/PostgreSql/PostgreSqlProviderConnectorOptions.cs#L62: `AddKeyValue` writes a fragment like:
```
MyKey=MyValue;
```

It should write the semicolon at the start, not at the end.

PR #1084, opened 2022-12-22
catch missing delimiters when building connection strings

TimHess:
## Description

Handle cases where a user-provided connection string does not end with the delimiter and other settings are appended

Addresses #1076 in release/3.2 - A similar fix needs to happen for 4.0 but for that release we can choose to _only_ prepend the delimiter instead of appending

## Quality checklist


hananiel (approved):
LGTM;

PR #1101, opened 2023-02-08
Remove direct dependency on OTEL for Management Endpoints

### LINKED ISSUE #1105: Remove the dependency on Otel packages in Steeltoe management
Currently Steeltoe uses the [Dotnet Metrics API ](https://learn.microsoft.com/en-us/dotnet/core/diagnostics/compare-metric-apis#systemdiagnosticsmetrics) to instrument internal code as well as to export metrics via custom formats. The Dotnet API has minimal functionality to read metrics but need further aggregation. Currently we use the [Opentelemetry-Dotnet](https://github.com/open-telemetry/opentelemetry-dotnet) project to aggregate and export metrics via:
1. Metrics Actuator
2. Prometheus Actuator
3. Wavefront Trace and Metrics exporters. 

However this brings a serious usability issue for customers instrumenting their own code with Opentelemetry. We are forced to ask customers to use a Steeltoe overload instead of the provided Opentelemetry extension methods. 

In addition, current support for prometheus used a copy of equivalent code from the otel projects because the components that were required did not have the correct interface to be leveraged with the existing code. 

We can avoid this problem altogether by implementing our own aggregation mechanism for the metrics actuator. Doing so will allow us to give a convenient mechanism to `AddAllActuators` including the metrics actuators without any reference to the Opentelemetry project. A nice benefit of Otel using the dotnet instruments under the hood means that any client instrumentation will also be available via the metrics actuator should they wish to configure it so. 

The prometheus actuator can be implemented as a separate project implemented as a simple wrapper around the Open Telemetry Dotnet implementation while continuing the support the actuator options from Steeltoe.  This helps us support dotnet apps in TAS requiring custom metrics.

hananiel:
## Description

Removes the dependency on Opentelemetry Metrics. 

Fixes #1105

## Quality checklist


src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
```diff
@@ -310,18 +310,8 @@ public static IWebHostBuilder AddCloudFoundryActuator(this IWebHostBuilder hostB
...
-    /// <param name="hostBuilder">
-    /// Your HostBuilder.
-    /// </param>
-    public static IWebHostBuilder AddWavefrontMetrics(this IWebHostBuilder hostBuilder)
-    {
-        return hostBuilder.ConfigureServices((context, collection) => collection.AddWavefrontMetrics());
-    }
-
     internal static void GetManagementUrl(this IWebHostBuilder webHostBuilder, out int? httpPort, out int? httpsPort)
+#pragma warning restore SA1612 // Element parameter documentation should match element parameters
```
TimHess, line 314:
```suggestion
```

hananiel: Thanks, I addressed this along with some other changes.

dtillman (approved):
Looks good H... I see nothing major at this point in the refactoring... looking forward to the next round of changes/cleanup

PR #1114, opened 2023-03-14
Refactor Management to use IOptions pattern. 

### LINKED ISSUE #1106: Update Management Projects to use IOptions Pattern
Currently the implementation of options in steeltoe predates the IOptions Pattern and needs to be updated.

1. Use Strongly typed IOptions Pattern
2. Review the configuration objects in use
3. Resolve #503 and verify it is removed everywhere
4. IConfiguration params in extensions should be removed unless there is a good reason
5. AllowedVerbs and Exact match should not be in configuration
6. Review management tests #1028 
7. Address concerns in #504 and #545

hananiel:
## Description
Fixes #1106 

The main goal of this PR is to update the Options that endpoints use to a consistent pattern with IOptions. While doing this I have addressed a number of items mentioned in #1106. Here are few notable changes:
1. Remove the IConfiguration Parameter to extension methods - where applicable,  this is provided via the DI container.
1. This also removes the need to call `BuildServiceProvider` multiple times.
1. Use of IOptionsMonitor:   Since it gives us the maximum flexibility with being able to refresh options at runtime, I have updated them to use IOptionsMonitor even through we are are not using its capabilities fully. 
1. Deleting Abstract Endpoint - it was just reflecting endpointoptions and had helper properties for ID and Path that were part of endpointOptions. To extend functionality any endpoint can be derived from or implement a specific interface for ex. IHealthEndpoint and add it to the DI container instead of the Steeltoe provided implementation.  
1. EndpointOptions class was previously Abstract but now is concrete class that holds defaults and is renamed to EndpointOptionsBase.  IEndpointOptions no longer has IsAccessAllowed because that is only used by the Cloudfoundry endpoint.
1. IManagementOptions used to be an interface implementing different context paths ex /Actuator /Cloudfoundry and was basically a way to inform middleware of the current context and some decisions to be made such as exposure etc. There was no difference in the properties or field and was holding some glue logic. This is now implemented as named options one for each logical context (ex. Cloudfoundry, Actuator, anything new ...).
1. EndpointMiddleware are now [factory-enabled](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/middleware/extensibility?view=aspnetcore-7.0) and implement the [IMiddleware](https://learn.microsoft.com/en-us/dotnet/api/microsoft.aspnetcore.http.imiddleware?view=aspnetcore-7.0) abstraction.
1. We are still using [Endpoint-enabled ](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/routing?view=aspnetcore-7.0) middleware and still use a single startup filter to configure the pipeline order of necessary middleware - for example, managementPort or CloudfoundrySecurity, however, the mapping is now done by an [ActuatorEndpointMapper](https://github.com/SteeltoeOSS/Steeltoe/blob/ded7661c4c1221eaa67c070a71ce8a282bff1c30/src/Management/src/Endpoint/ActuatorEndpointMapper.cs#L24). This works because all the middlewares that need to routed are registered as `IEndpointMiddleware` which allows them to be injected to the mapper. Since each endpoint middleware, knows its endpoint and the options, it is easy keep this logic simple. 

## Quality checklist


dtillman (approved):
LGTM

PR #1118, opened 2023-04-04
Handle ILogger dependency consistently

hananiel:
## Description
Handle ILogger dependency consistently
Pass NullLogger<T> in tests

## Quality checklist


src/Management/src/Tracing/TracingBaseServiceCollectionExtensions.cs
```diff
@@ -89,7 +89,7 @@ public static IServiceCollection AddDistributedTracing(this IServiceCollection s
                 var traceOpts = serviceProvider.GetRequiredService<ITracingOptions>();
                 ILogger logger = serviceProvider.GetService<ILoggerFactory>()?.CreateLogger("Steeltoe.Management.Tracing.Setup");
 
-                logger?.LogTrace("Found Zipkin exporter: {exportToZipkin}. Found Jaeger exporter: {exportToJaeger}. Found OTLP exporter: {exportToOtlp}.",
+                logger.LogTrace("Found Zipkin exporter: {exportToZipkin}. Found Jaeger exporter: {exportToJaeger}. Found OTLP exporter: {exportToOtlp}.",
```
bart-vmware, line 92: I suppose the call directly above this line should be `.GetRequiredService()`, to avoid a `NullReferenceException`.

src/Management/src/Wavefront/Exporters/WavefrontMetricsExporter.cs
```diff
@@ -106,18 +108,18 @@ public override ExportResult Export(in Batch<Metric> batch)
             }
             catch (Exception ex)
             {
-                _logger?.LogError(ex, "Error sending metrics to wavefront: {message}", ex.Message);
+                _logger.LogError(ex, "Error sending metrics to wavefront: {message}", ex.Message);
```
bart-vmware, line 111: Null check is missing on the `logger` constructor parameter

src/Management/src/Wavefront/Exporters/WavefrontTraceExporter.cs
```diff
@@ -74,11 +74,11 @@ public override ExportResult Export(in Batch<Activity> batch)
             }
             catch (Exception ex)
             {
-                _logger?.LogError(ex, "Error sending metrics to wavefront: {message}", ex.Message);
+                _logger.LogError(ex, "Error sending metrics to wavefront: {message}", ex.Message);
```
bart-vmware, line 77: Null check is missing on the `logger` constructor parameter

src/Management/test/Endpoint.Test/CloudFoundry/SecurityBaseTest.cs
```diff
@@ -49,7 +50,7 @@ private static SecurityBase GetSecurityBase(out CloudFoundryEndpointOptions clou
     {
         cloudOpts = GetOptionsFromSettings<CloudFoundryEndpointOptions>();
         managementOptions = GetOptionsMonitorFromSettings<ManagementEndpointOptions>().Get(CFContext.Name);
-        var securityBase = new SecurityBase(cloudOpts, managementOptions);
+        var securityBase = new SecurityBase(cloudOpts, managementOptions, NullLogger.Instance);
```
bart-vmware, line 53: Unrelated to this PR, and perhaps a silly question: Looking at the name, `SecurityBase` seems to be intended as a base class. However, it has no virtual and no protected members. And there are no types in the solution that inherit from it. Which makes me wonder: what is the intended purpose of this class?

src/Management/test/Endpoint.Test/Health/HealthEndpointCoreTest.cs
```diff
@@ -28,11 +29,20 @@ public void Constructor_ThrowsOnNulls()
...
-        var optionsEx = Assert.Throws<ArgumentNullException>(() => new HealthEndpointCore(null, _aggregator, contributors, ServiceOptions(), _provider));
-        var aggregatorEx = Assert.Throws<ArgumentNullException>(() => new HealthEndpointCore(_options, null, contributors, ServiceOptions(), _provider));
-        var contribEx = Assert.Throws<ArgumentNullException>(() => new HealthEndpointCore(_options, _aggregator, null, ServiceOptions(), _provider));
-        var svcOptsEx = Assert.Throws<ArgumentNullException>(() => new HealthEndpointCore(_options, _aggregator, contributors, null, _provider));
-        var providerEx = Assert.Throws<ArgumentNullException>(() => new HealthEndpointCore(_options, _aggregator, contributors, ServiceOptions(), null));
+        var optionsEx = Assert.Throws<ArgumentNullException>(() =>
+            new HealthEndpointCore(null, _aggregator, contributors, ServiceOptions(), _provider, NullLogger<HealthEndpointCore>.Instance));
+
+        var aggregatorEx = Assert.Throws<ArgumentNullException>(() =>
+            new HealthEndpointCore(_options, null, contributors, ServiceOptions(), _provider, NullLogger<HealthEndpointCore>.Instance));
```
bart-vmware, line 36: I wonder if the `ILogger<HealthEndpointCore> logger` parameter on `HealthEndpointCore` should be `ILogger` instead. Is this type non-sealed with protected and virtual members solely to make it possible to inherit from in our tests, or is the type intended to be inherited from by Steeltoe users?

src/Management/test/Endpoint.Test/Metrics/MetricsEndpointTest.cs
```diff
@@ -28,9 +28,9 @@ public MetricsEndpointTest(ITestOutputHelper output)
     [Fact]
     public void Constructor_ThrowsIfNulls()
     {
-        Assert.Throws<ArgumentNullException>(() => new MetricsEndpoint(null, null));
+        Assert.Throws<ArgumentNullException>(() => new MetricsEndpoint(null, null, null));
         IOptionsMonitor<MetricsEndpointOptions> options = GetOptionsMonitorFromSettings<MetricsEndpointOptions>();
-        Assert.Throws<ArgumentNullException>(() => new MetricsEndpoint(options, null));
+        Assert.Throws<ArgumentNullException>(() => new MetricsEndpoint(options, null, null));
```
bart-vmware, line 33: Personally, I don't see the point of having these tests for constructor-injected types. However, this test now looks incomplete, because it does not handle the case where all parameters are non-null, except `logger`. So I propose to either add that case, or remove the test altogether.

Maybe this is something to talk about with the rest of the team?

src/Management/test/Endpoint.Test/Trace/TraceEndpointTest.cs
```diff
@@ -24,7 +24,7 @@ public TraceEndpointTest(ITestOutputHelper output)
     public void Constructor_ThrowsIfNullRepo()
     {
         IOptionsMonitor<TraceEndpointOptions> opts = GetOptionsMonitorFromSettings<TraceEndpointOptions>();
-        Assert.Throws<ArgumentNullException>(() => new TraceEndpoint(opts, null));
+        Assert.Throws<ArgumentNullException>(() => new TraceEndpoint(opts, null, null));
```
bart-vmware, line 27: I have the same concern as raised earlier. I think we should either complete this test or remove it.

bart-vmware (changes requested):
Looks good overall, but it seems to me that you've missed a few cases.

src/Management/test/Endpoint.Test/CloudFoundry/SecurityBaseTest.cs
hananiel: Good catch, it used to be inheritance but got changed and it doesn't need to be public.

src/Management/test/Endpoint.Test/Health/HealthEndpointCoreTest.cs
hananiel: It is meant for inheritance by Steeltoe users and they will be forced to use ILogger<HealthEndpointCore> as a parameter to the base class, which imo should be ok. see TestHealthEndpoint in the tests ...

src/Management/test/Endpoint.Test/Metrics/MetricsEndpointTest.cs
hananiel: I agree for now I'll add the test to make it consistent.

src/Management/src/Endpoint/Utils.cs
```diff
@@ -24,7 +24,7 @@ public static class Utils
     /// <returns>
     /// A file stream with the file's compressed contents.
     /// </returns>
-    public static Stream CompressFile(string filename, string gzFilename, ILogger logger = null)
+    public static Stream CompressFile(string filename, string gzFilename, ILogger logger)
```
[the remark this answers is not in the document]
bart-vmware: The null check should be outside the try/catch block, to prevent the catch handler from throwing NullReferenceException while trying to log the original error.

src/Management/src/Endpoint/Utils.cs
```diff
@@ -64,7 +64,7 @@ public static Stream CompressFile(string filename, string gzFilename, ILogger lo
     /// <returns>
     /// A file stream with the file's compressed contents.
     /// </returns>
-    public static async Task<Stream> CompressFileAsync(string filename, string gzFilename, ILogger logger = null)
+    public static async Task<Stream> CompressFileAsync(string filename, string gzFilename, ILogger logger)
```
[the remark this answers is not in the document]
bart-vmware: The null check should be outside the try/catch block, to prevent the catch handler from throwing NullReferenceException while trying to log the original error.

src/Management/test/Endpoint.Test/Health/HealthEndpointCoreTest.cs
bart-vmware: Ok, let me try to make my point by inheriting from the base type:

```c#
public class CustomHealthEndpointCore : HealthEndpointCore
{
    private readonly ILogger<HealthEndpointCore> _logger;

    public CustomHealthEndpointCore(/* other ctor parameters left out for brevity */
        ILogger<HealthEndpointCore> logger)
        : base(options, aggregator, contributors, serviceOptions, provider, logger)
    {
        _logger = logger;
    }

    public override HealthEndpointResponse Invoke(ISecurityContext securityContext)
    {
        _logger.LogDebug("Invoke derived.");
        return base.Invoke(securityContext);
    }
}
```

And for demonstration purposes, let's insert a log line to the base method:
```c#
    public virtual HealthEndpointResponse Invoke(ISecurityContext securityContext)
    {
        _logger.LogDebug("Invoke base.");
        return BuildHealth(securityContext);
    }
```

This would log:
```
HealthEndpointCore: Invoke derived.
HealthEndpointCore: Invoke base.
```

This doesn't look right, because the first log statement occurred in `CustomHealthEndpointCore` instead of `HealthEndpointCore`. To fix that, we'll need to add an extra constructor parameter with our own typed `ILogger`:

```c#
public class CustomHealthEndpointCore : HealthEndpointCore
{
    private readonly ILogger<CustomHealthEndpointCore> _customLogger;

    public CustomHealthEndpointCore(/* other ctor parameters left out for brevity */
        ILogger<HealthEndpointCore> logger,
        ILogger<CustomHealthEndpointCore> customLogger)
        : base(options, aggregator, contributors, serviceOptions, provider, logger)
    {
        _customLogger = customLogger;
    }

    public override HealthEndpointResponse Invoke(ISecurityContext securityContext)
    {
        _customLogger.LogDebug("Invoke derived.");
        return base.Invoke(securityContext);
    }
}
```

This would log:
```
CustomHealthEndpointCore: Invoke derived.
HealthEndpointCore: Invoke base.
```

This is correct, but it's a bit annoying that every derived type needs to add its own typed `ILogger` parameter. I suspect that's why many other Steeltoe types take an untyped `ILogger` parameter. For example, `AvailabilityHealthContributor` and `DiagnosticObserver`.

The downside of that approach is that logging statements from the base type now become logged under the name of the most-derived type. To let every type in the hierarchy log under its own name without adding many constructor parameters, each type in the hierarchy should take an `ILoggerFactory` instead. Then each type can create its own typed logger, so that it becomes easy to trace back which class a line originated from.

However, using an untyped `ILogger` in abstract types seems to be an established pattern in the Steeltoe codebase (I'm seeing hundreds of usages), so it seems to me we should stick to that.

src/Management/test/Endpoint.Test/CloudFoundry/SecurityBaseTest.cs
bart-vmware: In that case, can we make the type `internal sealed` with a public constructor? There's already an `[InternalsVisibleTo]` so tests can still access it.

What I learned the hard way, is that it would have saved me a lot of work if I'd started out with making types internal and/or sealed before looking at the Sonar violations. Many of them only apply to the public API surface, and some don't apply to sealed types. For each type, the basic question is: do we intend Steeltoe users to inherit from it? If so, it should have protected and/or virtual members. By narrowing down our public API, we're getting a lot more freedom to change things without having to wait for the next major version.

bart-vmware (changes requested):
See comments.

bart-vmware (approved):

PR #1139, opened 2023-07-10
Connectors cleanup

bart-vmware:
## Description

This PR removes the legacy Connectors code. I recommend reviewing per commit.

## Quality checklist


src/Connectors/src/Connectors/CosmosDb/CosmosDbServiceCollectionExtensions.cs
```diff
@@ -73,7 +98,11 @@ public static IServiceCollection AddCosmosDb(this IServiceCollection services, I
         string hostName = GetHostNameFromConnectionString(connectorShim.Options.ConnectionString);
         var logger = serviceProvider.GetRequiredService<ILogger<CosmosDbHealthContributor>>();
 
-        return new CosmosDbHealthContributor(cosmosClient, $"CosmosDB-{serviceBindingName}", hostName, logger);
+        return new CosmosDbHealthContributor(cosmosClient, hostName, logger)
+        {
+            ServiceName = serviceBindingName,
+            Timeout = TimeSpan.FromSeconds(5)
```
TimHess, line 104: Is there a simple way we can make this timeout configurable?

src/Connectors/src/Connectors/MongoDb/MongoDbWebApplicationBuilderExtensions.cs
```diff
@@ -2,20 +2,42 @@
...
-
 using Microsoft.AspNetCore.Builder;
 using Steeltoe.Common;
 
 namespace Steeltoe.Connectors.MongoDb;
 
 public static class MongoDbWebApplicationBuilderExtensions
 {
+    /// <summary>
+    /// Registers a <see cref="ConnectorFactory{TOptions,TConnection}" /> to connect to a MongoDB database.
```
TimHess, line 13: Would it be more user-friendly if these comments specified the implementation types they're setting up instead of showing the generics?

src/Connectors/src/Connectors/RabbitMQ/RabbitMQHealthContributor.cs
```diff
@@ -4,110 +4,93 @@
...
         try
         {
-            _connection ??= ReflectionHelpers.Invoke(RabbitMQTypeLocator.CreateConnectionMethod, _connFactory, null);
+            _connectionInterfaceShim ??= _connectionFactoryInterfaceShim.CreateConnection();
 
-            if (_connection == null)
+            if (!_connectionInterfaceShim.IsOpen)
             {
-                throw new ConnectorException("Failed to open RabbitMQ connection!");
+                throw new IOException("RabbitMQ connection is closed!");
```
TimHess, line 59: Doesn't `IOException` usually imply there's a file or disk-related issue?

src/Connectors/src/Connectors/MongoDb/MongoDbWebApplicationBuilderExtensions.cs
bart-vmware: I tried that, but the values get discarded by the compiler, so they won't show up in IntelliSense. The only thing I can think of is to spell them out literally, for example:
```
Registers a <see cref="ConnectorFactory{TOptions,TConnection}" />
with type parameters <see cref="MongoDbOptions"/> and MongoDB.Driver.MongoClient
to connect to...
```
Which I think would be helpful. What do you think?

src/Connectors/src/Connectors/RabbitMQ/RabbitMQHealthContributor.cs
bart-vmware: I searched for a while, but found no better candidate in the BCL exception types. I chose this one because Sonar doesn't like me throwing `Exception`, `ApplicationException`, or `SystemException` (too generic). However, I can suppress that rule in the code.

On a side note, I stepped away from throwing `ConnectorException` because it implies users can catch that and be fine. Which is definitely not the case in so many scenarios. We'd need to wrap all reflection calls in a try/catch, which takes away the possibility for clients to catch `TimeoutException`, but not `SqlServerException`.

Do you have a better exception type in mind?

src/Connectors/src/Connectors/CosmosDb/CosmosDbServiceCollectionExtensions.cs
bart-vmware: This is a tricky one. The CosmosDB driver provides `options.RequestTimeout`, but no timeout setting for _establishing_ the socket connection. I found it fails on my machine after 30 seconds with a `SocketException` when the host is unreachable. The CosmosDB driver just passes `CancellationToken.None` to the socket API. I can dig deeper to find where the 30 seconds comes from.

Also, `ReadAccountAsync` doesn't take a `CancellationToken`, so we're stuck with the 30 seconds _when the host is unreachable_. Otherwise, the `options.RequestTimeout` is applied, which users can already set as demonstrated [here](https://github.com/SteeltoeOSS/Samples/blob/latest/Connectors/src/CosmosDb/Program.cs#L20).

So the best we can do is wait for the returned `Task` with our own time-out. However, that doesn't terminate the task itself and is merely a hack. All the other drivers do not have this problem, which is why only the CosmosDB health contributor has this `Timeout` property.

Exposing the property may lead users to believe this is _the_ timeout being used. But that's not the case: if `options.RequestTimeout` has a smaller value, or if `CosmosDbHealthContributor.Timeout` is higher than 30 seconds, then this property has no effect.

Perhaps the timeout is a theoretical case. It is far more likely that an Azure instance is responding slowly, than that its host (something.azure.com) is unreachable.

I wouldn't mind changing the value from 5 seconds to something else (or removing it entirely), but I don't think we should expose it.

src/Connectors/src/Connectors/RabbitMQ/RabbitMQHealthContributor.cs
TimHess: I think this was the original primary use case for `ConnectorException` -- something went wrong in Connectors that isn't covered by another common exception type. If you don't want to use `ConnectorException` in any of the new code, what do you think about an `InvalidOperationException` here?

src/Connectors/src/Connectors/MongoDb/MongoDbWebApplicationBuilderExtensions.cs
TimHess: That seems like it could be useful if Steeltoe users will need to use those types. I'm not currently deep enough into what it's like using these new connectors to have a solid opinion yet, so I'll leave it to your discretion as to whether it's worth the effort to add those

src/Connectors/src/Connectors/CosmosDb/CosmosDbServiceCollectionExtensions.cs
TimHess: I think I'm OK with it as-is given this explanation. I was thinking we would call it something like `HealthContributorTimeout` if it was exposed so that it wouldn't be confusing, but it should also be fine this way for now. feel free to resolve this one if no further discussion is desired

src/Connectors/test/CloudFoundry.Test/CloudFoundryServiceInfoCreatorTest.cs
```diff
@@ -66,353 +64,4 @@ public void BuildServiceInfos_NoCloudFoundryServices_BuildsExpected()
...
-                            ""uri"": ""10.66.42.50:9001"",
-                            ""accesskey"": ""foo"",
-                            ""secretkey"": ""bar""
-                        },
-                        ""syslog_drain_url"": """",
-                        ""volume_mounts"": [],
-                        ""label"": ""user-provided"",
-                        ""name"": ""foo_minio"",
-                        ""tags"": []
-                    }]
```
TimHess, line 391: This question probably goes beyond the scope of the work you've been doing, and it might not be quite the right location to raise it, but when I asked (outside of this PR) about CUPS support, I was referring to whether we might be able to recognize and support a binding that looks like this:

```
                   ""user-provided"": [{
                        ""credentials"": {
                            ""uri"": ""amqp://si_username:si_password@localhost:5672/si_vhost""
                        },
                        ""syslog_drain_url"": """",
                        ""volume_mounts"": [],
                        ""label"": ""user-provided"",
                        ""name"": ""foo_minio"",
                        ""tags"": []
                    }]
```

TimHess:
Do you see any benefit in keeping `Steeltoe.Connectors.Abstractions` as a package? It seems we're turning virtually all of the contents into `internal` implementation details now, and potentially going even further as we get into Discovery and Security

src/Connectors/src/Connectors/CosmosDb/CosmosDbServiceCollectionExtensions.cs
bart-vmware: Ok, I'm resolving this one for now, but thanks for bringing it up. We may need to revisit this if/when adopting the community ASP.NET health checks.

src/Connectors/src/Connectors/RabbitMQ/RabbitMQHealthContributor.cs
bart-vmware: It doesn't literally say so, but based on the examples at https://learn.microsoft.com/en-us/dotnet/api/system.invalidoperationexception?view=net-7.0, `InvalidOperationException` usually indicates that _you_ (the developer) are doing something wrong. Instead of an intermittent, retriable failure, which is what we need here. While `IOException` isn't perfect, I still think it's the next best thing. It was proposed to add `ConnectionException` to the BCL, which [derives](https://github.com/dotnet/runtime/issues/39941) from `IOException`. And it got merged, but somehow it didn't survive.

src/Connectors/src/Connectors/MongoDb/MongoDbWebApplicationBuilderExtensions.cs
bart-vmware: Reopening to address this.

src/Connectors/test/CloudFoundry.Test/CloudFoundryServiceInfoCreatorTest.cs
bart-vmware: Thanks for the pointer. Let's revisit this during the design of extensibility for connectors.

bart-vmware:
> Do you see any benefit in keeping `Steeltoe.Connectors.Abstractions` as a package? It seems we're turning virtually all of the contents into `internal` implementation details now, and potentially going even further as we get into Discovery and Security

Good point. We may (or may not) need it again when adding extension points for connectors. Let's keep it for now and reconsider when we open up connectors for extensibility.

src/Connectors/src/Connectors/MongoDb/MongoDbWebApplicationBuilderExtensions.cs
bart-vmware: It now looks like this:

image

TimHess (approved):
LGTM

PR #1153, opened 2023-07-27
Add empty arrays for Headers, Params instead of missing json objects

### LINKED ISSUE #1151: Make mappings actuator output compatible with SpringBootAdmin
Currently viewing the mappings page on spring boot admin shows a blank, with a javascript error: `Cannot read property 'length' of undefined `

Spring boot admin is looking for 'Headers', 'Params' and in general any route details to have a Json array even if empty.

hananiel:
## Description

Adding additional fields in Json output to support Spring Boot Admin. The fields in question 'Headers' and 'Params' are Spring boot features on @RequestMapping annotation to match routes to controller actions. In .Net this is accomplished differently and is hard to infer as easily as in Spring. So, we will just add empty arrays to cause SpringBootAdmin not to error on this page.

Fixes #1151 

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/Mappings/AspNetCoreRouteDetails.cs
+++ b/src/Management/src/Endpoint/Mappings/AspNetCoreRouteDetails.cs
@@ -13,4 +13,8 @@ public class AspNetCoreRouteDetails : IRouteDetails
     public IList<string> Produces { get; set; }
 
     public IList<string> Consumes { get; set; }
+
+    public IList<string> Params { get; set; }
+
+    public IList<string> Headers { get; set; }
 }
--- a/src/Management/src/Endpoint/Mappings/MappingDescription.cs
+++ b/src/Management/src/Endpoint/Mappings/MappingDescription.cs
@@ -43,23 +43,17 @@ public class MappingDescription
 
     private RouteMappingDetails CreateMappingDetails(IRouteDetails routeDetails)
     {
-        return new RouteMappingDetails
+        
+        return new RouteMappingDetails()
         {
             RequestMappingConditions = new RequestMappingConditions
             {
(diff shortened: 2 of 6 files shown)


src/Management/src/Endpoint/Mappings/MappingDescription.cs
```diff
@@ -50,16 +50,22 @@ private RouteMappingDetails CreateMappingDetails(IRouteDetails routeDetails)
...
                 {
                     routeDetails.RouteTemplate
-                }
+                },
+                Params = new string[]
+                {
+                }, // does not apply for .NET
+                Headers = new string[]
+                {
+                } // Cannot infer here for .NET
```
bart-vmware, line 68: Why return a list in some cases, and an array in others? I suppose returning a list in all cases is slightly more convenient to calling code that wants to modify the collections.

src/Management/src/Endpoint/Mappings/AspNetCoreRouteDetails.cs
```diff
@@ -13,4 +13,8 @@ public class AspNetCoreRouteDetails : IRouteDetails
     public IList<string> Produces { get; set; }
 
     public IList<string> Consumes { get; set; }
+
+    public IList<string> Params { get; set; }
+
+    public IList<string> Headers { get; set; }
```
bart-vmware, line 19: Note I'm going to initialize these properties with empty collections when merging this into #1155, based on what we discussed yesterday. I already did that for all existing collection properties there, so I think it makes the most sense to handle that in my PR.

src/Management/src/Endpoint/Mappings/MappingDescription.cs
hananiel: fixed to return list all the time.

bart-vmware (approved):
LGTM

PR #1155, opened 2023-07-28
Management API review, continued

### LINKED ISSUE #905: Review API Surface area in Steeltoe.Management

### LINKED ISSUE #906: Review API Surface area in Steeltoe.Messaging

### LINKED ISSUE #1122: Review API surface area for Steeltoe.Management.*
### 
**Surface API:**

- Seal public types that don't seem (properly) designed for inheritance (please indicate where I'm wrong)
- Replace optional parameters with overloads in publicly exposed methods
- Add input validation (null checks) on the parameters of publicly exposed methods
- Change publicly exposed fields into properties, unless they are constants

**Activate the next Sonar rules:**

- [ ] - [Address SA1401: Fields should be private #928](https://github.com/SteeltoeOSS/Steeltoe/issues/928)
- [ ] - [Address S1168: Empty arrays and collections should be returned instead of null #916](https://github.com/SteeltoeOSS/Steeltoe/issues/916)
- [ ] - [Address S2360: Optional parameters should not be used #929](https://github.com/SteeltoeOSS/Steeltoe/issues/929)
- [ ] - S3900: Arguments of public methods should be validated against null
- [ ] - [Address S3956: "Generic.List" instances should not be part of public APIs #941](https://github.com/SteeltoeOSS/Steeltoe/issues/941)
- [ ] - [Address S4023: Interfaces should not be empty #947](https://github.com/SteeltoeOSS/Steeltoe/issues/947)
- [ ] - [Address S4004: Collection properties should be readonly #946](https://github.com/SteeltoeOSS/Steeltoe/issues/946)

bart-vmware:
## Description

Continuation of the work done in #1120. See the commit messages for an overview of what's changed.

API Review Checklist (copied from https://github.com/SteeltoeOSS/Steeltoe/issues/897#issuecomment-1198088843):
- [x] Review namespaces
  - [x] Consider the use of [pubternal](https://www.reddit.com/r/dotnet/comments/jwzgl9/word_of_the_day_pubternal/)
  - [x] Adjust namespaces based on directory (or adapt project structure to match namespaces)
  - [x] Don't put extensions methods in System namespaces. While this makes them easy to discover, sooner or later they are going to clash with other libraries, requiring users to write non-intuitive code to workaround the conflict
- [x] Remove obsolete types and members
- [x] Consider making public types/members internal, use `InternalsVisibleTo` to access them from tests
- [x] Seal public types, unless they are designed for inheritance
- [x] Avoid static classes, they are hard to test. Exceptions: extension method containers or classes to wrap a set of immutable constants
- [x] A class or interface should have a single purpose
  - [x] A class or interface should have a single purpose within the system it functions in. In general, a class either represents a primitive type like an email or ISBN number, an abstraction of some business concept, a plain data structure, or is responsible for orchestrating the interaction between other classes. It is never a combination of those. This rule is widely known as the [Single Responsibility Principle](https://8thlight.com/blog/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html), one of the S.O.L.I.D. principles
  - [x] Similarly, a property, method or local function should do only one thing
- [ ] Consider adding doc-comments to public types and members, but avoid [undocumentation](https://blog.codinghorror.com/avoiding-undocumentation/)
- [x] Public types should protect the consistency of their internal state: validate inputs and use guards like `AssertNotDisposed();`
- [ ] Use a method instead of a property when:
  - [ ] The work is more expensive than setting a field value
  - [ ] It represents a conversion such as the `Object.ToString` method
  - [ ] It returns a different result each time it is called, even if the arguments didn’t change. For example, the `NewGuid` method returns a different value each time it is called
  - [ ] The operation causes a side effect such as changing some internal state not directly related to the property (which violates the [Command Query Separation](http://martinfowler.com/bliki/CommandQuerySeparation.html) principle). Populating an internal cache or implementing [lazy-loading](http://www.martinfowler.com/eaaCatalog/lazyLoad.html) is a good exception
- [x] Suffix identifier names that represent a `Type`, `MethodInfo`, `PropertyInfo` etc, to distinguish them from instances. For example: rename `MethodInfo connect` to `MethodInfo connectMethod`
- [x] Avoid publicly exposed fields, unless they are `const` or `static readonly`. Convert to (auto-)properties or make them private/internal
- [x] Prefer collection interfaces such as `ICollection<>` and `IDictionary<,>` over arrays, list and dictionaries for inputs and outputs
  - [x] Be hesitant with incoming `IEnumerable<>` parameters: the implementation may change over time, causing the need for multiple enumerations, which may result in undesired side effects
  - [x] Consider to return unchangeable collections, such as `IReadOnlyCollection<>`, `IReadOnlyDictionary<,>` and `IImmutableList<T>`
- [x] Use overloads instead of optional parameters (they have several backwards-compatibility issues, as well as non-intuitive behavior when using interfaces or inheritance). Call the more overloaded method from other overloads
- [x] Properties, arguments and return values representing strings, collections or tasks should never be null. Return an empty collection, an empty string or `Task.CompletedTask`/`Task.FromResult()`
- [x] Don’t use ref or out parameters (return compound objects or tuples), unless the method implements the [TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse) pattern
- [x] Postfix asynchronous methods with Async or TaskAsync.
- [x] Consider using `ValueTask`/`ValueTask<T>` in high-performance code paths
- [x] Avoid abbreviations
  - [x] Use "management" instead of "mgmt", "options" instead of "opts", etc.
  - [x] Using "config" in the context of ConfigServer is good, but not: `IConfiguration config`
  - [x] Avoid single character variable names, such as i or q. Use index or query instead
  - [x] Use well-known acronyms and abbreviations that are widely accepted or well-known in your work domain. For instance, use acronym UI instead of UserInterface and abbreviation Id instead of Identity
- [x] Remove the "Core" suffix when referring to .NET (ASP.NET Core and Entity Framework Core are okay)

Resolves #905

## Quality checklist


src/Management/src/MetricCollectors/Aggregations/IAggregationStatistics.cs
```diff
@@ -0,0 +1,12 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
hananiel: What is the benefit of separating into different files when we are trying to match the code from original source?

hananiel (approved):
LGTM

src/Management/src/MetricCollectors/Aggregations/IAggregationStatistics.cs
bart-vmware: I changed this before we decided to match the original source. I'm going to push another commit for that, which will revert this separation.

bart-vmware: I've pushed two commits: The first is the unmodified version from .NET 7 stable. The next one contains our modifications.

hananiel (dismissed):
Looks great.

hananiel (approved):
LGTM

hananiel (approved):

PR #1165, opened 2023-08-21
Fix broken heapdump tests for OSX

hananiel:
## Description

Running the full dump in OSX appears to be broken even outside Steeltoe using dotnet dump. Running actuators in OSX server side appears unlikely, yet to support developers working on a Mac, we want to provide some functionality here. Instead of disabling the heap dump actuator entirely, we are configuring it to return a gcdump instead which appears to work fine. 

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/HeapDump/ConfigureHeapDumpEndpointOptions.cs
+++ b/src/Management/src/Endpoint/HeapDump/ConfigureHeapDumpEndpointOptions.cs
@@ -3,16 +3,33 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+using Steeltoe.Common;
 using Steeltoe.Management.Endpoint.Options;
 
 namespace Steeltoe.Management.Endpoint.HeapDump;
 
 internal sealed class ConfigureHeapDumpEndpointOptions : ConfigureEndpointOptions<HeapDumpEndpointOptions>
 {
+    private readonly ILogger<ConfigureHeapDumpEndpointOptions> _logger;
     private const string ManagementInfoPrefix = "management:endpoints:heapdump";
 
-    public ConfigureHeapDumpEndpointOptions(IConfiguration configuration)
+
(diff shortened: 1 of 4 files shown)


src/Management/test/Endpoint.Test/ActuatorRouteBuilderExtensionsTest.cs
```diff
@@ -20,18 +20,31 @@ namespace Steeltoe.Management.Endpoint.Test;
...
+    public static IEnumerable<object[]> ActuatorOptions
+
+    {
+        get
+        {
+            IHostBuilder hostBuilder = GetHostBuilder(policy => policy.RequireClaim("scope", "actuators.read"));
+            return hostBuilder.Build().Services.GetServices<EndpointOptions>().Select(options => new object[]{options});
+        }
+    }
+    [Theory]
```
hananiel, line 33: Changed this to a theory to it is easier to test, for example, if dealing with an issue with a particular actuator we no longer have to set conditional breakpoints etc to figure out what is going on.

src/Management/src/Endpoint/HeapDump/ConfigureHeapDumpEndpointOptions.cs
```diff
@@ -3,16 +3,33 @@
...
+    }
+
+    public override void Configure(HeapDumpEndpointOptions options)
+    {
+        base.Configure(options);
+
+        // Full dumps are broken on Osx, so default to gcdump
+        if (Platform.IsOSX)
+        {
+            _logger.LogInformation("Full dumps are not supported on OSX, defaulting to gcdump.");
```
TimHess, line 31:
```suggestion
        // Full dumps are broken on OSX, so only allow gcdump
        if (Platform.IsOSX && options.HeapDumpType != "gcdump")
        {
            _logger.LogInformation("Only GC dumps are supported on OSX.");
```

We're not just "defaulting" if we're forcing it to always be the same. According to the docs, we allow several types beyond full and gc, have you tested the others?

> Acceptable values are Normal, WithHeap, Triage and Full.

Apparently we need to add gcdump to the docs too

PR #1167, opened 2023-08-27
Add and use Metadata in ConsulDiscoveryOptions

### LINKED ISSUE #438: Update Consul client to use Metadata instead of Tags
Consul previously did not supports [metadata](https://www.consul.io/docs/discovery/services), but now it does. We use tags instead but can now revert to using metadata. Need to update the client to use meta instead of tags.  Here is our documentation ticket for this: https://github.com/SteeltoeOSS/Documentation/issues/19

cieciurm:
## Description

Hello!

Thank you for delivering Steeltoe 😄 .

I was browsing through the `good first issue` label and wanted to contribue a change for #438.

I'd like to open discussions on the goals and some details regarding implementation before executing it without the alignment.

Here are my thoughts:
1. Currently `ConsulDiscoveryOptions` has a property called `Tags` (list of strings)
1. Tags are expected to be in format `key=value`
1. Internally some tags are added (such as `group`, `instance zone`)
1. Currently the fallback from Tags to Metadata happens in `ConsulRegistration`:
```csharp
  Metadata = ConsulServerUtils.GetMetadata(agentServiceRegistration.Tags);
```

Here are my questions:
1. Would you like to keep the current behavior of usings tags as metadata? Should it be controlled by an option in `ConsulDiscoveryOptions`, f.e. for backwards compatibility?
1. I guess, you like to keep adding the internal metadata (such as `group`, `instance zone`)
1. What's the naming preference in `ConsulDiscoveryOptions` - should it be `Metadata` (well-known term) or `Meta` (Consul's `AgentServiceRegistration` propety is called `Meta`)

I hope, it's not too much - but I wanted to clarify before jumping to code immediately.

Once confirmed and agreed, I will finish the implementation (obviously including tests etc.).

Thanks in advance 😄 !

Fixes #438

## Quality checklist


hananiel:
Thank you for this PR. Since this PR is targeting main which will be release 4.x , we can deprecate the previous functionality with parsing tags as metadata. For functionality such as default groups, as well as property names, we should match as closely as possible the functionality described [here](https://docs.spring.io/spring-cloud-consul/docs/current/reference/html/index.html#metadata): so Metadata for us. The prefix stays same as what is currently i.e. `consul:discovery `
When back porting to 3.x, we can consider a flag such as ['tagsAsMetadata'](https://github.com/spring-cloud/spring-cloud-consul/blob/bbdff2365a491d2e371e5a1e469f7dafc30493bf/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/discovery/ConsulDiscoveryProperties.java#L62C18-L62C33)

cieciurm:
Thanks for your response, @hananiel!

I've completed my first attempt at the implementation. Scope:

- Added the `Metadata` to `ConsulDiscoveryOptions` 
- Switched the metadata generation (including the internal generated ones) from the original metadata  instead of Tags
- Also exposed regular Tags to `ConsulServiceInstance`
- I've adjusted unit tests 

Please let me know if something needs to be adjusted or changed. 

Thanks a lot in advance!

hananiel (approved):
@cieciurm This looks good. Would you like to take a crack at updating the docs as well? The v4 docs are on this branch: https://github.com/SteeltoeOSS/Documentation/tree/v4

cieciurm:
Thanks, @hananiel! You got it!

https://github.com/SteeltoeOSS/Documentation/pull/302

PR #1172, opened 2023-09-05
Add Connector support for MongoDB on Kubernetes

bart-vmware:
## Description

This PR adds support for MongoDB running on Kubernetes, adds more tests, and fixes several bugs. See the commit messages for details.

## Quality checklist


src/Configuration/src/Kubernetes.ServiceBinding/PostProcessors/RabbitMQKubernetesPostProcessor.cs
```diff
@@ -21,8 +21,9 @@ public void PostProcessConfiguration(PostProcessorConfigurationProvider provider
             // Mapping from Kubernetes secrets to driver-specific connection string parameters.
             // At the time of writing (June 2023), there's no complete official documentation for the available secrets. Some pointers:
             // - Generic secrets: https://github.com/servicebinding/spec#well-known-secret-entries
-            // - Input keys used at https://github.com/spring-cloud/spring-cloud-bindings/blob/main/src/main/java/org/springframework/cloud/bindings/boot/RabbitMqBindingsPropertiesProcessor.java
+            // - Input keys used at https://github.com/spring-cloud/spring-cloud-bindings/blob/main/spring-cloud-bindings/src/main/java/org/springframework/cloud/bindings/boot/RabbitMqBindingsPropertiesProcessor.java
 
+            mapper.SetToValue("useTls", "false");
```
TimHess, line 26: I have some concern this item will come back around in short order given that the industry is still working on TLS everywhere. I see that the [Bitnami RabbitMQ service option supports specifying a repo for a helm chart](https://docs.vmware.com/en/VMware-Tanzu-Application-Platform/1.6/tap/bitnami-services-reference-package-values.html) and the [RabbitMQ Kubernetes RabbitMQ operator supports TLS](https://www.rabbitmq.com/kubernetes/operator/using-operator.html#tls), so it seems like it should be possible already.

With that said, we could conceivably take this change now and just expect feedback on how this should actually work when a use case comes along

TimHess (approved):

src/Configuration/src/Kubernetes.ServiceBinding/PostProcessors/RabbitMQKubernetesPostProcessor.cs
bart-vmware: This isn't possible according to https://github.com/spring-cloud/spring-cloud-bindings/blob/main/spring-cloud-bindings/src/main/java/org/springframework/cloud/bindings/boot/RabbitMqBindingsPropertiesProcessor.java#L46-L51, which is our best source.

But if you can get it to work somehow, perhaps we can do better? https://github.com/servicebinding/spec#well-known-secret-entries defines `certificates` and `private-key` entries, which we'll need for mTLS.

Otherwise, I'd prefer shipping non-working code over untestable code with undefined behavior. We should get it right the first time, not paint ourselves in a corner now, requiring a breaking change later.

Just to let you know, all this line does is: "If you turned on TLS locally, turn it off now." Removing that line of code isn't going to make TLS work. For that, we need to know the name and type of the secret that controls it. But we can only guess; it's not in the standard.

bart-vmware: Ultimately, the connection string can always be adapted from user code. See the example at: https://github.com/SteeltoeOSS/Samples/blob/latest/Connectors/src/RabbitMQ/Program.cs#L14-L32.

PR #1179, opened 2023-09-08
Correct nullability in Steeltoe.Configuration

bart-vmware:
## Description

Correct the nullability annotations in Steeltoe.Configuration, based on .NET 7/8. The related annotations in the BCL didn't yet exist in .NET 6, so we didn't get warnings on them before. Getting them right makes it easier to sync the Steeltoe codebase with .NET 7/8.

## Quality checklist

### DIFF
--- a/src/Connectors/src/Connectors/ConnectionStringPostProcessor.cs
+++ b/src/Connectors/src/Connectors/ConnectionStringPostProcessor.cs
@@ -27,7 +27,7 @@ internal abstract class ConnectionStringPostProcessor : IConfigurationPostProces
         return new DbConnectionStringBuilderWrapper(new DbConnectionStringBuilder());
     }
 
-    public void PostProcessConfiguration(PostProcessorConfigurationProvider provider, IDictionary<string, string> configurationData)
+    public void PostProcessConfiguration(PostProcessorConfigurationProvider provider, IDictionary<string, string?> configurationData)
     {
         IConfigurationRoot parentConfiguration = provider.Source.GetParentConfiguration();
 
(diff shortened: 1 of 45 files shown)


src/Connectors/src/Connectors/ConnectionStringPostProcessor.cs
```diff
@@ -122,19 +122,19 @@ private IEnumerable<IConfigurationSection> GetBindingSections(IConfiguration con
...
         if (bindingInfo.ClientBindingSection != null)
         {
             foreach (IConfigurationSection secretSection in bindingInfo.ClientBindingSection.GetChildren())
             {
                 string secretName = secretSection.Key;
-                string secretValue = secretSection.Value;
+                string? secretValue = secretSection.Value;
 
-                if (string.Equals(secretName, ConnectionStringName, StringComparison.OrdinalIgnoreCase) && !string.IsNullOrEmpty(secretValue))
+                if (string.Equals(secretName, ConnectionStringName, StringComparison.OrdinalIgnoreCase))
```
TimHess, line 137: Is there no longer a concern that `secretValue` could be an empty string?

TimHess (approved):
If the one question I raised isn't an issue, this is OK to merge

src/Connectors/src/Connectors/ConnectionStringPostProcessor.cs
bart-vmware: This is intentional. The setter of `IConnectionStringBuilder.ConnectionString` accepts `null` as a parameter value, as defined [here](https://github.com/SteeltoeOSS/Steeltoe/blob/71b633f18da79661f9e68e7bfe809f1fc5f57d99/src/Connectors/src/Connectors/IConnectionStringBuilder.cs#L20-L21).

In the case of a server binding (code below), a `null` secret value originates from a post-processor that needs to clear a locally set value (for example [here](https://github.com/SteeltoeOSS/Steeltoe/blob/71b633f18da79661f9e68e7bfe809f1fc5f57d99/src/Configuration/src/Kubernetes.ServiceBinding/PostProcessors/RedisKubernetesPostProcessor.cs#L29)). But here we're dealing with client bindings. Today we have no code that overrules earlier client bindings, but someone might set up configuration providers in that way for some reason.

It wouldn't be correct to fall into the `else` clause when the _secret value_ is null or empty, because the _secret name_ determines whether it's a separate secret. Whether the secret value is `null`, an empty string, or something else shouldn't matter.

So effectively, this change is a bug fix for a potential scenario that's unlikely to occur.

Issue #23, opened 2018-08-24
Async Health() method in IHealthContributor interface


jkonicki:
Note: Moved from Connectors repository

Origin: User feedback
Issue 1: Is there a reason Steeltoe doesn't have an async Health() method within this Steeltoe.Common.HealthChecks.IHealthContributor interface?

Issue 2: Any reason a noun like Health() was chosen as a method name vs. something like AssessHealth()? Could we look into changing this?

ahaleiii:
Does Steeltoe block the thread when running `Health()` on IHealthContributor? Does this mean that there should not be implementations of IHealthContributor for external network calls?

hananiel:
No it doesn’t block. The middleware is invoked asynchronously.

hananiel:
Steeltoe supports the IHealthContributor to mirror the Spring Boot [interfaces](https://github.com/spring-projects/spring-boot/blob/v2.3.1.RELEASE/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/HealthIndicator.java). 
Steeltoe also supports the Microsoft IHealthCheck [interfaces](https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.diagnostics.healthchecks.ihealthcheck?view=dotnet-plat-ext-3.1).

We will create a sample using IHealthCheck, continue supporting IHealthContributor for teams already using them and change the healthCheck middleware to check health in parallel.

hananiel:
resolved by #371

jkonicki:
PR #364 added parallel processing for existing Steeltoe Health checks.

PR #1187, opened 2023-09-27
Make health contributors async

### LINKED ISSUE #878: Health Endpoint responds incorrectly to group names
Non existent group names for example actuator/health/foo should return 404.

bart-vmware:
## Description

This PR changes `IHealthContributor` to become async and take a `CancellationToken`. This improves scalability and removes a fair number of `task.GetAwaiter().GetResult()` calls.

Additionally, I've cleaned up one of the tests that often fail during cibuild; let's hope this improves reliability, or at least give some insight into why they randomly fail.

I found the existing timeout/cancellation behavior of `IMessageChannel.SendAsync` and `IPollableChannel.ReceiveAsync` awkward, but did not change it. They take a `CancellationToken` that the caller must trigger to force a timeout. Because of this, there is no way to distinguish between a timeout downstream versus the incoming request being canceled. In both cases, it throws an `OperationCanceledException` and there's no way to distinguish between them. Instead, these methods should take both a `CancellationToken` and a `TimeSpan timeout` parameter. When the timeout expires, it should throw a `TimeoutException`. Internally, all async methods delegate to sync implementations, which should be the other way around. Unfortunately, `RabbitMQ.Client` does not provide cancellable async APIs, so we can't fully solve this.

Fixes #878.

## Quality checklist

### DIFF
--- a/src/Integration/test/Integration.Test/Channel/QueueChannelTest.cs
+++ b/src/Integration/test/Integration.Test/Channel/QueueChannelTest.cs
@@ -169,7 +169,7 @@ public sealed class QueueChannelTest
             IMessage message = await channel.ReceiveAsync(cancellationTokenSource.Token);
             messageNull = message == null;
             latch.Signal();
-        });
+        }, CancellationToken.None);
 
         cancellationTokenSource.Cancel();
         Assert.True(latch.Wait(10000));
@@ -214,7 +214,7 @@ public sealed class QueueChannelTest
             IMessage message = await channel.ReceiveAsync(cancellationTokenSource.Token);
             messageNull = message == null;
             latch.Signal();
-        });
+        }, CancellationToken.None);
 
         cancellationTokenSource.Cancel();
         Assert.True(latch.Wait(10000));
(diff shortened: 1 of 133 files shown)


src/Integration/test/Integration.Test/Channel/QueueChannelTest.cs
```diff
@@ -13,338 +14,223 @@ namespace Steeltoe.Integration.Test.Channel;
 
 public sealed class QueueChannelTest
 {
+    private static readonly TimeSpan DefaultTestTimeout = TimeSpan.FromSeconds(5);
```
TimHess, line 17: @hananiel, please review the changes in this file

TimHess:
#23

src/Integration/test/Integration.Test/Channel/QueueChannelTest.cs
hananiel: A lot of these tests follow what's in spring closely - so deviating from it most likely will cause maintenance headaches. Second reading the diff, although the final pieces are same - I am not sure the tests are testing the same thing. The question is not how tasks work in C# but if the multithreaded context in which it will finally run is still valid - hence the synchronization mechanisms. I do not recommend this change.

Issue #1186, opened 2023-09-27
Service is not register in Eureka because of Autofac


xXAvoraXx:
## Describe the bug

I am trying to register my .net 7 web api application to eureka server, but I am getting the following error, why?

Code Explain

I use Autofac for service registrations, I make autofac configurations in a separate dotnet 7 library project. Even though I wrote the necessary configurations to connect to the Eureka service, I get an error.

**Codes:** CoreModule 

```
    using Project.Services.Core.Extensions;
    using Project.Services.Core.Utilities.IoC;
    using Microsoft.Extensions.DependencyInjection;
    using StackExchange.Redis;
    using System.Diagnostics;
    using Serilog;
    using Microsoft.Extensions.Configuration;
    using Steeltoe.Discovery.Client;
    using Steeltoe.Discovery.Eureka;
    
    namespace Project.Services.Core.DependencyResolvers
    {
        public class CoreModule : ICoreModule
        {
            public void Load(IServiceCollection services, IConfiguration configuration)
            {
                services.AddServiceDiscovery(o => o.UseEureka());
    
    
                services.AddMemoryCache();
                services.AddSingleton<ICacheManager, MemoryCacheManager>();
                services.AddSingleton<IDistributedCacheManager, DistributedCacheManager>();
                services.AddHttpContextAccessor();
                services.AddSingleton<Stopwatch>();
    
            }
        }
    }
```

**Program.cs**

```
    builder.Services.AddDependencyResolvers(new ICoreModule[]
                {
                    new CoreModule()
                }, builder.Configuration);
```

**appsettings.json config**

```
      "Spring": {
        "Application": {
          "Name": "STOCK-SERVICE"
        }
      },
      "Eureka": {
        "Client": {
          "ServiceUrl": "http://localhost:867/eureka/",
          "ValidateCertificates": false,
          "ShouldRegisterWithEureka": true
        },
    
        // No docke instance added
        "Instance": {
          //"port": 5004,
          "NonSecurePort": 5004,
          "HostName": "localhost",
          "InstanceId": "Stock.API,Port:5004",
          "StatusPageUrlPath": "/swagger/index.html"
          //"HealthCheckUrlPath": "/api/values/healthcheck"
        }
      }
```

**Error Details**

```
    Autofac.Core.DependencyResolutionException
      HResult=0x80131500
[...]

PR #1189, opened 2023-09-29
Handle trailing semicolon in address list

TimHess:
Addresses #1186 in 3.2

### DIFF
--- a/src/Discovery/src/ClientBase/ConfigurationUrlHelpers.cs
+++ b/src/Discovery/src/ClientBase/ConfigurationUrlHelpers.cs
@@ -5,6 +5,7 @@
 using Microsoft.Extensions.Configuration;
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Steeltoe.Discovery.Client;
 
@@ -19,7 +20,7 @@ public static class ConfigurationUrlHelpers
         if (!string.IsNullOrEmpty(urls))
         {
             var addresses = urls.Split(';');
-            foreach (var address in addresses)
+            foreach (var address in addresses.Where(a => !string.IsNullOrEmpty(a)))
             {
                 if (!Uri.TryCreate(address, UriKind.Absolute, out var uri)
                     && (address.Contains("*") || address.Contains("::") || address.Contains("+")))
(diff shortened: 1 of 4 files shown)


src/Discovery/src/ClientBase/ConfigurationUrlHelpers.cs
```diff
@@ -19,7 +20,7 @@ public static List<Uri> GetAspNetCoreUrls(this IConfiguration config)
         if (!string.IsNullOrEmpty(urls))
         {
             var addresses = urls.Split(';');
-            foreach (var address in addresses)
+            foreach (var address in addresses.Where(a => !string.IsNullOrEmpty(a)))
```
bart-vmware, line 23: Why not just specify the [split behavior](https://learn.microsoft.com/en-us/dotnet/api/system.string.split?view=net-6.0#system-string-split(system-char-system-stringsplitoptions)) directly?

```c#
var addresses = urls.Split(';', StringSplitOptions.RemoveEmptyEntries);
```

It's probably not so relevant here, but these things often get vectorized for maximum performance. Expressing intent instead of implementation enables the runtime to optimize such calls (today or in the future).

There's also a flag to trim entries, though I don't know if we need that here.

src/Discovery/test/ClientBase.Test/ConfigurationUrlHelpers.cs
```diff
@@ -0,0 +1,32 @@
...
+using Xunit;
+
+namespace Steeltoe.Discovery.ClientBase.Test;
+
+public class ConfigurationUrlHelpers
+{
+    [Fact]
+    public void GetAspNetCoreUrlsHandlesTrailingSemicolon()
+    {
+        var config = new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string, string>() { { "urls", "https://myapp:1234;http://0.0.0.0:1233;http://::1233;http://*:1233;" } }).Build();
```
bart-vmware, line 20: Is the double colon in `http://::1233` intentional? Is that a valid value?

TimHess: It is intentional, but it is not really valid. The format should be [::], but this is the way it has been so I didn't want to change it in 3.2

bart-vmware (approved):
LGTM

PR #1196, opened 2023-10-04
Add support for Application Configuration Service

### LINKED ISSUE #1195: Support Tanzu Application Configuration Service
Given that [Tanzu Application Configuration Service](https://docs.vmware.com/en/Application-Configuration-Service-for-VMware-Tanzu/2.0/acs/GUID-overview.html) can be viewed as a successor to Spring Cloud Config Server, Steeltoe should also support reading configuration provided by it.

Preliminary support provided by the new Kubernetes.ServiceBindings package already reads in configuration provided by the service and for a configurationSlice named "spring-properties" will result in IConfiguration entries such as this:

```json
"k8s:bindings:spring-properties:cook.special": {
  "value": "Pickled Cactus"
},
"k8s:bindings:spring-properties:dessertMenu": {
  "value": "Strawberry Cake with Whipped Cream"
},
"k8s:bindings:spring-properties:provider": {
  "value": "acs"
},
"k8s:bindings:spring-properties:secretMenu": {
  "value": "Crawfish Casserole"
},
"k8s:bindings:spring-properties:type": {
  "value": "config"
}
```

We should remap any non-system keys to the root of the configuration so they can be easily accessed by developers. We should not map system keys like "provider" or "type" as those are only likely to cause confusion

TimHess:
## Description

Adds a post-processor that remaps all non-system keys (eg: type and provider) to the root of the IConfiguration

Fixes #1195 

## Quality checklist


TimHess:
This should be ready after we make sure hierarchy is support in config keys via double-underscore, the same as documented in [Microsoft's key-per-file provider](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/configuration/?view=aspnetcore-8.0#key-per-file-configuration-provider)

TimHess:
This work is being deprioritized right now, if it would mean something to you (the reader of this message) please let us know and we'll finish it up and ship it

bart-vmware:
This PR has test failures in Connectors, which doesn't show up in the checks because Steeltoe.All has not run. This proves having only component builds is fundamentally unreliable, as the trigger conditions are unaware of project dependencies. So I propose we always run the full build on PRs, not as a manual step.

TimHess:
> This PR has test failures in Connectors, which doesn't show up in the checks because Steeltoe.All has not run. This proves having only component builds is fundamentally unreliable, as the trigger conditions are unaware of project dependencies. So I propose we always run the full build on PRs, not as a manual step.

Or we could add the dependencies to the component builds... but Steeltoe.All does run a lot faster than it did a year or two ago, so I'm open to trying out that change

src/Common/test/Common.Test/Configuration/StringExtensionsTest.cs
```diff
@@ -0,0 +1,29 @@
...
+    [InlineData("foo.bar", "foo:bar")]
+    [InlineData("foo__bar", "foo:bar")]
+    [InlineData("foo__bar_", "foo:bar_")]
+    [InlineData("foo__bar__", "foo:bar__")]
+    [InlineData("foo[bar]", "foo[bar]")]
+    [InlineData("foobar[1234]", "foobar:1234")]
+    [InlineData("foobar[1234][5678]", "foobar:1234:5678")]
+    [InlineData("foobar[1234][5678]barbar", "foobar[1234][5678]barbar")]
+    [InlineData("a.b.foobar[1234][5678].barfoo.boo[123]", "a:b:foobar:1234:5678:barfoo:boo:123")]
+    [InlineData(@"a.b\.foobar", "a:b.foobar")]
```
bart-vmware, line 24: These should contain three segments (to ensure it works multiple times) of different lengths (to ensure we're at the right offsets), ie: `one.four.seven` instead of `foo.bar`.

I'm wondering if the following cases are correct:
- `foo__bar__`: Looking into the built-in .NET implementation, it doesn't preserve the trailing __, so why should we?
- `foobar[1234][5678]barbar` is this a legal key, or should it be: `foobar[1234][5678].barbar`?
- `a.b\.foobar` would be better to cover both escape cases, ie: `[InlineData(@"one\.four\\.seven", @"one.four\:seven")]`

src/Common/test/Common.Test/Configuration/StringExtensionsTest.cs
```diff
@@ -0,0 +1,29 @@
...
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using FluentAssertions;
+using Steeltoe.Common.Configuration;
+using Xunit;
+
+namespace Steeltoe.Common.Test.Configuration;
+
+public sealed class StringExtensionsTest
```
bart-vmware, line 11: It's unclear to me what escaping rules come into play. I've googled a bit and found the following:
- yaml keys that contain a dot or whitespace (and several other characters) need to be surrounded by brackets
	https://stackoverflow.com/questions/34070987/escaping-a-dot-in-a-map-key-in-yaml-in-spring-boot
	https://docs.spring.io/spring-boot/docs/2.2.1.RELEASE/reference/html/spring-boot-features.html#boot-features-external-config-relaxed-binding
	https://artofcode.wordpress.com/2020/10/29/special-characters-in-a-key-in-spring-boot-yaml-file/
	https://stackoverflow.com/questions/57795723/sping-boot-config-how-to-keep-whitespace-in-yaml-key-being-used-to-populate-map
	https://stackoverflow.com/questions/65346054/ignore-specify-slash-character-in-yaml-config-for-spring-boot-admin-application
- special handling of \$
	https://docs.spring.io/spring-cloud-config/reference/server/environment-repository/property-overrides.html
- unicode escapes and backslashes
	https://www.w3schools.io/file/properties-escape-characters/
	https://en.wikipedia.org/wiki/.properties
	https://www.kajabity.com/kajabity-tools/java-properties-classes/the-java-properties-file-format/

It would help to point to some spec references (in tests or in the implementation), so we'll know what needs to be accounted for,

src/Common/src/Common/Configuration/StringExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+
+    private static IEnumerable<string> UniversalHierarchySplit(string source)
+    {
+        ArgumentGuard.NotNull(source);
+
+        var result = new List<string>();
+
+        int segmentStart = 0;
+
+        for (int i = 0; i < source.Length; i++)
```
bart-vmware, line 48: I'm concerned about performance, looping over every character in every key and running non-compiled regular expressions. It could be optimized using `string.IndexOfAny()`, which is vectorized. I'd like to propose something more efficient, once I better understand the escaping rules.

bart-vmware:
> Or we could add the dependencies to the component builds... but Steeltoe.All does run a lot faster than it did a year or two ago, so I'm open to trying out that change

That'll break sooner than later. I prefer to add Steeltoe.All, so we'll have some quick feedback from the component builds, while at the same time validating the full build.

bart-vmware (commented):
I didn't get past the first few files yet, but here's some initial feedback.

src/Common/src/Common/Configuration/StringExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+{
+    private const string DotDelimiterString = ".";
+    private const char DotDelimiterChar = '.';
+    private const char UnderscoreDelimiterChar = '_';
+    private const char EscapeChar = '\\';
+    private const string EscapeString = "\\";
+
+    private static readonly Regex ArrayRegex = new(@"(\[[0-9]+\])*$");
+
+    public static string AsDotNetConfigurationKey(this string key)
```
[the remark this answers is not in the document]
TimHess: I added support here for handling double-underscores in the keys, so it isn't just handling things from Spring formats, but I'm fine with changing it from being a string extension

src/Common/test/Common.Test/Configuration/StringExtensionsTest.cs
TimHess: > I'm wondering if the following cases are correct:
> 
> * `foo__bar__`: Looking into the built-in .NET implementation, it doesn't preserve the trailing __, so why should we?
> * `foobar[1234][5678]barbar` is this a legal key, or should it be: `foobar[1234][5678].barbar`?
> * `a.b\.foobar` would be better to cover both escape cases, ie: `[InlineData(@"one\.four\\.seven", @"one.four\:seven")]`

The last two are pre-existing test cases from the ConfigServer code. I don't mind adding the additional escaping, and I think the point of the `[1234][5678]` case is that it should be left alone. For the trailing underscore cases, I didn't think they would be valid, so intended to leave keys with those patterns as-is.

src/Common/src/Common/Configuration/StringExtensions.cs
TimHess: This is mostly pre-existing code, I didn't attempt to reengineer it. I think all I changed here is the naming, location and addition of `__` handling, which is think is all that's needed at this point, but if you've got some ideas I'm open

src/Common/test/Common.Test/Configuration/StringExtensionsTest.cs
TimHess: I am also unclear on what should apply. Escaping was not a topic I was intending to address in any way other than preserving what was here before

src/Configuration/src/SpringBoot/SpringBootCommandLineProvider.cs
```diff
@@ -39,8 +40,7 @@ public override void Load()
 
             if (key.StartsWith(KeyPrefix, StringComparison.OrdinalIgnoreCase))
             {
-                string newKey = key.Replace('.', ':');
-                Data[newKey] = pair.Value;
+                Data[ConfigurationKeyConverter.AsDotNetConfigurationKey(key)] = pair.Value;
```
bart-vmware, line 43: Why call the new method here? I'd prefer to isolate usage of the questionable implementation as much as possible, instead of using it everywhere.

src/Configuration/src/SpringBoot/SpringBootEnvironmentVariableProvider.cs
```diff
@@ -60,13 +61,7 @@ public override void Load()
...
             foreach (string key in keys)
             {
-                string? value = Data[key];
-
-                if (key.Contains('.') && value != null)
-                {
-                    string newKey = key.Replace('.', ':');
-                    Data[newKey] = value;
-                }
+                Data[ConfigurationKeyConverter.AsDotNetConfigurationKey(key)] = Data[key];
```
bart-vmware, line 64: Same concern here: I wouldn't want to suggest escaping/arrays/etc can be used in environment variables, unless there's a use case.

src/Configuration/test/Kubernetes.ServiceBinding.Test/PostProcessorsTest.cs
```diff
@@ -147,4 +148,57 @@ public void Processes_Redis_configuration()
...
     }
+
+    [Fact]
+    public void Processes_ApplicationConfigurationService_ConfigurationData()
+    {
+        var postProcessor = new ApplicationConfigurationServicePostProcessor();
+
+        var secrets = new[]
+        {
+            Tuple.Create("provider", "acs"),
```
bart-vmware, line 159: This should include "type" as well (see assertion below).

bart-vmware (changes requested):
Completed review pass.

src/Common/test/Common.Test/Configuration/StringExtensionsTest.cs
bart-vmware: Related https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-Configuration-Binding#map-based-binding and https://github.com/SteeltoeOSS/Steeltoe/commit/42df149c7d11ae53e7b3d85157a66985ff98f18e.

src/Configuration/src/SpringBoot/SpringBootEnvironmentVariableProvider.cs
TimHess: I think this provider is related to Spring Cloud Data Flow, which may or may not even be relevant anymore... Apparently the docs were never written for either of the spring boot config providers - I'm inclined to leave these with the shared method since this one should have already had support for `__` too but I don't think that would have worked before

src/Management/src/Endpoint/Info/Contributor/GitInfoContributor.cs
```diff
@@ -72,7 +73,7 @@ public async Task ContributeAsync(IInfoBuilder builder, CancellationToken cancel
                         continue;
                     }
 
-                    string key = keyValuePair[0].Trim().Replace('.', ':');
+                    string key = ConfigurationKeyConverter.AsDotNetConfigurationKey(keyValuePair[0].Trim());
```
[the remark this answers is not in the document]
TimHess: fair, it is definite overkill here

src/Configuration/src/Kubernetes.ServiceBinding/PublicAPI.Unshipped.txt
```diff
@@ -1,5 +1,6 @@
 #nullable enable
 static Steeltoe.Configuration.Kubernetes.ServiceBinding.ConfigurationBuilderExtensions.AddKubernetesServiceBindings(this Microsoft.Extensions.Configuration.IConfigurationBuilder! builder) -> Microsoft.Extensions.Configuration.IConfigurationBuilder!
+static Steeltoe.Configuration.Kubernetes.ServiceBinding.ConfigurationBuilderExtensions.AddKubernetesServiceBindings(this Microsoft.Extensions.Configuration.IConfigurationBuilder! builder, bool optional) -> Microsoft.Extensions.Configuration.IConfigurationBuilder!
```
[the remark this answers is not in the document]
TimHess: I'll just remove it

nuget.config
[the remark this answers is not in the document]
TimHess: I don't know what changed, you could be right... I changed the line endings to LF and changes are gone now

src/Configuration/src/SpringBoot/SpringBootCommandLineProvider.cs
bart-vmware: I disagree. From what we understand so far, the method was created to handle escaped arrays in YAML files that exist in a git repo being served through Config Server. That's a very specific scenario, which doesn't apply to all other places. We haven't actually verified what ConfigServer makes of YAML files, so we may as well be unescaping too much here, or handling the wrong escape sequence. I'd rather isolate that uncertainty to ConfigServer, than to spread it over various places. If YAML files use brackets to escape values, then why does our code only allow numbers between brackets? Does ConfigServer unescape them already? There are many more questions, and it requires more investigation and trying it out to see what Steeltoe needs to handle.

If all the other places are wrong, we should have evidence of that and fix them accordingly. They may just as well need a different escape scheme. Just calling one complex method that's poorly defined (and with concerning performance impact) from everywhere, only in case it might help anyone, may as well work against us. Enhancing a simple conversion as a bugfix is fine, but introducing breaking changes because we fix one place and break others without noticing is not.

src/Configuration/src/ConfigServer/Steeltoe.Configuration.ConfigServer.csproj
```diff
@@ -18,8 +18,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\Common\src\Common.Hosting\Steeltoe.Common.Hosting.csproj" />
     <ProjectReference Include="..\..\..\Common\src\Common.Http\Steeltoe.Common.Http.csproj" />
     <ProjectReference Include="..\CloudFoundry\Steeltoe.Configuration.CloudFoundry.csproj" />
+    <ProjectReference Include="..\Kubernetes.ServiceBinding\Steeltoe.Configuration.Kubernetes.ServiceBinding.csproj" />
     <ProjectReference Include="..\Placeholder\Steeltoe.Configuration.Placeholder.csproj" />
     <ProjectReference Include="..\..\..\Discovery\src\Client\Steeltoe.Discovery.Client.csproj" PrivateAssets="All" />
```
bart-vmware, line 26: Why is this a private reference? Isn't that just satisfying the compiler but causing a runtime error when the package isn't loaded? It's needed by `ConfigServerDiscoveryService`, but that method isn't decorated with `[MethodImpl(MethodImplOptions.NoInlining)]` like it was in the Bootstrap project.

src/Configuration/src/ConfigServer/Steeltoe.Configuration.ConfigServer.csproj
```diff
@@ -18,8 +18,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\Common\src\Common.Hosting\Steeltoe.Common.Hosting.csproj" />
     <ProjectReference Include="..\..\..\Common\src\Common.Http\Steeltoe.Common.Http.csproj" />
     <ProjectReference Include="..\CloudFoundry\Steeltoe.Configuration.CloudFoundry.csproj" />
+    <ProjectReference Include="..\Kubernetes.ServiceBinding\Steeltoe.Configuration.Kubernetes.ServiceBinding.csproj" />
```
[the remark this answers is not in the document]
bart-vmware: This can be removed, it is already referenced indirectly.

src/Configuration/src/ConfigServer/Steeltoe.Configuration.ConfigServer.csproj
```diff
@@ -18,8 +18,10 @@
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\..\..\Common\src\Common.Hosting\Steeltoe.Common.Hosting.csproj" />
```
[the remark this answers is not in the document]
bart-vmware: This can be removed, it is already referenced indirectly.

src/Common/src/Common/Configuration/StringExtensions.cs
```diff
@@ -0,0 +1,89 @@
...
+
+public static class StringExtensions
+{
+    private const string DotDelimiterString = ".";
+    private const char DotDelimiterChar = '.';
+    private const char UnderscoreDelimiterChar = '_';
+    private const char EscapeChar = '\\';
+    private const string EscapeString = "\\";
+
+    private static readonly Regex ArrayRegex = new(@"(\[[0-9]+\])*$");
```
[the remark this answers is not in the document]
bart-vmware: This wasn't addressed properly. It is still using imperative code for the replacement. Instead, use:
```c#
return ArrayRegex.Replace(key, ":${digits}");
```

bart-vmware (changes requested):

bart-vmware (approved):

PR #1198, opened 2023-10-09
Fix for Configuration changes are not detected.

### LINKED ISSUE #1174: Management/health configuration changes are not detected
## Describe the bug
When the disk space contributor configuration changes at runtime, the changes are not applied at the `/actuator/health` endpoint.

## Steps to reproduce 
1. Use the PostgreSQL sample at https://github.com/SteeltoeOSS/Samples/tree/latest/Connectors/src
2. Add the following to `appsettings.Development.json`:
  ```json
  "management": {
    "endpoints": {
      "health": {
        "diskspace": {
          "enabled": true,
          "path": "c:\\"
        }
      }
    }
  }
  ```
3. Run the application and navigate to `/actuator/health` in the browser
4. Change "enabled" to `false` in `appsettings.Development.json` and save the file
5. Refresh the page and observe the diskSpace contributor is still shown.
6. Restart the application and observe the diskSpace contributor is no longer shown.

## Expected behavior
Configuration changes are applied at runtime without restarting the application.

## Workaround
Add the next line at the top of `Program.cs`:
```c#
builder.Services.Configure<DiskSpaceContributorOptions>(builder.Configuration.GetSection(
    "management:endpoints:health:diskspace"));
```

## Analysis
The IOptions pattern isn't correctly implemented throughout the Management namespace. In this case, `ConfigureDiskSpaceContributorOptions` is registered to execute each time ASP.NET creates a `DiskSpaceContributorOptions` instance. When it executes, it performs a one-time bind of a configuration section against the to-be-configured options instance using:
```c#
_configuration.GetSection("management:endpoints:health:diskspace").Bind(options)
```

However, there's no code that subscribes to the `IConfiguration` change token. Therefore, `ConfigureDiskSpaceContributorOptions` executes only once at application startup, where it manually binds from a configuration section.

Once the proper subscription is in place (as described in the workaround above), the code to manually bind from a configuration section should be removed because ASP.NET already does that. Aside from not working properly, the current implementation overwrites the result of any previous Configure callbacks that consumers may have registered. `IConfigureOptions<>` is designed to adapt existing options, not blindly overwrite their values.

For an overview of how the IOption pattern is designed, see the blog post series at https://andrewlock.net/configuring-named-options-using-iconfigurenamedoptions-and-configureall/.

## Environment:
 - Steeltoe Version: main branch at commit eff40d44435ff19ddb58c03a41ba56340159e327
 - Platform: local
 - OS: Windows
 - .NET Version: .NET 6

hananiel:
## Description
Resolved the issue with updating configuration when IOptionsMonitor is used. As described in the issue, the `IOptionsChangeTokenSource` is not automatically registered when using some of provided extension methods. So we are registering it ourselves. 

I tried to eliminate the `Bind` method, but its not possible without passing around the IConfiguration to the service extension - so this is the next best alternative I found. 

Fixes #1174 

## Quality checklist


src/Management/src/Endpoint/Options/ConfigureEndpointOptions.cs
bart-vmware, line 1: `Configuration.GetSection(_prefix).Bind(options);` (line 35) is the problematic part that needs to go away. It works the other way around. Instead of binding when `Configure` executes, options must have already been bound upfront, so the `Configure` method only changes existing properties instead of discarding what happened before. There can be multiple configure callbacks, right now only changes from the last one are applied.

hananiel: If we don't do it here, where will it read from configuration? See this example for how its being done in microsoft code to implement` IConfigureOptions<T>` : They are not using the `Bind` method, but iterating over config [key/values].(https://github.com/dotnet/runtime/blob/8d2ada730b9a44aa630014eb0da1aa907e07c437/src/libraries/Microsoft.Extensions.Diagnostics/src/Metrics/Configuration/MetricsConfigureOptions.cs#L97C26-L97C41) . Are you oppsed to the bind method or just reading from configuration at this point? This fires before any postconfigures so I dont see where there is anything changing before this except the notification that a configuration source has changed and we need to rebind.

bart-vmware: > If we don't do it here, where will it read from configuration?

It happens automatically. As described at https://andrewlock.net/configuring-named-options-using-iconfigurenamedoptions-and-configureall/, the basic pattern is:
1. Connect options type to configuration path (optional, registers on change token):
    ```c#
    IServiceCollection services;
    services.Configure<MyOptions>(Configuration.GetSection("Path:To:MyOptions")); 
    ```
1. Configure options using callbacks, optional and potentially multiple times:
    ```c#
    IServiceCollection services;
    services.Configure<MyOptions>(options => options.Value += "X"); 
    services.Configure<MyOptions>(options => options.Value += "Y"); 
    services.PostConfigure<MyOptions>(options => options.Value += "Z"); 
    ```
1. Obtain singleton, snapshot or monitor:
    ```c#
    IServiceProvider serviceProvider;
    var snapshot = serviceProvider.GetRequiredService<IOptionsMonitor<MyOptions>>();
    ```

> See this example for how its being done in microsoft code to implement` IConfigureOptions<T>` : They are not using the `Bind` method, but iterating over config [key/values].(https://github.com/dotnet/runtime/blob/8d2ada730b9a44aa630014eb0da1aa907e07c437/src/libraries/Microsoft.Extensions.Diagnostics/src/Metrics/Configuration/MetricsConfigureOptions.cs#L97C26-L97C41).

That doesn't seem to follow the pattern. Perhaps they are doing it like that to support a more complicated configuration structure?

> Are you opposed to the bind method or just reading from configuration at this point?

Configurers generally shouldn't read from configuration, only update what was already provided. This enables callers to build the options from code instead of configuration, which is more flexible.

> This fires before any postconfigures so I dont see where there is anything changing before this except the notification that a configuration source has changed and we need to rebind.

See my example above, there could be multiple for various reasons.

hananiel: Consider the alternative to this (what you consider "The pattern"). We will walk through configuring the Ioptions pattern using your preferred method: 


Config

Steps into the following sequentially: 

> Source : Microsoft.Extensions.Options.ConfigurationExtensions.nuget
>     -> public static class OptionsConfigurationServiceCollectionExtensions

config2
config3
config4
> Source : Microsoft.Extensions.Options.ConfigurationExtensions.nuget
>   -> public class NamedConfigureFromConfigurationOptions<TOptions> : ConfigureNamedOptions<TOptions>
   
config5
>   -> public class NamedConfigureFromConfigurationOptions<TOptions> : ConfigureNamedOptions<TOptions>
config6
>  -> public class ConfigureNamedOptions<TOptions> ...
Which is exactly what we were doing before. However I did miss adding the `IOptionsChangeTokenSource`. And since I copied your suggested changes, I think there is still a difference, but either way, I think I think binding to configuration in the configuremethod is fine - unless I am missing something else.

bart-vmware: Thanks for taking the time to help me understand I was wrong. While documented as disparate mechanisms, under the covers it's using the same. So I now agree this change is the right solution.

bart-vmware (changes requested):

bart-vmware (approved):
LGTM

Issue #1199, opened 2023-10-09
Authentication fails when using service discovery on Azure Spring Apps Enterprise


TimHess:
## Describe the bug
App fails to register with discovery server secured with .p12 certificate

## Environment (please complete the following information):
 - Platform: Azure Spring Apps Enterprise
 - Found with Steeltoe 3.2.4, present in 4.0 line as well

## Additional context or links
Stack trace:
```
fail: Steeltoe.Discovery.Eureka.Transport.EurekaHttpClient[0]                                                                         
      RegisterAsync Failed, request was made to https://<instance>.svc.azuremicroservices.io/eureka/default/eureka/apps/<app>, retry: 0
      System.Net.Http.HttpRequestException: The SSL connection could not be established, see inner exception.                                    
       ---> System.Security.Authentication.AuthenticationException: Authentication failed, see inner exception.                                  
       ---> System.ComponentModel.Win32Exception (0x8009030E): No credentials are available in the security package                              
         at System.Net.SSPIWrapper.AcquireCredentialsHandle(ISSPIInterface secModule, String package, CredentialUse intent, SCH_CREDENTIALS* scc)
         at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(CredentialUse credUsage, SCH_CREDENTIALS* secureCredential)
         at System.Net.Security.SslStreamPal.AcquireCredentialsHandleSchCredentials(SslStreamCertificateContext certificateContext, SslProtocols protocols, EncryptionPolicy policy, Boolean isServer)
         at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(SslStreamCertificateContext certificateContext, SslProtocols protocols, EncryptionPolicy policy, Boolean isServer)
         --- End of inner exception stack trace ---
         at System.Net.Security.SslStreamPal.AcquireCredentialsHandle(SslStreamCertificateContext certificateContext, SslProtocols protocols, EncryptionPolicy policy, Boolean isServer)
         at System.Net.Security.SecureChannel.AcquireClientCredentials(Byte[]& thumbPrint)
         at System.Net.Security.SecureChannel.GenerateToken(ReadOnlySpan`1 inputBuffer, Byte[]& output)
         at System.Net.Security.SecureChannel.NextMessage(ReadOnlySpan`1 incomingBuffer)
         at System.Net.Security.SslStream.ProcessBlob(Int32 frameSize)
[...]

PR #1200, opened 2023-10-09
Fix service discovery auth issue on ASA-E

TimHess:
## Description

Use the default constructor when creating an x509Certificate from a path on disk

Fix for 3.2 #1199

## Quality checklist

### DIFF
--- a/src/Common/src/Common.Security/ConfigureCertificateOptions.cs
+++ b/src/Common/src/Common.Security/ConfigureCertificateOptions.cs
@@ -40,7 +40,7 @@ public class ConfigureCertificateOptions : IConfigureNamedOptions<CertificateOpt
             return;
         }
 
-        options.Certificate = new X509Certificate2(certPath, string.Empty, X509KeyStorageFlags.EphemeralKeySet);
+        options.Certificate = new X509Certificate2(certPath);
     }
 
     public void Configure(CertificateOptions options)
--- a/src/Common/src/Common.Security/ConfigureCertificateOptions.cs
+++ b/src/Common/src/Common.Security/ConfigureCertificateOptions.cs
@@ -40,7 +40,7 @@ public class ConfigureCertificateOptions : IConfigureNamedOptions<CertificateOpt
             return;
         }
 
-        options.Certificate = new X509Certificate2(certPath);
+        options.Certificate = new X509Certificate2(certPath, string.Empty, X509KeyStorageFlags.EphemeralKeySet | X509KeyStorageFlags.Exportable);
     }
 
     public void Configure(CertificateOptions options)
--- a/src/Common/test/Common.Security.Test/CertificateRotationServiceTest.cs
+++ b/src/Common/test/Common.Security.Test/CertificateRotationServiceTest.cs
@@ -24,7 +24,7 @@ public class CertificateRotationServiceTest
     [Trait("Category", "SkipOnMacOS")]
     public async Task ServiceLoadsCertificate()
     {
-        var personalCertStore = new X509Store(StoreName.My, StoreLocation.CurrentUser);
+        using var personalCertStore = new X509Store(StoreName.My, StoreLocation.CurrentUser);
         personalCertStore.Open(OpenFlags.ReadWrite);
         personalCertStore.Certificates.Find(X509FindType.FindByIssuerName, "Diego Instance Identity Intermediate CA", false).Should().BeEmpty();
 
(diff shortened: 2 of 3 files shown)


src/Common/src/Common.Security/ConfigureCertificateOptions.cs
```diff
@@ -40,7 +40,7 @@ public void Configure(string name, CertificateOptions options)
             return;
         }
 
-        options.Certificate = new X509Certificate2(certPath, string.Empty, X509KeyStorageFlags.EphemeralKeySet);
+        options.Certificate = new X509Certificate2(certPath);
```
hananiel, line 43: By Default this is not EphemeralKeySet ([docs](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509keystorageflags?view=net-7.0)). I also found this [blog](https://www.pkisolutions.com/handling-x509keystorageflags-in-applications/).  I am also confused as to why this would fix the issue.

hananiel (approved):

src/Common/test/Common.Security.Test/CertificateRotationServiceTest.cs
```diff
@@ -0,0 +1,88 @@
...
+            Path.Combine(LocalCertificateWriter.AppBasePath, "GeneratedCertificates", "SteeltoeInstanceCert.pem"),
+            Path.Combine(LocalCertificateWriter.AppBasePath, "GeneratedCertificates", "SteeltoeInstanceKey.pem"));
+        await File.WriteAllBytesAsync(filename, cert.Export(X509ContentType.Pkcs12));
+        await Task.Delay(2000);
+
+        var newCollection = personalCertStore.Certificates.Find(X509FindType.FindByIssuerName, "Diego Instance Identity Intermediate CA", false);
+        newCollection.Should().NotIntersectWith(collection);
+#else
+        // avoid warning about missing await for .NET Core 3.1 target
+        await Task.Delay(1);
```
bart-vmware, line 72:
```suggestion
        await Task.Yield();
```

src/Common/test/Common.Security.Test/CertificateRotationServiceTest.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Steeltoe.Common.Security.Test;
+
+public class CertificateRotationServiceTest
+{
+    [Fact]
+    [Trait("Category", "SkipOnMacOS")]
+    public async Task ServiceLoadsCertificate()
+    {
+        var personalCertStore = new X509Store(StoreName.My, StoreLocation.CurrentUser);
```
bart-vmware, line 27:
```suggestion
        using var personalCertStore = new X509Store(StoreName.My, StoreLocation.CurrentUser);
```

src/Common/test/Common.Security.Test/CertificateRotationServiceTest.cs
```diff
@@ -0,0 +1,88 @@
...
+        var config = new ConfigurationBuilder()
+            .AddCertificateFile(filename)
+            .Build();
+        var services = new ServiceCollection();
+        services.AddSingleton<IConfiguration>(config);
+        services.AddOptions();
+        services.AddSingleton<IConfigureOptions<CertificateOptions>, ConfigureCertificateOptions>();
+        services.AddSingleton<ICertificateRotationService, CertificateRotationService>();
+
+        var serviceProvider = services.BuildServiceProvider();
```
bart-vmware, line 44:
```suggestion
        using var serviceProvider = services.BuildServiceProvider();
```

src/Common/test/Common.Security.Test/CertificateRotationServiceTest.cs
```diff
@@ -0,0 +1,88 @@
...
+        var service = serviceProvider.GetRequiredService<ICertificateRotationService>();
+
+        service.Start();
+
+        var collection = personalCertStore.Certificates.Find(X509FindType.FindByIssuerName, "Diego Instance Identity Intermediate CA", false);
+        collection.Should().NotBeNull();
+
+#if NET6_0_OR_GREATER
+        if (!File.Exists(Path.Combine(LocalCertificateWriter.AppBasePath, "GeneratedCertificates", "SteeltoeInstanceCert.pem")))
```
bart-vmware, line 53: Why check if the file exists? It uses the sandbox, so I would expect it to either always or never exist.

TimHess: This portion of the test is not using the sandbox. These certificates _may_ be generated by another test that has already run, or they may not be present yet. Ultimately, I was looking for an easy path to get valid certificate data that did not exactly match the .p12 certificate (I think the .pem files that are committed to source right next to it have the same content) so that we can detect changes. I don't feel strongly in either direction about keeping this check.

hananiel (approved):

Issue #962, opened 2022-07-29
Adopt nullable reference types


bart-vmware:
This issue tracks annotating the codebase for [nullable reference types](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-reference-types).

General guidance:
- Compile/test against .NET 7 or higher (only the core libraries are annotated in .NET 6, and several are wrong)
- Always declare collection properties using interfaces (`IList<JsonService>` instead of `List<JsonService>`)
- IOptions pattern
  - Declare collection properties as non-nullable _get-only_ and initialize with an empty collection
    - They cannot be set to null from JSON, see https://github.com/dotnet/extensions/issues/1341
    - The configuration binder supports adding to an existing collection instance
  - Always declare scalar reference-type properties as nullable (and null-check before use)
    - There's no guarantee that `IConfigureOptions<>` has executed before usage
    - Developers may register their own `IConfigureOptions<>` that sets the value to `null`
    - The property may be explicitly set to `null` from appsettings.json
- Types used in JSON serialization
  - Declare collection properties with both a getter and a setter
    - Although `[JsonObjectCreationHandling(JsonObjectCreationHandling.Populate)]` was added in .NET 8, it cannot be combined with custom converters (so adding one later would require a breaking change)
  - Deserialization
    - Declare collection properties as nullable if `null` has a semantic meaning that differs from an empty collection, or when it's unknown whether the source may produce nulls.
    - Initialize with an empty collection if not declared as nullable
  - Serialization
    - It's safest to declare all reference-type properties as nullable, because it affects the JSON output
      - If a collection is conditionally populated (ie `null` has meaning), make the property nullable
      - If a collection is always populated, declare it as non-nullable and initialize with an empty collection
- Other types
  - Declare all required properties as non-nullable
    - Replace property setters with a constructor that takes all required properties, and add guard checks
- Try to minimize the use of the `!` operator to suppress nullability warnings; consider adding a comment when unobvious

PR #1223, opened 2023-11-24
Bootstrap public API review

### LINKED ISSUE #1220: Review API Surface area in Steeltoe.Bootstrap

bart-vmware:
This PR entails the public API review of the `Steeltoe.Bootstrap` namespace.

## High-level overview of changes

- Unified logic to configure host builders through host-agnostic wrapper (this removes a lot of duplicated code)
- Make assembly loading non-static (easier to test)
- Replace `Regex` usage with more efficient parsing using `AssemblyName`
- Change exluded assemblies from list to read-only set
- Remove a few redundant host builder extensions (there are many more to consider)
- Fix skipped tests
- Fix tests on management ports (do not use TestServer, which listens on all ports)

## Bugfixes (with added test coverage)

- Correct the bootstrap logger name used
- Correct omitted registrations and/or arguments while unifying host builder extensions
- Use case-insensitive comparison of assembly names
- Prometheus auto-wire did not work

## Considerations from https://github.com/SteeltoeOSS/Steeltoe/issues/897#issuecomment-1198088843

- [x] Review namespaces
  - [x] Consider the use of [pubternal](https://www.reddit.com/r/dotnet/comments/jwzgl9/word_of_the_day_pubternal/)
  - [x] Adjust namespaces based on directory (or adapt project structure to match namespaces)
- [x] Consider making public types/members internal, use `InternalsVisibleTo` to access them from tests
- [x] Seal public types, unless they are designed for inheritance
- [x] Avoid static classes, they are hard to test. Exceptions: extension method containers or classes to wrap a set of immutable constants
- [x] A class or interface should have a single purpose
  - [x] A class or interface should have a single purpose within the system it functions in. In general, a class either represents a primitive type like an email or ISBN number, an abstraction of some business concept, a plain data structure, or is responsible for orchestrating the interaction between other classes. It is never a combination of those. This rule is widely known as the [Single Responsibility Principle](https://8thlight.com/blog/uncle-bob/2014/05/08/SingleReponsibilityPrinciple.html), one of the S.O.L.I.D. principles
  - [x] Similarly, a property, method or local function should do only one thing
- [x] Consider adding doc-comments to public types and members, but avoid [undocumentation](https://blog.codinghorror.com/avoiding-undocumentation/)
- [x] Public types should protect the consistency of their internal state: validate inputs and use guards like `AssertNotDisposed();`
- [x] Use a method instead of a property when:
  - [x] The work is more expensive than setting a field value
  - [x] It represents a conversion such as the `Object.ToString` method
  - [x] It returns a different result each time it is called, even if the arguments didn’t change. For example, the `NewGuid` method returns a different value each time it is called
  - [x] The operation causes a side effect such as changing some internal state not directly related to the property (which violates the [Command Query Separation](http://martinfowler.com/bliki/CommandQuerySeparation.html) principle). Populating an internal cache or implementing [lazy-loading](http://www.martinfowler.com/eaaCatalog/lazyLoad.html) is a good exception
- [x] Suffix identifier names that represent a `Type`, `MethodInfo`, `PropertyInfo` etc, to distinguish them from instances. For example: rename `MethodInfo connect` to `MethodInfo connectMethod`
- [x] Avoid publicly exposed fields, unless they are `const` or `static readonly`. Convert to (auto-)properties or make them private/internal
- [x] Prefer collection interfaces such as `ICollection<>` and `IDictionary<,>` over arrays, list and dictionaries for inputs and outputs
  - [x] Be hesitant with incoming `IEnumerable<>` parameters: the implementation may change over time, causing the need for multiple enumerations, which may result in undesired side effects
  - [x] Consider to return unchangeable collections, such as `IReadOnlyCollection<>`, `IReadOnlyDictionary<,>` and `IImmutableList<T>`
- [x] Use overloads instead of optional parameters (they have several backwards-compatibility issues, as well as non-intuitive behavior when using interfaces or inheritance). Call the more overloaded method from other overloads
- [x] Properties, arguments and return values representing strings, collections or tasks should never be null. Return an empty collection, an empty string or `Task.CompletedTask`/`Task.FromResult()`
- [x] Don’t use ref or out parameters (return compound objects or tuples), unless the method implements the [TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.int32.tryparse) pattern
- [x] Consider using `ValueTask`/`ValueTask<T>` in high-performance code paths
- [x] Avoid abbreviations
  - [x] Use "management" instead of "mgmt", "options" instead of "opts", etc.
  - [x] Using "config" in the context of ConfigServer is good, but not: `IConfiguration config`
  - [x] Avoid single character variable names, such as i or q. Use index or query instead
  - [x] Use well-known acronyms and abbreviations that are widely accepted or well-known in your work domain. For instance, use acronym UI instead of UserInterface and abbreviation Id instead of Identity
- [x] Remove the "Core" suffix when referring to .NET (ASP.NET Core and Entity Framework Core are okay)
- [x] While you're at it, consider annotation for NRT (see #962 for tips).

Closes #1220.

## Quality checklist


TimHess (approved):
LGTM, fantastic to see all the deduplication possible with the new `HostBuilderWrapper`

PR #1245, opened 2024-01-11
Consul :: Add tests to cover health check path being used when heartbeat is off

### LINKED ISSUE #1062: Consul health check path is not working
## Describe the bug
Consul health check path is not working

## Steps to reproduce 
1. Add `Consul:Discovery:HealthCheckPath` configuration

image

## Expected behavior
Health check endpoint should be added to service instance on consul portal but it only shows the default TTL check.

image

## Environment
 - Steeltoe Version: NuGet package **Steeltoe.Discovery.Consul 3.2.1**
 - Platform: **Local test / Azure**
 - OS: **Windows**
 - .NET Version: **.NET 5**

cieciurm:
## Description

Hi all,

This PR fixes an issue when health check path is not respected when heartbeat is enabled.

This is a very simple attempt to fix the issue, by moving the short circuiting statement couple of lines lower. 
Please let me know if it's too simple ;) 
I've added unit tests to cover the issue scenario.

As a bonus, I've added a unit test to assert an exception being thrown when the specified port is a negative number.

If the solution is approved, then I can submit a similar fix for the v4 version (targeting `main` branch)

Fixes #1062

## Quality checklist

### DIFF
--- a/src/Discovery/src/Consul/Registry/ConsulRegistration.cs
+++ b/src/Discovery/src/Consul/Registry/ConsulRegistration.cs
@@ -284,12 +284,6 @@ public class ConsulRegistration : IConsulRegistration
             check.DeregisterCriticalServiceAfter = DateTimeConversions.ToTimeSpan(options.HealthCheckCriticalTimeout);
         }
 
-        if (options.IsHeartBeatEnabled)
-        {
-            check.TTL = DateTimeConversions.ToTimeSpan(options.Heartbeat.Ttl);
-            return check;
-        }
-
         if (port <= 0)
         {
             throw new ArgumentException("CreateCheck port must be greater than 0");
@@ -303,6 +297,12 @@ public class ConsulRegistration : IConsulRegistration
         {
             var uri = new Uri($"{options.Scheme}://{options.HostName}:{port}{options.HealthCheckPath}");
             check.HTTP = uri.ToString();
+        }
+
+        if (options.IsHeartBeatEnabled)
+        {
+            check.TTL = DateTimeConversions.ToTimeSpan(options.Heartbeat.Ttl);
+            return check;
         }
 
         // check.setHeader(properties.getHealthCheckHeaders());
--- a/src/Discovery/src/Consul/Registry/ConsulRegistration.cs
+++ b/src/Discovery/src/Consul/Registry/ConsulRegistration.cs
@@ -284,7 +284,7 @@ public class ConsulRegistration : IConsulRegistration
             check.DeregisterCriticalServiceAfter = DateTimeConversions.ToTimeSpan(options.HealthCheckCriticalTimeout);
         }
 
-        if (options.IsHeartBeatEnabled && options.HealthCheckUrl == null && options.HealthCheckPath == ConsulDiscoveryOptions.CONSUL_DEFAULT_HEALTH_PATH)
+        if (options.IsHeartBeatEnabled)
         {
             check.TTL = DateTimeConversions.ToTimeSpan(options.Heartbeat.Ttl);
             return check;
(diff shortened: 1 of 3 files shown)


src/Discovery/src/Consul/Registry/ConsulRegistration.cs
```diff
@@ -305,6 +299,12 @@ internal static AgentServiceCheck CreateCheck(int port, ConsulDiscoveryOptions o
             check.HTTP = uri.ToString();
         }
 
+        if (options.IsHeartBeatEnabled)
```
TimHess, line 302: So I looked at this issue a few months ago and did not get to a great resolution (at least in part because I'm not super familiar with how Consul expects these checks to work), but I have a local stash of the code that looks more like this for this line:
```
if (options.IsHeartBeatEnabled && options.HealthCheckUrl == null && options.HealthCheckPath == ConsulDiscoveryOptions.DEFAULT_HEALTH_PATH)
```

In hopes of a quicker turnaround to my questions than research, do you have any insights on how this feature is generally (outside of Steeltoe) expected to be configured? I don't know offhand if there is a generally accepted order of which check to use when, and it seems like there should be a set of rules _somewhere_ if there isn't an explicit setting for what type of check to use

cieciurm: Hi @TimHess,
Thanks for your response!
To be perfectly honest with you, I don't have any insights on the order of checks... 
I noticed the issue and I was keen on providing a fix - but as I wrote the fix was very simple. 

I tried to get some inspiration from spring-cloud library but haven't found anything decisive.
Just docs and code:
- [Docs](https://github.com/spring-cloud/spring-cloud-consul/blob/dfbe776df8c7c3c4645c7c4a74f7e98c9109bb24/docs/modules/ROOT/partials/_configprops.adoc#L35)
- [More docs](https://github.com/spring-cloud/spring-cloud-consul/blob/dfbe776df8c7c3c4645c7c4a74f7e98c9109bb24/docs/modules/ROOT/pages/discovery.adoc#L186)
- [Code](https://github.com/spring-cloud/spring-cloud-consul/blob/dfbe776df8c7c3c4645c7c4a74f7e98c9109bb24/spring-cloud-consul-discovery/src/main/java/org/springframework/cloud/consul/serviceregistry/ConsulAutoRegistration.java#L218)

I've applied the change you mentioned you had stashed - it would do the trick as well.

Wondering then about the test for negative port then... It's only valid when we don't shortcircuit - so it requires to opt-out of the usage of heartbeat.

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
```diff
@@ -284,7 +284,7 @@ internal static AgentServiceCheck CreateCheck(int port, ConsulDiscoveryOptions o
             check.DeregisterCriticalServiceAfter = DateTimeConversions.ToTimeSpan(options.HealthCheckCriticalTimeout);
         }
 
-        if (options.IsHeartBeatEnabled)
+        if (options.IsHeartBeatEnabled && options.HealthCheckUrl == null && options.HealthCheckPath == ConsulDiscoveryOptions.CONSUL_DEFAULT_HEALTH_PATH)
```
cieciurm, line 287: @TimHess 
That's the updated line.

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
TimHess: The port is only relevant for an HTTP check. The TTL check requires the client to send a heartbeat signal to Consul (which records, but does not communicate with that port). The test for negative port seems correct to me

cieciurm: Got it, thanks!

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
cieciurm: Thanks for the short summary and explanation @TimHess!
After reading this destilled summary, indeed it doesn't require a fix.

But perhaps adding two unit tests to cover the health check path behavior in relation to heartbeat being on/off could be helpful in the future.

Also one extra test to verify the validation of a negative port for HTTP check.

Thanks for taking your time to look into this and digesting it out in the public 😃

TimHess (approved):
LGTM

TimHess:
CI failures are unrelated

PR #1253, opened 2024-01-21
#1250 cleaning of package reference's

### LINKED ISSUE #1250: Clean of package references due to framework references
I am wanting to decrease the number of transient packages in my projects and instead focus on using framework references where appropriate.

I would like projects which have a framework reference to only have packages references for those packages not available as part of the framework.

Initial list for 4.0:
- Steeltoe.Security.Authentication.Mtls.csproj
- Steeltoe.Configuration.CloudFoundry.csproj
- Steeltoe.Configuration.ConfigServer.csproj
- Steeltoe.Configuration.Encryption.csproj
- Steeltoe.Configuration.Kubernetes.csproj
- Steeltoe.Configuration.SpringBoot.csproj
- Steeltoe.Management.Endpoint.csproj
- Steeltoe.Logging.DynamicSerilog.csproj
- Steeltoe.Logging.DynamicLogger.csproj
- Steeltoe.Discovery.Client.csproj
- Steeltoe.Management.Task.csproj

I would like a similar spring clean done on the 3.x branch.

N/a

## Additional context
Full list of packages bundled into framework: https://github.com/dotnet/core/blob/main/release-notes/net-package-deprecation.md

Should framework references be added for any other projects?
Can target frameworks be reduced due to addressing package versioning & possibly restore previous frameworks?

thompson-tomo:
## Description

The PR goes about cleaning up the package dependencies such that any project that uses a framework does not have an explicit reference to any library which is being supplied by the framework. At the same time, if a project has a dependency on a project which has a framework reference, that project also gains the framework. This approach improves manageability of the dependencies.

The abstraction packages have been left without a framework refernce.

Fixes #1250 

## Quality checklist

### DIFF
--- a/src/Bootstrap/src/AutoConfiguration/Steeltoe.Bootstrap.AutoConfiguration.csproj
+++ b/src/Bootstrap/src/AutoConfiguration/Steeltoe.Bootstrap.AutoConfiguration.csproj
@@ -10,7 +10,7 @@
   <Import Project="..\..\..\..\shared.props" />
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="$(FoundationalVersion)" />
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
 
   <ItemGroup>
--- a/src/Common/test/Common.Http.Test/Steeltoe.Common.Http.Test.csproj
+++ b/src/Common/test/Common.Http.Test/Steeltoe.Common.Http.Test.csproj
@@ -10,7 +10,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.Extensions.Configuration.Json" Version="$(FoundationalVersion)" />
-    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(FoundationalVersion)" />
+    <FrameworkReference Include="Microsoft.AspNetCore.App" />
   </ItemGroup>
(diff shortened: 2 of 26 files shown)


TimHess:
Hi @thompson-tomo,

Some of these look good (primarily removing packages that aren't needed) but there are concerns with these changes:

1. Is there a size impact when PublishSingleFile is used?
1. We do not want to _force_ a dependency on ASP.NET Core if it isn't needed (eg: console/worker apps don't need it)
   1. Using framework refs in the tests _could_ create situations where the tests include more dependencies than a downstream app might and create an inaccurate test environment

I don't immediately have answers to those and haven't had time yet to develop an understanding of those impacts.

thompson-tomo:
Hi @TimHess 

The only projects which gained a framework reference are those which had a project dependency on another steeltoe project which already the framework reference hence I didn't see an issue in adding it. The test projects were done for consistency but I have no issue with them being switched back if you would like.

Based on the presentation's I have seen & the training I did for a recent new starter, I don't see a file size impact by using publish single file. A difference in sizing arises when changing the publishing method of the parent project to self-contained from framework dependent. This is because framework dependent relies on the runtime from https://dotnet.microsoft.com/en-us/download/dotnet/8.0 being installed.

What I can't remember seeing/testing is if you compile a project using the base framework but it has a dependency on a project with the aspnet framework does it copy the necessary libraries to the output or is it assumed they are on the machine? This shouldn't be too hard to test.

bart-vmware:
Closing this PR in favor of #1327.

PR #1259, opened 2024-02-03
#910 Enable rule S104 and split graph

### LINKED ISSUE #910: Address S104 violations
Address existing violations of [S104: Files should not have too many lines of code](https://rules.sonarsource.com/csharp/RSPEC-104) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

thompson-tomo:
## Description

This will enable rule S104 and in doing so the grpah file has needed to be split. Also to address false positives the impact test projects have the rule disabled as i don't see test scripts being over 1000 line being an issue

Closes #910

## Quality checklist

### DIFF
--- a/Steeltoe.Debug.ruleset
+++ b/Steeltoe.Debug.ruleset
@@ -8,6 +8,7 @@
   </Rules>
   <Rules AnalyzerId="SonarAnalyzer.CSharp" RuleNamespace="SonarAnalyzer.CSharp">
     <Rule Id="S100" Action="Warning" />
+    <Rule Id="S104" Action="Warning" />
     <Rule Id="S106" Action="Warning" />
     <Rule Id="S1067" Action="Warning" />
     <Rule Id="S107" Action="Info" />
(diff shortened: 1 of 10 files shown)


bart-vmware:
The files in src/Management/src/Endpoint/HeapDump/DotNetHeapDump are copies from somewhere else; see the README.md in that directory. Therefore, we intend to leave them as-is. These files are excluded from Resharper cleanup.

_If_ we are going to enable S104, we need to find a way to suppress that rule in only these files. Apparently, `#pragma warning disable` doesn't do the trick. However, the copy was taken quite some time ago and may be outdated. And the reason for making a copy may no longer be justified. Before making any decisions here, we'd like to explore if there are better ways to provide this functionality (without copying in external code). Addressing this is currently not our highest priority, so we don't intend to spend time on it before the first v4 release.

Large files are uncomfortable to navigate/review/update, etc. This applies to tests as well as regular code. In general, tests should meet the same quality criteria as any other code, because they are part of our deliverable.

PR #1262, opened 2024-02-11
[wip] Add sanitizing to refresh endpoint

### LINKED ISSUE #656: Apply added KeysToSanitize after /refresh
## Question
Can we force the new sanitizing key to be applied to the configuration on /refresh call?

 - .NET Core 3.1.3 
 - Steeltoe Version 3.0.2

## Additional context or links
I tried to work with Steeltoe's /env and /refresh endpoints. What I noticed is that the sanitizer works only when a project is started. When you add a key to sanitize, call /refresh and then /env, then I see that the configuration was not sanitized with the added key.

cieciurm:
## Description

Hi all,

I've stumpled upon another `help wanted` issue (#656) and decided to give it a try.
However, the longer I started inspecting the code, the more questions appeared.

So let's try to agree on solution (if there is any) before continuing the code changes.

Here's the result of my investigation:
- At the moment `Sanitizer` is used in `EnvironmentEndpoint` which returns a list of key-value pairs with settings/properties/configuration
- It makes sense to sanitize (so replace with asterisks) some of the values which should not be exposed to the outer world
- The requester of #656 claims that similar treatment should be given to `RefreshEndpoint`
- After futher investigation, `RefreshEndpoint` returns only a list of keys (not values) from the configuration. So there isn't anything to sanitize - unless we'd like to return values as well?
- I've compared `3.x` and `4.x` and it's the same output from the `RefreshEndpoint` (keys only)

So kindly please advise how to approach this issue and whether it's actually an existing issue.

I'm happy to continue with code changes, if we agree what's the desired solution.

cc: @bart-vmware & @TimHess 

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/Refresh/ConfigureRefreshEndpointOptions.cs
+++ b/src/Management/src/Endpoint/Refresh/ConfigureRefreshEndpointOptions.cs
@@ -3,6 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using Microsoft.Extensions.Configuration;
+using Steeltoe.Common;
 using Steeltoe.Management.Endpoint.Options;
 
 namespace Steeltoe.Management.Endpoint.Refresh;
@@ -11,8 +12,32 @@ internal sealed class ConfigureRefreshEndpointOptions : ConfigureEndpointOptions
 {
     private const string ManagementInfoPrefix = "management:endpoints:refresh";
 
+    private static readonly string[] DefaultKeysToSanitize =
+    {
+        "password",
+        "secret",
+        "key",
+        "token",
+        ".*credentials.*",
+        "vcap_services"
+    };
+
     public ConfigureRefreshEndpointOptions(IConfiguration configuration)
         : base(configuration, ManagementInfoPrefix, "refresh")
     {
+    }
+
+    public override void Configure(RefreshEndpointOptions options)
+    {
+        ArgumentGuard.NotNull(options);
+
+        base.Configure(options);
+
+        // It's not possible to distinguish between null and an empty list in configuration.
+        // See https://github.com/dotnet/extensions/issues/1341.
+        if (options.KeysToSanitize.Count == 0)
+        {
+            options.KeysToSanitize = DefaultKeysToSanitize;
+        }
     }
 }
--- a/src/Management/src/Endpoint/Refresh/RefreshEndpointHandler.cs
+++ b/src/Management/src/Endpoint/Refresh/RefreshEndpointHandler.cs
@@ -6,6 +6,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Steeltoe.Common;
+using Steeltoe.Management.Endpoint.Environment;
 
 namespace Steeltoe.Management.Endpoint.Refresh;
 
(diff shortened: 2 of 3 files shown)


TimHess:
Hi @cieciurm,

Thanks for looking into this - the issue is less about what is returned from `/refresh` and more about what happens when the application's configuration is updated. 

Here is the step by step process to recreate what the open issue is talking about:

1. Change the value of `Management:Endpoints:Env:KeysToSanitize` via any config provider that can be updated at runtime
2. `/refresh` is called
3. this is the bug --> Any subsequent request to `/env` should use the now-current values in `KeysToSanitize`

With the revisions made in Steeltoe v4/main, this should already work and `/refresh` _shouldn't_ need to be called explicitly, but given that no PR has linked to the issue, I don't conclusively know if works as expected. 

In Steeltoe v3 and prior, some values that come from configuration were only read at startup and did not use IOptions or any other mechanism for supporting runtime reconfiguration.

cieciurm:
Hi again,

Apologies it took so long to reply. Hopefully my persistence in veryfing, if it works correctly again `main` (`v4.0`) can make it up ;)

Steps to reproduce:

1. Create WebApp 
2. Reference `Steeltoe.Management.Endpoint` and `Steeltoe.Logging.DynamicLogger`
3. Setup `appsettings.json`

```json
  "Management": {
    "Endpoints": {
      "Actuator": {
        "Exposure": {
          "Include": [ "*" ]
        }
      },
      "Env": {
        "KeysToSanitize": [ ]
        //"KeysToSanitize": [ "info:appsetting1" ]
      }
    }
  },
  "info": {
    "appsetting1": "appsetting1",
    "appsetting2": "appsetting2",
    "password": "secret"
  },
```

4. Run the WebApp
5. Go to `/actuator/env` and see:

```json
   {
      "name": "JsonConfigurationProvider: [appsettings.json]",
      "properties": {
        "AllowedHosts": {
          "value": "*"
        },
        "info:appsetting1": {
          "value": "appsetting1"
        },
        "info:appsetting2": {
          "value": "appsetting2"
        },
        "info:password": {
          "value": "******"
        },
```

(`password` is one of the `DefaultKeysToSanitize` to verify if sanitize works at all)

6. Change value to `"KeysToSanitize": [ "info:appsetting1" ]`
7. Call POST `/actuator/refresh` from Postman
8. Go again to `/actuator/env` and still see:

```json
   {
      "name": "JsonConfigurationProvider: [appsettings.json]",
      "properties": {
        "AllowedHosts": {
          "value": "*"
        },
        "info:appsetting1": {
          "value": "appsetting1"
        },
        "info:appsetting2": {
          "value": "appsetting2"
        },
        "info:password": {
          "value": "******"
        },
```

Could this potentially mean that the keys to sanitize doesn't refresh at all even on `v4`?

TimHess:
> Could this potentially mean that the keys to sanitize doesn't refresh at all even on `v4`?

You're right, the keys to sanitize do not refresh. The reason appears to be the list of keys is set when the sanitizer is [initialized](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/src/Endpoint/Environment/EnvironmentEndpointHandler.cs#L35) and never updated, so we would need to either enhance the sanitizer to support a more dynamic list or re-create it if the list changes.

cieciurm:
Hi again,
apologies but due to private obligations (just became a dad), I won't be able to work on this issue.

PR #1272, opened 2024-03-15
Fix/clientcertificatehandler

macsux:
## Description

Fix ClientCertificateHttpHandler breaking when existing cert is not in cert store during rotation. Add ability for it to be used with named options.


macsux:
Ignore this pull request till I fix it please, looks like I haven't properly merged on top of latest

bart-vmware:
@TimHess @macsux Do we need to keep this open? Does #1306 address everything we need?

TimHess:
I do not believe we need this one anymore

PR #1276, opened 2024-04-12
Prevent adding PlaceholderResolver multiple times

bart-vmware:
## Description

Fixed: Prevent adding `PlaceholderResolver` multiple times, which breaks the ENV actuator.

## Quality checklist

### DIFF
--- a/src/Configuration/src/Placeholder/PlaceholderConfigurationExtensions.cs
+++ b/src/Configuration/src/Placeholder/PlaceholderConfigurationExtensions.cs
@@ -48,15 +48,18 @@ public static class PlaceholderConfigurationExtensions
         ArgumentGuard.NotNull(builder);
         ArgumentGuard.NotNull(loggerFactory);
 
-        if (builder is IConfigurationRoot configuration)
+        if (!builder.Sources.Any(source => source is PlaceholderResolverSource))
         {
-            builder.Add(new PlaceholderResolverSource(configuration, loggerFactory));
-        }
-        else
-        {
-            var resolver = new PlaceholderResolverSource(builder.Sources, loggerFactory);
-            builder.Sources.Clear();
-            builder.Add(resolver);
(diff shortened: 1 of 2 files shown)


TimHess (approved):
LGTM as-is, only thought is we could log when it isn't added so the user knows they can remove the extra entry

bart-vmware:
> LGTM as-is, only thought is we could log when it isn't added so the user knows they can remove the extra entry

Oh, I saw your comment after merging. We could, but `AddPlaceholderResolver` is sometimes called by other Steeltoe code: Bootstrap, Config Server, Management. In that case, the user can't fix it. I don't think we should warn in those cases. Also, it's pretty common to run registrations multiple time in ASP.NET. For example, `AddOptions`, `AddLogging` etc are called multiple times, just to make sure the basics are in place.

fennekit:
I think this also applies to the EncryptionResolver (see. [EncryptionConfigurationExtensions.cs](https://github.com/SteeltoeOSS/Steeltoe/blob/7121e8f51223c09fd9271ea2f742d89ebb1b871e/src/Configuration/src/Encryption/EncryptionConfigurationExtensions.cs#L58))

bart-vmware:
@fennekit Thanks, I think you're right.

I've created this PR because we've had internal reports from people explicitly calling `.AddPlaceholderResolver()` and `.AddConfigServer()` (where the latter implicitly calls this method as well), resulting in problems, such as the ENV endpoint not listing all of the configuration providers.

This problem is less likely to occur with `EncryptionResolver` because (as far as I'm aware) there's no other Steeltoe code implicitly calling `AddEncryptionResolver`. However, it doesn't hurt to apply the same safeguard there. Would you be interested in creating a PR for that?

fennekit:
@bart-vmware OK. I will create a PR in the coming week.

PR #1300, opened 2024-05-15
Eureka fixes

bart-vmware:
## Description

Eureka health fixes: 

- Make `IHealthCheckHandler` also read from ASP.NET health checks (requires move into Common)
- `EurekaHealthCheckHandler` is now automatically registered, but off by default in configuration (was: not registered but on by default, so effectively off)
- Fixes in translation from `HealthStatus` to `InstanceStatus`, add more tests
  - When a contributor is turned off, it must be ignored (not result in `InstanceStatus.Unknown`)
  - When a contributor reports warning, translate to `InstanceStatus.Up` (not `Unknown`)
  - When there are no contributors, translate to `InstanceStatus.Up` (not `Unknown`)
  - Ignore the result from `EurekaServerHealthContributor` the first time, as its dependent data are still being determined.
- Fixed a case where the IoC container would create multiple instances of `EurekaDiscoveryClient`.
- Fixed ConfigServer health status when turned off.
- Skip running health checks while the instance is still in `Starting` state.
- Remove warning that the app name changed, when only its casing changed.

Additionally, turned off macOS in cibuilds, because it fails randomly.

## Quality checklist

### DIFF
--- a/src/Common/src/Common/Steeltoe.Common.csproj
+++ b/src/Common/src/Common/Steeltoe.Common.csproj
@@ -12,7 +12,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(FoundationalVersion)" />
-    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(MatchTargetFrameworkVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions" Version="$(MatchTargetFrameworkVersion)" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(FoundationalVersion)" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(FoundationalVersion)" />
--- a/src/Common/src/Common/Steeltoe.Common.csproj
+++ b/src/Common/src/Common/Steeltoe.Common.csproj
@@ -12,6 +12,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(FoundationalVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(MatchTargetFrameworkVersion)" />
     <PackageReference Include="Microsoft.Extensions.Hosting.Abstractions" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.Options.ConfigurationExtensions" Version="$(FoundationalVersion)" />
     <PackageReference Include="System.Diagnostics.DiagnosticSource" Version="$(FoundationalVersion)" />
--- a/src/Discovery/src/Eureka/EurekaDiscoveryClient.cs
+++ b/src/Discovery/src/Eureka/EurekaDiscoveryClient.cs
@@ -31,6 +31,7 @@ public sealed class EurekaDiscoveryClient : IDiscoveryClient
     private readonly Timer? _cacheRefreshTimer;
     private readonly SemaphoreSlim _registerUnregisterAsyncLock = new(1);
     private readonly SemaphoreSlim _registryFetchAsyncLock = new(1);
+    private volatile bool _hasFirstHeartbeatCompleted;
 
     private volatile ApplicationInfoCollection _remoteApps;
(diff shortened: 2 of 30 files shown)


src/Discovery/src/Eureka/EurekaDiscoveryClient.cs
```diff
@@ -412,7 +412,8 @@ internal async Task RunHealthChecksAsync(CancellationToken cancellationToken)
 
                 InstanceInfo snapshot = _appInfoManager.Instance;
 
-                if (aggregatedStatus != snapshot.Status)
+                // When running the first time, EurekaServerHealthContributor returns UNKNOWN. Don't take this app out of service because of that.
+                if (aggregatedStatus != InstanceStatus.Unknown && aggregatedStatus != snapshot.Status)
```
TimHess, line 416: What if the aggregated status becomes `unknown` at some time in the future?

src/Common/src/Common/Steeltoe.Common.csproj
```diff
@@ -12,6 +12,7 @@
     <PackageReference Include="Microsoft.Extensions.Caching.Abstractions" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="$(FoundationalVersion)" />
+    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks" Version="$(MatchTargetFrameworkVersion)" />
```
TimHess, line 15:
```suggestion
    <PackageReference Include="Microsoft.Extensions.Diagnostics.HealthChecks.Abstractions" Version="$(MatchTargetFrameworkVersion)" />
```
If we need to move this reference to Common, let's also scale it back so we only take the Abstractions package

src/Discovery/test/Eureka.Test/EurekaHealthCheckHandlerTest.cs
```diff
@@ -23,10 +25,10 @@ public sealed class EurekaHealthCheckHandlerTest
     [Theory]
     [InlineData(HealthStatus.Down, InstanceStatus.Down)]
     [InlineData(HealthStatus.Up, InstanceStatus.Up)]
-    [InlineData(HealthStatus.Warning, InstanceStatus.Unknown)]
+    [InlineData(HealthStatus.Warning, InstanceStatus.Up)]
     [InlineData(HealthStatus.Unknown, InstanceStatus.Unknown)]
     [InlineData(HealthStatus.OutOfService, InstanceStatus.OutOfService)]
-    [InlineData(null, InstanceStatus.Unknown)]
+    [InlineData(null, InstanceStatus.Up)]
```
TimHess, line 31: I have an initial reaction of disliking this change, as part of me feels like a null health check _should_ result in an Unknown instance status. It looks like you are thinking that for this purpose, Steeltoe should treat `null` as erroneous and ignore it, is that accurate?

TimHess (changes requested):

src/Discovery/test/Eureka.Test/EurekaHealthCheckHandlerTest.cs
bart-vmware: `IHealthContributor` returns `null` to indicate the check is currently turned off. When _all_ registered checks are turned off, the aggregate health status is effectively unavailable and should not affect the instance status. I suppose this should be changed to `[InlineData(null, null)]` and have the caller deal with that case.

`IHealthContributor` returns `Unknown` when the health check can't be performed. Possible causes:
- An unhandled exception during evaluation
- The system is misconfigured
- Some information isn't available (yet) *1)

*1) This is the tricky one. Normally, health checks execute as part of an HTTP request, _after_ the app has fully started. For service discovery, however, it starts running health checks from a background thread as part of `EurekaDiscoveryClient` instantiation (so possibly _before_ the app startup has completed, depending on configured timings).
- `LivenessHealthContributor` and `ReadinessHealthContributor` initialize during `IHostApplicationLifetime.ApplicationStarted`, returning `Unknown` when called before that happened.
- `EurekaServerHealthContributor` outcome depends on:
  - The remote instance status (which hasn't been fetched yet)
  - Last applications fetch status (which hasn't happened yet)
  - Heartbeat status (which is currently executing, so there's no outcome yet).

  The caller being the heartbeat thread, none of these are available, so it returns `Unknown`.

One thing we can do is not run health checks when the instance status is `Starting`. That solves the problem in case `IsInstanceEnabledOnInit` is configured as `false` and the user marks when startup has completed. However, this doesn't solve the default configuration (`false`). In that case, we could run our own hosted service, signaling when it's okay to run health checks. With those in place, we can translate health status `Unknown` to instance status `Unknown`.

What do you think? I'm open to suggestions.

bart-vmware: We've discussed this and landed on the plan to translate health outcome `Unknown` to instance status `Up` (so the load balancer keeps sending traffic to it), but log a warning. However, today I realized that doesn't solve the `EurekaServerHealthContributor` chicken-and-egg problem.

What I ultimately ended up with, is ignoring the _first_ execution of `EurekaServerHealthContributor`. On later executions, an earlier heartbeat must have completed (succeeded or failed).

The `LivenessHealthContributor`/`ReadinessHealthContributor` taking the instance out of rotation when the app hasn't been started yet is a good thing (don't add them if you don't want to postpone traffic to it until fully started), so it doesn't need to be special-cased.

TimHess (approved):
LGTM

PR #1301, opened 2024-05-21
Align discovery doc-comments with online documentation

bart-vmware:
## Description

Update doc-comments based on documentation website, remove "configuration property" comments, document and inline default values, made `EurekaApplicationsHealthContributor` public.

## Quality checklist


src/Discovery/src/Eureka/Configuration/EurekaClientOptions.cs
```diff
@@ -10,91 +10,80 @@ namespace Steeltoe.Discovery.Eureka.Configuration;
...
 
     /// <summary>
-    /// Gets or sets a value indicating whether the Eureka client should disable fetching of deltas and should rather resort to getting the full registry
-    /// information. Note that the delta fetches can reduce the traffic tremendously, because the rate of change with the Eureka server is normally much
-    /// lower than the rate of fetches. The changes are effective at runtime at the next registry fetch cycle as specified by
-    /// <see cref="RegistryFetchIntervalSeconds" />. Configuration property: eureka:client:shouldDisableDelta.
+    /// Gets or sets a value indicating whether to fetch the full registry each time or fetch only deltas. Note that the delta fetches can reduce the traffic
+    /// tremendously, because the rate of change in the Eureka server is normally much lower than the rate of fetches. Default value: false.
     /// </summary>
     [ConfigurationKeyName("ShouldDisableDelta")]
```
TimHess, line 45: These entries that use `ConfigurationKeyName` seem like they might have a much stronger case for keeping the configuration property in the XML comments somewhere. Would it make sense to add them as <remarks>?

src/Discovery/src/Eureka/Configuration/EurekaHealthOptions.cs
```diff
@@ -3,24 +3,28 @@
...
+using Steeltoe.Common.HealthChecks;
 
 namespace Steeltoe.Discovery.Eureka.Configuration;
 
 public sealed class EurekaHealthOptions
 {
     /// <summary>
-    /// Gets or sets a value indicating whether to enable the management health contributor. Configuration property: eureka:client:health:enabled.
+    /// Gets or sets a value indicating whether to activate an <see cref="IHealthContributor" /> that verifies connectivity to the Eureka server. Default
+    /// value: true.
```
TimHess, line 14:
```suggestion
    /// Gets or sets a value indicating whether to activate an <see cref="IHealthContributor" /> that verifies connectivity to the Eureka server.
    /// Default value: true.
```
readability nit

src/Discovery/src/Eureka/Configuration/EurekaClientOptions.cs
bart-vmware: We could, but we haven't done that anywhere else in Steeltoe, as far as I'm aware of. These comments will ultimately surface in the schema for autocompletion in `appsettings.json`, where the adapted key name is already known (our schema generation tool should take `[ConfigurationKeyName]` into account, if it doesn't do so already). When displayed in IntelliSense, the developer is already configuring from code and not `appsettings.json`, so it's not too helpful there. 

Based on that, do you still want them to stay?

src/Discovery/src/Eureka/Configuration/EurekaHealthOptions.cs
bart-vmware: Can't do this, the code cleanup formatter auto-inserts line breaks, based on the line length.
However, VS is smart enough to read through them:
image

src/Discovery/src/Eureka/Configuration/EurekaClientOptions.cs
TimHess: And I suppose this information would only show up when you're configuring the values in code, where you'd _probably_ not care as much about the specific configuration key name

TimHess (approved):

PR #1321, opened 2024-07-11
Remove dependency on IFileProvider in ConfigureCertificateOptions

### LINKED ISSUE #1314: Remove IFileProvider overload on ConfigureCertificateOptions
In #1306, we added an overload on `ConfigureCertificateOptions` that takes an `IFileProvider`. The idea was to make it possible to specify your own `PhysicalFileProvider` whose root directory is above the webroot, and then use relative paths to certificate files in configuration. However, `ConfigureCertificateOptions` doesn't have a dependency on `IFileProvider`, so it doesn't work as expected.

The `IFileProvider` abstraction works against us, because we want to allow accessing certificates that aren't in (a subdirectory of) the web root.

Instead, we shouldn't use `IFileProvider` at all, but directly access the file system. When relative paths occur in configuration, they will be resolved against the current directory, which is what `File.ReadAllText` already does. This requires refactoring `FilePathInOptionsChangeTokenSource` to use the underlying `FileSystemWatcher` directly instead of the `IFileProvider` abstraction, which has a different API to register for changes.

bart-vmware:
## Description

Using `FileSystemWatcher` directly involves adding a lot more code. Instead, I've found the ([poorly designed](https://github.com/dotnet/runtime/issues/48602)) `PhysicalFilesWatcher`, which sits in-between `PhysicalFileProvider` and `FileSystemWatcher`.

It currently doesn't work on macOS, because it requires polling. But to make polling actually work, an internal property must be set. Although that's not hard to implement, I didn't do so because we're not using macOS anymore, and these runtime internals are subject to change.

Fixes #1314.

## Quality checklist

### DIFF
--- a/src/Common/src/Common.Certificates/CertificateServiceCollectionExtensions.cs
+++ b/src/Common/src/Common.Certificates/CertificateServiceCollectionExtensions.cs
@@ -5,7 +5,6 @@
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.DependencyInjection.Extensions;
-using Microsoft.Extensions.FileProviders;
 using Microsoft.Extensions.Options;
 using Steeltoe.Common.Configuration;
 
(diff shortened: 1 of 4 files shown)


TimHess (approved):
LGTM

Issue #1293, opened 2024-05-10
Requests to /cloudfoundryapplication possibly return results for /actuator


TimHess:
## Describe the bug
In CF context, only /health and /info are reported as being enabled

Sample response:
```json
{
    "type": "steeltoe",
    "_links": {
        "info": {
            "href": "https://steeltoe-samples-server.apps.dhaka.cf-app.com/cloudfoundryapplication/info",
            "templated": false
        },
        "health": {
            "href": "https://steeltoe-samples-server.apps.dhaka.cf-app.com/cloudfoundryapplication/health",
            "templated": false
        },
        "self": {
            "href": "https://steeltoe-samples-server.apps.dhaka.cf-app.com/cloudfoundryapplication",
            "templated": false
        }
    }
}
```

## Steps to reproduce 
1. Add Actuators to app
2. Deploy to CF
3. View app in apps manager

## Expected behavior
All endpoints supported on Cloud Foundry should be returned/enabled automatically

## Environment (please complete the following information):
 - Steeltoe Version `main`
 - Platform: Cloud Foundry
 - OS: Linux
 - .NET Version 8.0

TimHess:
Probably related:
```
2024-05-29T13:57:16.458-05:00 [APP/PROC/WEB/0] [OUT] IDX10242: Security token: '[Security Artifact of type 'Microsoft.IdentityModel.JsonWebTokens.JsonWebToken' is hidden. For more details, see https://aka.ms/IdentityModel/SecurityArtifactLogging.]' has a valid signature.
2024-05-29T13:57:16.459-05:00 [APP/PROC/WEB/0] [OUT] IDX10239: Lifetime of the token is valid.
2024-05-29T13:57:16.459-05:00 [APP/PROC/WEB/0] [OUT] IDX10214: Audience validation failed. Audiences: 'cloud_controller, actuator, apps_manager_js'. Did not match: validationParameters.ValidAudience: 'null' or validationParameters.ValidAudiences: 'sampleapi, 4e6f8e34-f42b-440e-a042-f2b13c1d5bed'.
2024-05-29T13:57:16.459-05:00 [APP/PROC/WEB/0] [OUT] info: Microsoft.AspNetCore.Authentication.JwtBearer.JwtBearerHandler[1]
2024-05-29T13:57:16.459-05:00 [APP/PROC/WEB/0] [OUT] Failed to validate the token.
[...]

PR #1327, opened 2024-07-17
Common cleanup (initial steps)

### LINKED ISSUE #1250: Clean of package references due to framework references

bart-vmware:
## Description

Initial cleanup of Common namespace. See commit messages for details.

Fixes #1250.

## Quality checklist


src/Management/src/Endpoint/HostBuilderWrapperExtensions.cs
```diff
@@ -135,7 +122,12 @@ public static void AddServicesActuator(this HostBuilderWrapper wrapper)
 
     public static void AddCloudFoundryActuator(this HostBuilderWrapper wrapper)
     {
-        wrapper.ConfigureServices(services => services.AddCloudFoundryActuator());
+        wrapper.ConfigureServices(services =>
+        {
+            services.AddCloudFoundryActuator();
+            services.AddCloudFoundrySecurity();
```
TimHess, line 128: I'm a little confused on why AddCloudFoundrySecurity is required as a separate call. My recollection was that the security pieces were automatically added with the CloudFoundry Actuator, but based on the changes here it seems like it wasn't all setup correctly before? I wonder if [this ](https://github.com/SteeltoeOSS/Steeltoe/issues/1293) is related

bart-vmware: What I did here is similar to Eureka and ConfigServer: use HttpClientFactory. The original code maintained a lazily-created long-lived `HttpClient` instance, which I replaced with pulling one from the factory each time. But the factory needs to be registered in the IoC container at startup. Since there already was `UseCloudFoundrySecurity` to activate the middleware, I introduced `AddCloudFoundrySecurity` to run these registrations. Because it's a new extension method and a prerequisite for `UseCloudFoundrySecurity`, I added a check to throw when `AddCloudFoundrySecurity` hasn't been called.

I don't sufficiently understand the full feature at play here to be able to test the end-to-end scenario, or determine whether a new extension is justified over calling it from an existing place. But what I do understand is there's an outgoing HTTP request where certificate validation can be turned off through configuration. So I just refactored that. I made `SecurityUtils` injectable, so it responds to configuration changes. The changes in this commit may accidentally fix existing bugs, I have insufficient background knowledge to determine that.

bart-vmware: The proposal is to move the logic inside `AddCloudFoundrySecurity` into `AddCloudFoundryActuator` and not have this separate extension method.

TimHess (approved):
LGTM

PR #1334, opened 2024-08-14
API review of Steeltoe.Common

### LINKED ISSUE #590: Switch towards Options pattern
Current Steeltoe API requires passing in `Configuration` object into many service registration calls. This is not only awkward from a syntax perspective, but it also goes against the intended use of .NET configuration architecture as it relates to Options pattern, and doesn't allow runtime config refreshes to occur.

### LINKED ISSUE #825: Revisit analyzer rules and consider automating compliance
Sonar and StyleCop rules should be revisited for 4.0. Enable desired rules (some have comments to be enabled for Steeltoe 3.0 but weren't addressed in time), disable rules we commonly ignore and revisit priorities as appropriate.

Additionally, formatting-related rule compliance could potentially be automated via `dotnet format` and/or Nuke (@macsux may be useful as a resource)

### LINKED ISSUE #897: Public API Surface Area Review

### LINKED ISSUE #898: Review API Surface area in Steeltoe.Common

### LINKED ISSUE #916: Address S1168: Empty arrays and collections should be returned instead of null
Address existing violations of [Address S1168: Empty arrays and collections should be returned instead of null](https://rules.sonarsource.com/csharp/RSPEC-1168) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

Related to #941 and #946.

### LINKED ISSUE #928: Address SA1401: Fields should be private
Address existing violations of [SA1401: Fields should be private](https://github.com/DotNetAnalyzers/StyleCopAnalyzers/blob/master/documentation/SA1401.md) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

- Make the field private
- Convert the field into an auto-property (consider a get-only auto-property)
- Make the field `const` or `readonly`

Note: less strict variants are [S1104](https://rules.sonarsource.com/csharp/RSPEC-1104) and [S2357](https://rules.sonarsource.com/csharp/RSPEC-2357), which solely warn on public fields. Whereas SA1401 applies to exposed fields (this includes `protected`, so it is more strict).

### LINKED ISSUE #929: Address S2360: Optional parameters should not be used
Address existing violations of [S2360: Optional parameters should not be used](https://rules.sonarsource.com/csharp/RSPEC-2360) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

### LINKED ISSUE #941: Address S3956: "Generic.List" instances should not be part of public APIs

### LINKED ISSUE #946: Address S4004: Collection properties should be readonly
Address existing violations of [S4004: Collection properties should be readonly](https://rules.sonarsource.com/csharp/RSPEC-4004) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

- Suppress the violation in code using `#pragma warning disable/restore`, preceded by a justification comment if not obvious (for example, if JSON serialization is involved)

Related to #916 and #941.

### LINKED ISSUE #947: Address S4023: Interfaces should not be empty
Address existing violations of [S4023: Interfaces should not be empty](https://rules.sonarsource.com/csharp/RSPEC-4023) in the codebase and set severity to `Warning` in `Steeltoe.Debug.ruleset` and `Steeltoe.Release.ruleset`.

### LINKED ISSUE #962: Adopt nullable reference types

### LINKED ISSUE #1257: Static HttpClient should use SocketsHttpHandler with a set  pooled connection lifetime
As a follow-up to #21, any long-lived `HttpClient` should be configured with a pooled connection lifetime so that DNS records are less likely to be stale:

```csharp
var handler = new SocketsHttpHandler
{
    PooledConnectionLifetime = TimeSpan.FromMinutes(15) // Recreate every 15 minutes
};
var sharedClient = new HttpClient(handler);
```

https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#dns-behavior

### LINKED ISSUE #1271: Finish replacement of Configuration.CloudFoundry
Steeltoe.Configuration.CloudFoundry.ServiceBinding provided a new way of handling VCAP_SERVICES, but there is no new implementation for reading VCAP_APPLICATION yet.

Before (or during) updates to reading VCAP_APPLICATION, these other packages should be addressed:

- [ ] Config Server
- [ ] Management
- [ ] Security.Authentication

At the time of writing this issue, Eureka still depends on the old bindings, but I believe changes will be made there during the service discovery api review

bart-vmware:
## Description

Public API review of Steeltoe.Common (first part). See the individual commit messages for details.

Closes #898, closes #897, closes #825, closes #928, closes #916, closes #929, closes #941, closes #946, closes #947, closes #590, closes #962, closes #1257, closes #1271.

## Quality checklist


src/Configuration/src/CloudFoundry/CloudFoundryApplicationOptions.cs
```diff
@@ -7,91 +7,124 @@
 
 namespace Steeltoe.Configuration.CloudFoundry;
 
-public sealed class CloudFoundryApplicationOptions : ApplicationInstanceInfo
+public sealed class CloudFoundryApplicationOptions : IApplicationInstanceInfo
```
TimHess, line 10: Changes in this class make it look sooo much more natural here, nice!

src/Configuration/src/Placeholder/PlaceholderResolverProvider.cs
```diff
@@ -77,13 +76,13 @@ public PlaceholderResolverProvider(IList<IConfigurationProvider> providers, ILog
...
-    public bool TryGet(string key, out string value)
+    public bool TryGet(string key, out string? value)
     {
-        ArgumentGuard.NotNull(key);
+        ArgumentException.ThrowIfNullOrEmpty(key);
         EnsureInitialized();
 
         string? originalValue = Configuration![key];
-        value = PropertyPlaceholderHelper.ResolvePlaceholders(originalValue, Configuration);
+        value = PropertyPlaceholderHelper.ResolvePlaceholders(originalValue, Configuration, NullLogger.Instance);
```
TimHess, line 85: Shouldn't we be passing a logger from the factory in here?

src/Configuration/src/ConfigServer/ConfigServerConfigurationProvider.cs
```diff
@@ -735,14 +734,14 @@ internal bool IsDiscoveryFirstEnabled()
...
+        ArgumentNullException.ThrowIfNull(options);
         ObjectDisposedException.ThrowIf(_httpClientHandler == null, this);
 
         var clientCertificateConfigurer = new ClientCertificateHttpClientHandlerConfigurer(OptionsMonitorWrapper.Create(options.ClientCertificate));
-        clientCertificateConfigurer.Configure(_httpClientHandler);
+        clientCertificateConfigurer.Configure("ConfigServer", _httpClientHandler);
 
         var validateCertificatesHandler = new ValidateCertificatesHttpClientHandlerConfigurer<ConfigServerClientOptions>(OptionsMonitorWrapper.Create(options));
-        validateCertificatesHandler.Configure(_httpClientHandler);
+        validateCertificatesHandler.Configure(string.Empty, _httpClientHandler);
```
TimHess, line 744: Can/should we be using `Options.DefaultName` here? It looks like there is currently a naming conflict, but we could rename the property

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
```diff
@@ -136,65 +125,6 @@ private static string[] CreateTags(ConsulDiscoveryOptions options)
...
-
-        if (string.IsNullOrEmpty(instanceId))
-        {
-            instanceId = Random.Shared.Next().ToString(CultureInfo.InvariantCulture);
-        }
-
-        return $"{appName}:{instanceId}";
-    }
-
-    internal static string NormalizeForConsul(string serviceId)
```
TimHess, line 163: Is this already handled in the client library?

src/Management/src/Endpoint/Info/Contributor/GitInfoContributor.cs
```diff
@@ -28,31 +27,29 @@ public GitInfoContributor(ILogger<GitInfoContributor> logger)
     {
     }
 
-    private GitInfoContributor(string propertiesPath, ILogger<GitInfoContributor> logger)
+    public GitInfoContributor(string propertiesPath, ILogger<GitInfoContributor> logger)
```
TimHess, line 30: I take it there wasn't actually a reasonable way to change the path later or use more than one file to contribute git info? Seems we didn't have any tests at least

TimHess (commented):
Looks really good to me, a few questions, comments and minor changes to make

src/Management/src/Endpoint/ThreadDump/EndpointServiceCollectionExtensions.cs
```diff
@@ -3,30 +3,44 @@
...
-    public static void AddThreadDumpActuator(this IServiceCollection services)
+    /// <returns>
+    /// The incoming <paramref name="services" /> so that additional calls can be chained.
+    /// </returns>
+    public static IServiceCollection AddThreadDumpActuator(this IServiceCollection services)
     {
-        ArgumentGuard.NotNull(services);
-
-        services.AddThreadDumpActuator(MediaTypeVersion.V2);
+        return AddThreadDumpActuator(services, MediaTypeVersion.V2);
```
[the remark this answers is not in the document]
bart-vmware: Good question. I have no idea. Can you open an issue to investigate?

src/Common/src/Logging/UpgradableBootstrapLoggerFactory.cs
```diff
@@ -122,31 +121,37 @@ private static void DefaultConfigure(ILoggingBuilder builder, IConfiguration con
...
-    internal sealed class BootstrapLoggerInst : ILogger
+    private sealed class BootstrapLoggerInstance : ILogger
     {
-        public volatile ILogger Logger;
+        private volatile ILogger _logger;
 
         public string Name { get; }
 
-        public BootstrapLoggerInst(ILogger logger, string name)
+        public BootstrapLoggerInstance(ILogger logger, string name)
```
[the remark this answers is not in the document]
bart-vmware: I know, there are more locations this applies to, as well as other optimizations. But I've held back taking Resharper suggestions to avoid making this PR longer. Reviewing them is on my list of pending work, they will be included in a follow-up PR.

src/Management/src/Endpoint/SpringBootAdminClient/ConfigureSpringBootAdminClientOptions.cs
```diff
@@ -33,11 +33,10 @@ public void Configure(SpringBootAdminClientOptions options)
     {
         _configuration.GetSection(ManagementInfoPrefix).Bind(options);
 
-        // Require base path to be supplied directly, in the configuration, or in the app instance info
-        options.BasePath ??= GetBasePath() ?? _applicationInstanceInfo.Uris?.FirstOrDefault() ??
+        options.BasePath ??= GetBasePath() ??
```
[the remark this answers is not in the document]
bart-vmware: Yes, it's on my list. I've added a comment to explain our plan.

src/Common/src/Common/Configuration/PropertyPlaceHolderHelper.cs
```diff
@@ -2,80 +2,97 @@
...
     /// <param name="logger">
-    /// optional logger.
+    /// Used for internal logging. Pass <see cref="NullLogger.Instance" /> to disable logging.
     /// </param>
     /// <returns>
-    /// the supplied value with the placeholders replaced inline.
+    /// The supplied value, with the placeholders replaced inline.
     /// </returns>
-    public static string ResolvePlaceholders(string property, IConfiguration configuration, ILogger logger = null)
+    public static string? ResolvePlaceholders(string? property, IConfiguration? configuration, ILogger logger)
```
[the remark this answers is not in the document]
bart-vmware: The type is internal, so it shouldn't be a problem.

src/Configuration/src/CloudFoundry/CloudFoundryApplicationOptions.cs
bart-vmware: Totally agree! :)

src/Configuration/src/ConfigServer/ConfigServerConfigurationProvider.cs
bart-vmware: You're right, I mistakenly assumed the error was caused by not having a dependency on the options library. I'll rename the property.

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
bart-vmware: No, it isn't. I stepped into the library sources and found it just sends the properties over the wire, without any content checks.

image

src/Management/src/Endpoint/Info/Contributor/GitInfoContributor.cs
bart-vmware: I'm not sure I get your point. So I'm going to assume you're referring to the removal of the `string propertiesPath` parameter from the `ReadGitPropertiesAsync` method. In that case: that's correct. The [original code](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/src/Endpoint/Info/Contributor/GitInfoContributor.cs#L45) always passed the read-only field value:

```c#
Configuration = await ReadGitPropertiesAsync(_propertiesPath, cancellationToken);
```

I suspect the parameter existed to facilitate writing tests, but it's not actually needed, so I removed it. Note the type itself is internal.

In case you meant to raise something else, can you please clarify?

src/Configuration/src/Placeholder/PlaceholderResolverProvider.cs
bart-vmware: Yes, we should, good catch. This is where it pays off to make the parameter non-optional.

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
TimHess: So then I would _**assume**_ that's why it was here... does the Consul server not actually care about these characters being used (anymore)?

src/Management/src/Endpoint/Info/Contributor/GitInfoContributor.cs
TimHess: Sorry about writing that out poorly, I was rushing a bit on that note. This is an item that's been (very low) on my list to understand a bit better because of how the git info frequently doesn't work out the best from our sample and some other container scenarios, but that is probably unrelated

src/Discovery/src/Consul/Registry/ConsulRegistration.cs
bart-vmware: Perhaps, but it's probably there for a reason. Even if the latest version of Consul is more relaxed, we'd be breaking against old versions for no good reason.

TimHess: Never mind, I missed where this validation moved to, there's no problem here

TimHess (approved):
LGTM

PR #1344, opened 2024-08-21
Ensure that new major versions start at Minor version 0

macsux:
Ensure next major version will be 4.0.0 not 4.0.1


bart-vmware:
@macsux Can you please elaborate on what problem this solves? Looking at the build output, the setting is not respected:
> Successfully created package '/home/vsts/work/1/a/packages/Steeltoe.Common.4.0.385-alpha-g3daa3b9af1.nupkg'.

We have set up a way in Azure DevOps to override the full version, so I think this can be closed.

PR #1350, opened 2024-09-02
Fix breaking change: calling AddCloudFoundry() when running locally results in application name "unknown"

bart-vmware:
## Description

Recent refactorings in #1334 changed the behavior of `services.AddCloudFoundryOptions()` to return "unknown" as the application name when running locally, instead of falling back to local settings. This change was observed in broken samples.

This PR restores the behavior that `AddCloudFoundry()` is a light-up feature: if not running on Cloud Foundry, it behaves the same as if `services.AddApplicationInstanceInfo()` was called.

Additionally, this PR ~adds explicit dependencies to `Steeltoe.Configuration.CloudFoundry` and~ removes the odd properties on `IApplicationInfo` that are specific to Cloud Foundry.

## Quality checklist


src/Discovery/src/Consul/Steeltoe.Discovery.Consul.csproj
```diff
@@ -18,5 +18,6 @@
 
   <ItemGroup>
     <ProjectReference Include="..\..\..\Common\src\Http\Steeltoe.Common.Http.csproj" />
+    <ProjectReference Include="..\..\..\Configuration\src\CloudFoundry\Steeltoe.Configuration.CloudFoundry.csproj" />
```
TimHess, line 21: Can/should this be a private reference so that it isn't a hard dependency?

bart-vmware: I suppose it can, but I'm not a huge fan of all those magical references. And to test it properly, we'd need two test projects: one with and one without the reference. We do that for the dependency from Config Server to Eureka/Consul because they're very different systems. But Cloud Foundry is so integrated already that I believe having an explicit reference makes dependencies just easier to understand. We already have `Platform.IsCloudFoundry` in Common that's being referenced, so the dependencies on Cloud Foundry are quite blurry.

On the other hand, having an explicit reference indirectly pulls in
```xml
<FrameworkReference Include="Microsoft.AspNetCore.App" />
```
so that may be a reason for concern. Should we support using Consul/Eureka from a console app?

I'm not sure what to do here, do you have a strong preference?

bart-vmware: > On the other hand, having an explicit reference indirectly pulls in

Oh, that's not true. The dependency is needed anyway.

TimHess: Agree on blurry lines, maybe it isn't a big deal as long as the things we use everywhere stay extremely lightweight... 
`Should we support using Consul/Eureka from a console app?` - ASP.NET Core should probably _not_ be a requirement in order to discover services. I don't remember that changing, but it does look like we already have the framework ref in Consul/Eureka

bart-vmware: I remember testing Eureka in a console app. While the reference is there, the consuming code uses `serviceProvider.GetService` instead of `GetRequiredService` to cope with that.

TimHess (approved):
LGTM

PR #1353, opened 2024-09-05
Allow to clear default entries in collections in options

### LINKED ISSUE #611: Usage of RefreshActuator with HTTP POST
## Question
I am trying to use RefreshActuator in order to reload configuration in run time
I have followed [this documentation](https://docs.steeltoe.io/api/v3/management/using-endpoints.html#http-access) 
 - .NET Core 3.1
 - Steeltoe Version [e.g. 3.0.1]
  ```
public static IWebHostBuilder CreateWebHostBuilder(string[] args)
        {
            var builder = WebHost.CreateDefaultBuilder(args)                
                
                .AddRefreshActuator()
                .AddConfigServer()
                
                .UseStartup<Startup>();
            return builder;
        }
```

also tried 
```
 public static IWebHostBuilder CreateWebHostBuilder(string[] args)
        {
            var builder = WebHost.CreateDefaultBuilder(args)                
                
                .AddAllActuators()
                .AddConfigServer()
                
                .UseStartup<Startup>();
            return builder;
        }
```

and my appsettings.json is
`"Management": {    
      "Endpoints": {
        "Exposure": {
          "Include": [ "*" ]
        }
      }
    }
  `
and appsettings.Development.json
`"Management": {
    "Endpoints": {
      "Exposure": {
        "Include": [ "*" ]
      }
    }
  }`
but when I call  from postman http://localhost:8080/actuator/refresh it the result is 
image

and in my webapi the new value is not reflected in the configuration but if I check the spring server values it is updated  
`http://localhost:8888/my-values/default` i see new values
I have tried to use `IOptionsMonitor<Logging>`  and also DI the `IConfiguration` both are not updated with new values

### PR DESCRIPTION
**Pull request by bart-vmware**
## Description

As explained at https://github.com/dotnet/extensions/issues/1341, the configuration binder is unable to "see" empty collections in options. This poses a problem when the intent is to clear a collection instead of adding default entries. As a workaround, we allow a single-element empty string to indicate _not_ to add defaults. For example, the following `appsettings.json`:
```json
{
  "Management": {
    "Endpoints": {
      "Env": {
        "keysToSanitize": {
          ""
        }
      }
    }
  }
}
```
will overwrite the default of:
```json
{
  "Management": {
    "Endpoints": {
      "Env": {
        "keysToSanitize": {
          "password",
          "secret",
          ...
        }
      }
    }
  }
}
```

This PR updates all cases where a collection is exposed in options that is non-empty by default.

Fixes #611.

## Quality checklist

### DIFF
--- a/src/Management/src/Abstractions/Configuration/EndpointOptions.cs
+++ b/src/Management/src/Abstractions/Configuration/EndpointOptions.cs
@@ -45,14 +45,11 @@ public abstract class EndpointOptions
     /// <summary>
     /// Gets the list of HTTP verbs that are allowed for this endpoint.
     /// </summary>
-    public IList<string> AllowedVerbs { get; }
+    public IList<string> AllowedVerbs { get; private set; } = new List<string>();
 
-    protected EndpointOptions()
+    internal void ApplyDefaultAllowedVerbs()
     {
-        // ReSharper disable once VirtualMemberCallInConstructor
-#pragma warning disable S1699 // Constructors should only call non-overridable methods
         AllowedVerbs = GetDefaultAllowedVerbs();
-#pragma warning restore S1699 // Constructors should only call non-overridable methods
     }
 
     /// <summary>
--- a/src/Management/src/Abstractions/Configuration/EndpointOptions.cs
+++ b/src/Management/src/Abstractions/Configuration/EndpointOptions.cs
@@ -47,6 +47,13 @@ public abstract class EndpointOptions
     /// </summary>
     public IList<string> AllowedVerbs { get; private set; } = new List<string>();
 
+    internal HashSet<string> GetSafeAllowedVerbs()
+    {
+        // Caution: Mapping with an empty string in the list results in exposing the endpoint at ALL verbs.
+        // And duplicate verbs that only differ in case result in an ambiguous match error when mapping routes.
+        return AllowedVerbs.Where(verb => verb.Length > 0).ToHashSet(StringComparer.OrdinalIgnoreCase);
+    }
+
     internal void ApplyDefaultAllowedVerbs()
     {
         AllowedVerbs = GetDefaultAllowedVerbs();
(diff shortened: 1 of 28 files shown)


src/Management/src/Abstractions/Configuration/EndpointOptions.cs
```diff
@@ -45,14 +45,18 @@ public virtual string? Path
     /// <summary>
     /// Gets the list of HTTP verbs that are allowed for this endpoint.
     /// </summary>
-    public IList<string> AllowedVerbs { get; }
+    public IList<string> AllowedVerbs { get; private set; } = new List<string>();
 
-    protected EndpointOptions()
+    internal HashSet<string> GetSafeAllowedVerbs()
+    {
+        // Caution: Mapping with an empty string in the list results in exposing the endpoint at ALL verbs.
```
TimHess, line 52: There's a test below that finds NOT FOUND for GET and POST after setting `allowedVerbs` to be an empty string, is this comment correct?

bart-vmware: Not sure what you mean by that, can you clarify?

TimHess: This comment says  `Caution: Mapping with an empty string in the list results in exposing the endpoint at ALL verbs.`
But the test `...Actuators.Refresh.HttpVerbInConventionalRoutingTest.Can_be_configured_to_allow_no_verbs` uses an empty string for `AllowedVerbs` and then asserts the response is NotFound

bart-vmware: `GetSafeAllowedVerbs` is called by the mapper, so that the test works. A user configures an empty string, which must result in not exposing the endpoints, instead of exposing it on all verbs.

TimHess: Ohhh. I was misunderstanding the purpose of this comment. It's explaining why we have `.Where(verb => verb.Length > 0)`, not describing bad ideas for configuration

TimHess (approved):
LGTM

src/Management/src/Abstractions/Configuration/EndpointOptions.cs
bart-vmware: The comment explains _why_ this method does what it does.

PR #1355, opened 2024-09-12
Merge ServiceBinding project into Configuration.CloudFoundry

bart-vmware:
## Description

This PR merges the `Steeltoe.Configuration.CloudFoundry.ServiceBinding` contents into the `Steeltoe.Configuration.CloudFoundry` project/package. Furthermore, it renames the `*.ServiceBinding` namespace to `*.ServiceBindings` (pluralized). For Kubernetes, the separate project remains, but it is also renamed to plural form, because it implements the [Service Bindings for Kubernetes specification](https://github.com/servicebinding/spec?tab=readme-ov-file#status).

Additionally, this PR reverts the dynamic loading of the `Steeltoe.Configuration.CloudFoundry` dependency in `Steeltoe.Discovery.Eureka`, which was introduced in #1350. The rationale is that Cloud Foundry is the primary reason Steeltoe exists and the dependency is very lightweight. What hasn't changed is that users still need to call both `AddCloudFoundry()` and `AddCloudFoundryServiceBindings()` where appropriate, because the latter is not lightweight enough to automatically include it when `AddCloudFoundry()` is called.

## Quality checklist


TimHess (approved):
LGTM

PR #1356, opened 2024-09-12
Trace Actuator Updates

### LINKED ISSUE #1351: Authorization Header is shown by Actuator trace endpoint
Hello

The actuator's trace endpoint shows the Authorization request header. Everybody who can access this endpoint can view basic authentication credentials or can stole bearer token.

In our case, system administrators and developers have access to this endpoint. So if one of this people is malicious he/she can usurp any user's identity.

Is there any way to prevent this?

Regards.

TimHess:
## Description

While adding support for header filtering, I discovered we're a bit out of sync with Spring on this actuator in general, so I also decided to:
- Remove everything related to the /trace (v1) actuator
- Rename httptrace to httpexchanges (to align with Spring Boot 3), but still use /httptrace on TPCF

I also ran into an unhandled exception with the mappings actuator and included a fix for that

Fixes #1351 

## Quality checklist


src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,202 @@
...
+    {
+        ArgumentNullException.ThrowIfNull(optionsMonitor);
+
+        _optionsMonitor = optionsMonitor;
+        _logger = loggerFactory.CreateLogger<HttpExchangesDiagnosticObserver>();
+    }
+
+    public HttpExchangesResult GetHttpExchanges()
+    {
+        return new HttpExchangesResult(Queue.ToList());
```
bart-vmware, line 40: Redaction should take place at the time the actuator executes. So that users can turn it off in configuration without restarting the app, and then see what they were looking for in the captured exchanges. Otherwise, they'll need to replay the scenario first.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,202 @@
...
+
+    private void RecordHttpTrace(Activity current, HttpContext context)
+    {
+        ArgumentNullException.ThrowIfNull(current);
+        ArgumentNullException.ThrowIfNull(context);
+
+        HttpExchange trace = GetHttpExchange(context, current.Duration);
+        Queue.Enqueue(trace);
+
+        if (Queue.Count > _optionsMonitor.CurrentValue.Capacity && !Queue.TryDequeue(out _))
```
bart-vmware, line 160: Why would dequeue fail when an entry was just added?

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,202 @@
...
+            result.Add(pair.Key.ToLowerInvariant(), HeaderShouldBeRedacted(pair.Key, allowedHeaders) ? new StringValues(Redacted) : pair.Value);
+#pragma warning restore S4040 // Strings should be normalized to uppercase
+        }
+
+        return result;
+    }
+
+    private static bool HeaderShouldBeRedacted(string currentHeader, IList<string> allowedHeaders)
+    {
+        return !allowedHeaders.Contains(currentHeader) || (allowedHeaders.Count == 1 && string.IsNullOrEmpty(allowedHeaders[0]));
```
bart-vmware, line 149: Is the right-hand side of the `||` expression reachable? Because the list gets cleared from `ConfigureOptions`.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,202 @@
...
+    internal static HttpContext? GetHttpContextPropertyValue(object instance)
+    {
+        return GetPropertyOrDefault<HttpContext>(instance, "HttpContext");
+    }
+
+    internal static Dictionary<string, IList<string?>> GetHeaders(IHeaderDictionary headers, IList<string> allowedHeaders)
+    {
+        var result = new Dictionary<string, IList<string?>>();
+
+        foreach (KeyValuePair<string, StringValues> pair in headers)
```
bart-vmware, line 137: Can be deconstructed into named variables with Resharper (note it has a bug that it makes every variable nullable; needs correction afterwards).

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,202 @@
...
+
+        return result;
+    }
+
+    private static bool HeaderShouldBeRedacted(string currentHeader, IList<string> allowedHeaders)
+    {
+        return !allowedHeaders.Contains(currentHeader) || (allowedHeaders.Count == 1 && string.IsNullOrEmpty(allowedHeaders[0]));
+    }
+
+    private void RecordHttpTrace(Activity current, HttpContext context)
```
bart-vmware, line 152: Should this be named exchange too?

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesEndpointHandler.cs
```diff
@@ -0,0 +1,39 @@
...
+
+namespace Steeltoe.Management.Endpoint.Actuators.HttpExchanges;
+
+internal sealed class HttpExchangesEndpointHandler : IHttpExchangesEndpointHandler
+{
+    private readonly IOptionsMonitor<HttpExchangesEndpointOptions> _optionsMonitor;
+    private readonly IHttpExchangesRepository _httpExchangeRepository;
+    private readonly ILogger<HttpExchangesEndpointHandler> _logger;
+
+    public MediaTypeVersion Version { get; set; } = MediaTypeVersion.V2;
```
bart-vmware, line 17: Is this still needed and is it still configurable?

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesEndpointOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+    /// <para>
+    /// If a response header is not present in the <see cref="ResponseHeaders" />, the header name will be logged with a redacted value.
+    /// </para>
+    /// </summary>
+    public bool AddResponseHeaders { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether the request path should be included in traces.
+    /// </summary>
+    public bool AddPathInfo { get; set; } = true;
```
bart-vmware, line 37: Based on the descriptions, names such as `IncludePathInfo` seem more appropriate. Unless this is for spring-compatibility.

src/Management/src/Endpoint/ConfigurationSchema.json
```diff
@@ -418,13 +421,27 @@
                   "type": "string",
                   "description": "Gets or sets the relative path at which this endpoint is exposed."
                 },
+                "RequestHeaders": {
+                  "type": "array",
+                  "items": {
+                    "type": "string"
+                  },
+                  "description": "Gets request header values that are allowed to be logged.\n\nIf a request header is not present in the , the header name will be logged with a redacted value. Request headers can contain authentication tokens, or private information which may have regulatory concerns under GDPR and other laws. Arbitrary request headers should not be logged unless logs are secure and access controlled and the privacy impact assessed."
```
bart-vmware, line 429: > If a request header is not present in the , the header

Looks like the `cref` doesn't properly resolve. Same issue in `ResponseHeaders` below.

src/Management/src/Endpoint/EndpointOptionsExtensions.cs
```diff
@@ -62,7 +62,14 @@ public static string GetPathMatchPattern(this EndpointOptions endpointOptions, M
             path += '/';
         }
 
-        path += endpointOptions.Path;
+        if (baseRequestPath == ConfigureManagementOptions.DefaultCloudFoundryPath && endpointOptions.Id == "httpexchanges")
```
bart-vmware, line 65: Can this reuse `PermissionsProvider.IsCloudFoundryRequest`?

src/Management/src/Endpoint/ManagementWebHostBuilderExtensions.cs
```diff
@@ -238,37 +238,20 @@ public static IWebHostBuilder AddThreadDumpActuator(this IWebHostBuilder builder
...
-    /// The <see cref="IWebHostBuilder" /> to configure.
-    /// </param>
-    /// <param name="mediaTypeVersion">
-    /// Specify the media type version to use in the response.
-    /// </param>
-    /// <returns>
-    /// The incoming <paramref name="builder" /> so that additional calls can be chained.
-    /// </returns>
-    public static IWebHostBuilder AddTraceActuator(this IWebHostBuilder builder, MediaTypeVersion mediaTypeVersion)
+    public static IWebHostBuilder AddHttpExchangesActuator(this IWebHostBuilder builder)
```
bart-vmware, line 249: I see both the names `HttpExchange` and `HttpExchanges` here. What's the correct terminology?

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/EndpointMiddlewareTest.cs
```diff
@@ -0,0 +1,54 @@
...
+        builder.Configuration.AddInMemoryCollection(AppSettings);
+        builder.AddHttpExchangesActuator();
+        await using WebApplication host = builder.Build();
+        host.UseRouting();
+        await host.StartAsync();
+        HttpClient httpClient = host.GetTestClient();
+        HttpResponseMessage response = await httpClient.GetAsync(new Uri("http://localhost/actuator/httpexchanges"));
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        string json = await response.Content.ReadAsStringAsync();
+        json.Should().Be("{\"traces\":[]}");
```
bart-vmware, line 37: Added some line breaks for clarity.
```suggestion
        WebApplicationBuilder builder = TestWebApplicationBuilderFactory.Create();
        builder.Configuration.AddInMemoryCollection(AppSettings);
        builder.AddHttpExchangesActuator();
        await using WebApplication host = builder.Build();

        host.UseRouting();
        await host.StartAsync();
        HttpClient httpClient = host.GetTestClient();

        HttpResponseMessage response = await httpClient.GetAsync(new Uri("http://localhost/actuator/httpexchanges"));
        response.StatusCode.Should().Be(HttpStatusCode.OK);

        string json = await response.Content.ReadAsStringAsync();
        json.Should().Be("{\"traces\":[]}");
```

bart-vmware (commented):
First glance over the changes. It's pretty hard to see what's changed on GitHub due to renames. I'll run a file-by-file diff in the next round.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
TimHess: Nope. This needs to be reworked and apparently needs at least one more test

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
TimHess: I'm not sure, I didn't change this code

bart-vmware: This happens when `options.Capacity` is set to a negative value. The configurer should prevent that from happening, and then this check can be removed.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/EndpointMiddlewareTest.cs
```diff
@@ -85,6 +93,10 @@ public async Task CloudFoundryEndpointMiddleware_ServiceContractNotBroken()
                   "href": "http://localhost/cloudfoundryapplication/info",
                   "templated": false
                 },
+                "httptrace": {
+                  "href": "http://localhost/cloudfoundryapplication/httptrace",
+                  "templated": false
+                },
```
bart-vmware, line 99: Was this added intentionally? It looks like it doesn't belong here, but perhaps I misunderstood.

src/Management/test/Endpoint.Test/ManagementWebApplicationBuilderExtensionsTest.cs
```diff
@@ -292,19 +292,20 @@ public async Task AddThreadDumpActuator_WebApplicationBuilder_IStartupFilterFire
...
         host.UseRouting();
         await host.StartAsync();
 
-        Assert.Single(host.Services.GetServices<IHttpTraceEndpointHandler>());
-        Assert.Single(host.Services.GetServices<IStartupFilter>().Where(filter => filter is AllActuatorsStartupFilter));
-        HttpResponseMessage response = await host.GetTestClient().GetAsync(new Uri("/actuator/httptrace", UriKind.Relative));
-        Assert.Equal(HttpStatusCode.OK, response.StatusCode);
+        host.Services.GetServices<IHttpExchangesEndpointHandler>().Should().ContainSingle();
+        host.Services.GetServices<IStartupFilter>().Where(filter => filter is AllActuatorsStartupFilter).Should().ContainSingle();
+        HttpClient httpClient = host.GetTestClient();
```
bart-vmware, line 306:
```suggestion
        using HttpClient httpClient = host.GetTestClient();
```

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,217 @@
...
+        foreach (HttpExchange exchange in recentExchanges)
+        {
+            FilterHeaders(exchange.Request.Headers, _optionsMonitor.CurrentValue.RequestHeaders);
+            FilterHeaders(exchange.Response.Headers, _optionsMonitor.CurrentValue.ResponseHeaders);
+        }
+
+        return new HttpExchangesResult(recentExchanges);
+    }
+
+    public override void ProcessEvent(string eventName, object? value)
```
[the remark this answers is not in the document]
bart-vmware: I think we should drop `HttpClientDesktopObserver` (which reads from `WebRequest`) entirely, because it was [obsoleted in .NET 6](https://learn.microsoft.com/en-us/dotnet/core/compatibility/networking/6.0/webrequest-deprecated).

Then rename the renaming one to just `HttpClientObserver`.

src/Management/src/Endpoint/ConfigurationSchema.json
bart-vmware: Fixed by https://github.com/SteeltoeOSS/Steeltoe/pull/1358.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: `value` is an anonymous type with a `HttpContext` as a property... I can't seem to get type checking to work in this scenario and haven't found anything in several web searches.

You're probably right about the other observers, I'm not sure yet if it's something I want to include in this PR though

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,217 @@
...
+    }
+
+    internal static Dictionary<string, IList<string>> GetHeaders(IHeaderDictionary headers)
+    {
+        var result = new Dictionary<string, IList<string>>();
+
+        foreach ((string key, StringValues value) in headers)
+        {
+#pragma warning disable S4040 // Strings should be normalized to uppercase
+            result.Add(key.ToLowerInvariant(), value);
```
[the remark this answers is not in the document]
TimHess: I'm assuming the all-lowercase was to make it look like some old Spring version. AFAICT there's no reason to be all lowercase anymore

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/EndpointMiddlewareTest.cs
TimHess: This was added as a cf-specific integration test, it's irrelevant now

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
TimHess: This line is how the queue is limited to `options.Capacity` - I think we need everything here except the log statement... but is it worth removing? It _shouldn't_ ever execute, but it might be interesting if it does

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesEndpointOptions.cs
```diff
@@ -0,0 +1,81 @@
...
+
+    /// <summary>
+    /// Gets request header values that are allowed to be logged.
+    /// <para>
+    /// If a request header is not present in the <see cref="HttpExchangesEndpointOptions.RequestHeaders" />, the header name will be logged with a redacted value. Request headers can
+    /// contain authentication tokens, or private information which may have regulatory concerns under GDPR and other laws. Arbitrary request headers should
+    /// not be logged unless logs are secure and access controlled and the privacy impact assessed.
+    /// </para>
+    /// </summary>
+    public IList<string> RequestHeaders { get; private set; } = new List<string>();
```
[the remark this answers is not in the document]
TimHess: All of the header keys will always be included in the `HttpExchange`. This is a list of headers that are not redacted/filtered/sanitized... If we want it to be more descriptive (this came from [ASP.NET Core's HttpLogging](https://github.com/dotnet/aspnetcore/blob/main/src/Middleware/HttpLogging/src/HttpLoggingOptions.cs#L30)), maybe `SafeRequestHeaders`?

src/Management/src/Endpoint/Actuators/HttpExchanges/ServiceCollectionExtensions.cs
```diff
@@ -6,41 +6,34 @@
...
-    /// <param name="version">
-    /// The media version to use.
+    /// <param name="configureOptions">
+    /// A delegate to configure the <see cref="HttpExchangesEndpointOptions" />.
     /// </param>
     /// <returns>
     /// The incoming <paramref name="services" /> so that additional calls can be chained.
     /// </returns>
-    public static IServiceCollection AddTraceActuatorServices(this IServiceCollection services, MediaTypeVersion version)
+    public static IServiceCollection AddHttpExchangesActuatorServices(this IServiceCollection services, Action<HttpExchangesEndpointOptions> configureOptions)
```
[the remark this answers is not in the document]
TimHess: I added when looking at ASP.NET Core's HttpLogging, which requires you to use a mechanism like this to set which headers should be logged. If we are happy with how Steeltoe supports binding from configuration, then I could pull this back out

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: 😞 I looked at this situation again ... spot why I'm sad about this in the response from a Spring app:
```json
{
    "exchanges": [
        {
            "timestamp": "2024-09-18T20:38:11.820657700Z",
            "request": {
                "uri": "http://localhost:8080/actuator/httpexchanges",
                "method": "GET",
                "headers": {
                    "host": [
                        "localhost:8080"
                    ],
                    "connection": [
                        "keep-alive"
                    ],
                    "pragma": [
                        "no-cache"
                    ],
                    "cache-control": [
                        "no-cache"
                    ],
                    "sec-ch-ua": [
                        "\"Chromium\";v=\"128\", \"Not;A=Brand\";v=\"24\", \"Microsoft Edge\";v=\"128\""
                    ],
                    "sec-ch-ua-mobile": [
                        "?0"
                    ],
                    "sec-ch-ua-platform": [
                        "\"Windows\""
                    ],
                    "upgrade-insecure-requests": [
                        "1"
                    ],
                    "user-agent": [
                        "Mozilla/5.0 (Windows NT 10.0; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/128.0.0.0 Safari/537.36 Edg/128.0.0.0"
                    ],
                    "accept": [
                        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7"
                    ],
                    "sec-fetch-site": [
                        "none"
                    ],
                    "sec-fetch-mode": [
                        "navigate"
                    ],
                    "sec-fetch-user": [
                        "?1"
                    ],
                    "sec-fetch-dest": [
                        "document"
                    ],
                    "accept-encoding": [
                        "gzip, deflate, br, zstd"
                    ],
                    "accept-language": [
                        "en-US,en;q=0.9"
                    ]
                }
            },
            "response": {
                "status": 200,
                "headers": {
                    "Content-Type": [
                        "application/vnd.spring-boot.actuator.v3+json"
                    ],
                    "Transfer-Encoding": [
                        "chunked"
                    ],
                    "Date": [
                        "Wed, 18 Sep 2024 20:38:11 GMT"
                    ],
                    "Keep-Alive": [
                        "timeout=60"
                    ],
                    "Connection": [
                        "keep-alive"
                    ]
                }
            },
            "timeTaken": "PT0.0014794S"
        }
```

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesResult.cs
```diff
@@ -0,0 +1,23 @@
...
+
+    public IList<HttpExchange> Exchanges { get; }
+
+    public HttpExchangesResult(IList<HttpExchange> exchanges)
+    {
+        ArgumentNullException.ThrowIfNull(exchanges);
+        ArgumentGuard.ElementsNotNull(exchanges);
+
+        Exchanges = exchanges;
+        CurrentVersion = MediaTypeVersion.V2;
```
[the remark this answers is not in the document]
TimHess: Hm... I don't think this was working at all, it looks like [responses will always be v2 unless something else is specified in a request header] (https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/src/Endpoint/ContentNegotiation/ContentNegotiationExtensions.cs#L28) 🫤

src/Management/src/Endpoint/Actuators/RouteMappings/RouteMappingsEndpointHandler.cs
```diff
@@ -68,19 +68,36 @@ private Dictionary<string, IList<RouteMappingDescription>> GetMappingDescription
 
         foreach (ApiDescription description in apiContext.Results)
         {
-            var descriptor = (ControllerActionDescriptor)description.ActionDescriptor;
-            AspNetCoreRouteDetails details = GetRouteDetails(description);
-            string controllerTypeName = descriptor.ControllerTypeInfo.FullName!;
-            mappingDescriptions.TryGetValue(controllerTypeName, out IList<RouteMappingDescription>? descriptions);
-
-            if (descriptions == null)
+            if (description.ActionDescriptor is ControllerActionDescriptor descriptor)
```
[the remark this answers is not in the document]
TimHess: I hit an issue here with a minimal api app that was new from the template... which registers like this:
```
app.MapGet("/weatherforecast", () =>
{
    WeatherForecast[] forecast = Enumerable.Range(1, 5).Select(index => new WeatherForecast(DateOnly.FromDateTime(DateTime.Now.AddDays(index)),
        Random.Shared.Next(-20, 55), summaries[Random.Shared.Next(summaries.Length)])).ToArray();

    return forecast;
}).WithName("GetWeatherForecast").WithOpenApi();
```

src/Management/src/Endpoint/ConfigurationSchema.json
TimHess: Rebased on that branch

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesEndpointOptions.cs
```diff
@@ -0,0 +1,87 @@
...
+
+public sealed class HttpExchangesEndpointOptions : EndpointOptions
+{
+    /// <summary>
+    /// Gets or sets a value indicating how many traces should be stored. Default value is 100.
+    /// </summary>
+    public int Capacity { get; set; } = -1;
+
+    /// <summary>
+    /// Gets or sets a value indicating whether HTTP headers from the request should be included in traces.
```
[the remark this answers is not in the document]
bart-vmware: The default value for `IncludeRemoteAddress` is still missing.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
bart-vmware: This is unreachable code now, so I don't see why we should preserve that. It's just raising questions on why it exists when examined. Suggested change:
```c#
if (Queue.Count > _optionsMonitor.CurrentValue.Capacity)
{
    Queue.TryDequeue(out _);
}
```

Multiple threads may be running this code concurrently, so there's a potential window where the queue gets slightly larger than the maximum, for a very short time, but that shouldn't be a problem.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,216 @@
...
+                filteredHeaders.Add(key, value);
+            }
+        }
+
+        return filteredHeaders;
+    }
+
+    private static bool HeaderShouldBeRedacted(string currentHeader, IList<string> allowedHeaders)
+    {
+        return allowedHeaders is not [""] && !allowedHeaders.Any(header => header.Equals(currentHeader, StringComparison.OrdinalIgnoreCase));
```
bart-vmware, line 181:
```suggestion
        return !allowedHeaders.Any(header => header.Equals(currentHeader, StringComparison.OrdinalIgnoreCase));
```

src/Management/src/Endpoint/EndpointOptionsExtensions.cs
bart-vmware: I do not see any more references to "httptrace" in the codebase. What's the intended endpoint path when running on Cloud Foundry?

src/Management/src/Endpoint/Actuators/HttpExchanges/ConfigureHttpExchangesEndpointOptions.cs
```diff
@@ -81,28 +82,26 @@ public override void Configure(HttpExchangesEndpointOptions options)
         // It's not possible to distinguish between null and an empty list in configuration.
         // See https://github.com/dotnet/extensions/issues/1341.
         // As a workaround, we interpret a single empty string element to clear the defaults.
-        if (options.RequestHeaders.Count == 0)
+        if (options.RequestHeaders.Count == 0 || !string.IsNullOrEmpty(options.RequestHeaders[0]))
```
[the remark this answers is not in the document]
bart-vmware: We discussed this and decided that anything in configuration should be additive to the default whitelist. Note this makes it _impossible_ to exclude entries from the default whitelist _from configuration_. Although unlikely to ever be needed, it can be accomplished by registering a (post)configurer.

src/Management/src/Endpoint/Actuators/HttpExchanges/ConfigureHttpExchangesEndpointOptions.cs
```diff
@@ -81,28 +82,26 @@ public override void Configure(HttpExchangesEndpointOptions options)
...
                 options.ResponseHeaders.Add(defaultKey);
             }
         }
-        else if (options.ResponseHeaders is [""])
+
+        // this should only be needed until Apps Manager supports /httpexchanges
+        if (Platform.IsCloudFoundry)
         {
-            options.ResponseHeaders.Clear();
+            options.Id = "httptrace";
```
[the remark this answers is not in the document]
bart-vmware: How does this decision impact Cloud Foundry and Spring Boot Admin?

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        if (Queue.Count > _optionsMonitor.CurrentValue.Capacity && !Queue.TryDequeue(out _))
+        {
+            _logger.LogDebug("Stop - Dequeue failed");
+        }
+    }
+
+    private static Uri FilterUri(Uri requestUri, HttpExchangesEndpointOptions options)
+    {
+        var filteredUri = new UriBuilder(requestUri);
```
bart-vmware, line 106: I think we should redact user/password here. See `UriExtensions.ToMaskedUri`.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,216 @@
...
+        foreach ((string key, StringValues value) in headers)
+        {
+            if (HeaderShouldBeRedacted(key, allowedHeaders))
+            {
+                filteredHeaders[key] = Redacted;
+            }
+            else
+            {
+                filteredHeaders.Add(key, value);
+            }
```
bart-vmware, line 173:
```suggestion
            filteredHeaders[key] = HeaderShouldBeRedacted(key, allowedHeaders) ? Redacted : value;
```

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchange.cs
```diff
@@ -0,0 +1,37 @@
...
+namespace Steeltoe.Management.Endpoint.Actuators.HttpExchanges;
+
+public sealed class HttpExchange
+{
+    public DateTime Timestamp { get; }
+    public HttpExchangePrincipal? Principal { get; }
+    public HttpExchangeSession? Session { get; }
+    public HttpExchangeRequest Request { get; }
+    public HttpExchangeResponse Response { get; }
+    public string TimeTaken { get; }
```
bart-vmware, line 16: I'd prefer if this were a `TimeSpan`, so it can be sorted on or used in custom filter logic. The formatting should be achieved by using `[JsonIgnore]` and adding an extra property with `[JsonPropertyName]` that calls `XmlConvert.ToString`. Then we don't need two constructors anymore.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,216 @@
...
+            let remoteAddress = options.IncludeRemoteAddress ? exchange.Request.RemoteAddress : null
+            let session = options.IncludeSessionId ? exchange.Session : null
+            let request = new HttpExchangeRequest(exchange.Request.Method, requestUri, filteredRequestHeaders, remoteAddress)
+            let response = new HttpExchangeResponse(exchange.Response.Status, filteredResponseHeaders)
+            select new HttpExchange(request, response, exchange.Timestamp, principal, session, exchange.TimeTaken));
+
+        if (options.Reverse)
+        {
+            recentExchanges = recentExchanges.OrderByDescending(exchange => exchange.Timestamp).ToList();
+        }
```
bart-vmware, line 58:
```suggestion
        recentExchanges.Sort((first, second) => options.Reverse ? second.Timestamp.CompareTo(first.Timestamp) : first.Timestamp.CompareTo(second.Timestamp));
```

I think we should always sort on timestamp. Multiple threads can add to the queue concurrently, so there's no strong guarantee the queue is ordered by timestamp. Also, using LINQ here copies the list, which is expensive.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,216 @@
...
+            let principal = options.IncludeUserPrincipal ? exchange.Principal : null
+            let remoteAddress = options.IncludeRemoteAddress ? exchange.Request.RemoteAddress : null
+            let session = options.IncludeSessionId ? exchange.Session : null
+            let request = new HttpExchangeRequest(exchange.Request.Method, requestUri, filteredRequestHeaders, remoteAddress)
+            let response = new HttpExchangeResponse(exchange.Response.Status, filteredResponseHeaders)
+            select new HttpExchange(request, response, exchange.Timestamp, principal, session, exchange.TimeTaken));
+
+        if (options.Reverse)
+        {
+            recentExchanges = recentExchanges.OrderByDescending(exchange => exchange.Timestamp).ToList();
```
bart-vmware, line 57: Please add a test to confirm the proper ordering.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,217 @@
...
+        _logger = loggerFactory.CreateLogger<HttpExchangesDiagnosticObserver>();
+    }
+
+    public HttpExchangesResult GetHttpExchanges()
+    {
+        HttpExchange[] recentExchanges = Queue.ToArray();
+        foreach (HttpExchange exchange in recentExchanges)
+        {
+            FilterHeaders(exchange.Request.Headers, _optionsMonitor.CurrentValue.RequestHeaders);
+            FilterHeaders(exchange.Response.Headers, _optionsMonitor.CurrentValue.ResponseHeaders);
```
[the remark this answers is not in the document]
bart-vmware: Nicely solved with query syntax!

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
bart-vmware: > value is an anonymous type with a HttpContext as a property... I can't seem to get type checking to work in this scenario and haven't found anything in several web searches.

These are just the tests acting silly by passing an anonymous type. As can be seen [here](https://github.com/dotnet/aspnetcore/blob/main/src/Hosting/Hosting/src/Internal/HostingApplicationDiagnostics.cs#L88), ASP.NET always passes in a `HttpContext` instance. So the code can just be:
```c#
if (value is HttpContext httpContext)
{
    RecordHttpExchange(current, httpContext);
}
```
When using that and debugging the test `AddSeveralActuators_WebApplicationBuilder_NoConflict` (breakpoint in method `HttpExchangesDiagnosticObserver.ProcessEvent`), it always sends an `HttpContext` instance.

> You're probably right about the other observers, I'm not sure yet if it's something I want to include in this PR though

Why not, do you have any concerns? The current names are just confusing.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
bart-vmware: I'm not seeing it, can you clarify?

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
```diff
@@ -0,0 +1,216 @@
...
+
+        var uriBuilder = new StringBuilder();
+        uriBuilder.Append(request.Scheme);
+        uriBuilder.Append(Uri.SchemeDelimiter);
+        uriBuilder.Append(request.Host);
+        uriBuilder.Append(request.Path);
+        uriBuilder.Append(request.QueryString);
+
+        return new Uri(uriBuilder.ToString());
+    }
```
bart-vmware, line 196: I find this rather brittle. For example, does `.QueryString` include the `?` or not? Why not just use `request.GetEncodedUrl()`? (looking into its implementation, it's more efficient too)

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesEndpointOptions.cs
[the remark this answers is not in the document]
bart-vmware: I'm fine with the current name, its documentation explains how it works.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesResult.cs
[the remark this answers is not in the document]
bart-vmware: As discussed, we should give this some more thought. I agree it doesn't make sense to set these headers from a central place, unrelated to what a specific endpoint handler supports.

`IEndpointHandler<>` should be enhanced to participate in content negotiation, possibly by introducing an extra optional interface. This needs more design first. Probably better to address at a later time. Do we have a tracking issue?

src/Management/src/Endpoint/Actuators/HttpExchanges/ServiceCollectionExtensions.cs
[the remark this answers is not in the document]
bart-vmware: Looks like this was reverted already, closing the conversation.

src/Management/src/Endpoint/Actuators/RouteMappings/RouteMappingsEndpointHandler.cs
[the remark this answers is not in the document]
bart-vmware: Ah, that explains it. There are no controllers. We should look into support for minimal APIs, either through the action descriptor or endpoint explorer APIs. Needs investigation. Can you open an issue?

bart-vmware (changes requested):

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
TimHess: I'll add that check to `GetHttpExchanges_ReturnsHttpExchanges`

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: > These are just the tests acting silly by passing an anonymous type...

I should know by now not to trust that these tests are always doing the right thing 🙄 

> Why not, do you have any concerns? The current names are just confusing. The other observers suffer from the same issues as addressed in this PR, such as obtaining `Activity.Current` when not needed. Please open an issue otherwise.

Not any specific concerns, more a question of maintaining focus... If I have time to tackle it before your next review I will, otherwise I'll create a follow-up issue so this PR doesn't keep dragging on

src/Management/src/Endpoint/EndpointOptionsExtensions.cs
TimHess: Apps Manager needs to be updated to support /httpexchanges as Spring also provides no support for /trace or /httptrace as of Boot 3, which is now 2 years old

src/Management/src/Endpoint/Actuators/HttpExchanges/ConfigureHttpExchangesEndpointOptions.cs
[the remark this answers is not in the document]
TimHess: Users of Tanzu Cloud Foundry will not be able to see their HTTP Exchanges until support is added in Apps Manager. As of a couple of weeks ago, apparently no requests from customers with Spring apps had made it to that engineering team, so it is unclear how to me how popular that feature is. That makes some amount of sense given that this is a fairly simple version of request tracing in comparison with what is offered by OpenTelemetry and other more advanced systems that provide the ability to trace requests between multiple services.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: Request header keys are all lower-case (regardless of how they were sent in), while Response headers are Pascal-Case

src/Management/src/Endpoint/Actuators/RouteMappings/RouteMappingsEndpointHandler.cs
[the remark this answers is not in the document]
TimHess: https://github.com/SteeltoeOSS/Steeltoe/issues/1366

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesResult.cs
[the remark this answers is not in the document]
TimHess: https://github.com/SteeltoeOSS/Steeltoe/issues/1367

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: I can do the delete & rename, but it looks like we can't currently adjust the reflection-based property access code: https://github.com/dotnet/runtime/issues/82910

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
bart-vmware: Yes, we already talked about that. Spring calls `tolower` on request headers, but not on response headers. Then we agreed that Steeltoe should not follow Spring in this case, so Steeltoe should preserve the casing of both request and response headers. Did something change or am I misunderstanding?

bart-vmware (changes requested):
I've completed a full sweep of all changes. Please don't rebase anymore until this PR is approved.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: Oh right... I didn't come back and check on other usages of `GetHttpContextPropertyValue` after trying to adjust the other observers. I'll remove this method and the two tests. The runtime issue is the reason why I can't do the same for the HttpClient Observer(s)

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: Not missing anything, I've no intention of making the casing match Spring here. We should just be reporting headers as they were exchanged

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/HttpExchangesEndpointOptionsTest.cs
```diff
@@ -0,0 +1,74 @@
...
+            ["management:endpoints:httpExchanges:includeRequestHeaders"] = "false",
+            ["management:endpoints:httpExchanges:includeResponseHeaders"] = "false",
+            ["management:endpoints:httpExchanges:includePathInfo"] = "false",
+            ["management:endpoints:httpExchanges:includeUserPrincipal"] = "true",
+            ["management:endpoints:httpExchanges:includeQueryString"] = "true",
+            ["management:endpoints:httpExchanges:includeRemoteAddress"] = "true",
+            ["management:endpoints:httpExchanges:includeSessionId"] = "true",
+            ["management:endpoints:httpExchanges:reverse"] = "false",
+            ["management:endpoints:cloudfoundry:enabled"] = "true"
+        };
```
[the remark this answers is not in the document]
TimHess: I don't think we should capture form parameters as that was never included in /httptrace, only a feature in /trace. So it was effectively removed years ago. I think we would be obligated to add filtering on that too and I'd just rather not have it at alll

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/HttpExchangesEndpointOptionsTest.cs
```diff
@@ -0,0 +1,74 @@
...
+
+        var endpointOptions = GetOptionsFromSettings<HttpExchangesEndpointOptions>(appsettings);
+        var cloudFoundryEndpointOptions = GetOptionsFromSettings<CloudFoundryEndpointOptions>(appsettings);
+
+        cloudFoundryEndpointOptions.Enabled.Should().BeTrue();
+        cloudFoundryEndpointOptions.Id.Should().Be(string.Empty);
+        cloudFoundryEndpointOptions.Path.Should().Be(string.Empty);
+        cloudFoundryEndpointOptions.ValidateCertificates.Should().BeTrue();
```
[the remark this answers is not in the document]
TimHess: There is no reason to include anything Cloud Foundry related here anymore... It made sense when the id/path changed for that context, but since that has been removed the CF bits can come out of this test.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
TimHess: adding test `HttpExchangesActuator_DoesNotCaptureAuthInUri`

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchange.cs
```diff
@@ -0,0 +1,37 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Text.Json.Serialization;
+using System.Xml;
+
+namespace Steeltoe.Management.Endpoint.Actuators.HttpExchanges;
+
+public sealed class HttpExchange
+{
+    public DateTime Timestamp { get; }
```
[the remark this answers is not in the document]
TimHess: Sure, it looks like we can easily assert that there's an http exchange timestamped with the current year-month-date.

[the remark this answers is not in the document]
bart-vmware: This is still a UTC date/time in the response, not a timestamp in Unix Epoch ticks.

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/EndpointMiddlewareTest.cs
```diff
@@ -0,0 +1,87 @@
...
+        HttpResponseMessage response = await httpClient.GetAsync(new Uri("http://localhost/actuator/httpexchanges"));
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+
+        var httpExchanges = await response.Content.ReadFromJsonAsync<HttpExchangesResult>();
+        httpExchanges!.Exchanges.Should().BeEmpty();
+
+        response = await httpClient.GetAsync(new Uri("http://localhost/actuator/httpexchanges"));
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        string json = await response.Content.ReadAsStringAsync();
+        json.Should().NotContain("username").And.NotContain("password");
```
bart-vmware, line 68: This doesn't make sense, there's no username/password anywhere in this test.

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/EndpointMiddlewareTest.cs
```diff
@@ -0,0 +1,87 @@
...
+        json.Should().Contain("http://localhost:80/actuator/httpexchanges");
+        json.Should().Contain($"\"timestamp\":\"{DateTime.UtcNow.Date:yyyy-MM-dd}");
```
bart-vmware, line 70: These assertions are way insufficient, see my other comment.

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchangesDiagnosticObserver.cs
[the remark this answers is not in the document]
bart-vmware: Okay, I understand that `HttpClientObserver` needs to keep using reflection due to the runtime issue. But that doesn't apply to `AspNetCoreHostingObserver`, so please update to remove the reflection there.

bart-vmware (changes requested):

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchange.cs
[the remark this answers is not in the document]
TimHess: Correct, I think it never should have been ticks in /httptrace either

src/Management/src/Endpoint/Actuators/HttpExchanges/HttpExchange.cs
```diff
@@ -0,0 +1,37 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Text.Json.Serialization;
+using System.Xml;
+
+namespace Steeltoe.Management.Endpoint.Actuators.HttpExchanges;
+
+public sealed class HttpExchange
+{
+    [JsonPropertyName("TimeTaken")]
```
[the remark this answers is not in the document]
TimHess: Sorry, that was a combination of copy/paste from the property name and being distracted by time _stamp_ in the response without paying close enough attention. Thanks for writing the tests, I pulled in that commit

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/EndpointMiddlewareTest.cs
TimHess: This test and the one above now effectively do the same thing but with different assertions, so I'm just going to consolidate them and fill out the assertion set for all of the properties

bart-vmware (approved):

PR #1386, opened 2024-10-18
Remove host builder extension methods for Actuator endpoints

bart-vmware:
## Description

The current public API for actuator endpoints provides various ways to achieve the same goals, which is confusing. This PR simplifies that by removing all the host builder extension methods and enhancing the existing service collection extension methods to perform the setups required for actuators to function properly. New host builders are regularly being introduced, so acting on the service collection is more flexible from a consumption point of view: Steeltoe doesn't need to be updated every time a new host builder surfaces.

Because the actuator setup requires middleware registration, it internally uses a startup filter that registers the required middleware. But because the order in which middleware executes is critical and may depend on the use of third-party middleware, the new extension methods take a boolean to indicate _not_ to configure the middleware. This gives callers complete control over the execution order, at the cost of manually calling the appropriate middleware needed for actuators.

At its core, adding actuators can be as simple as:
```c#
var builder = WebApplication.CreateBuilder();
builder.Services.AddEnvironmentActuator();
builder.Services.AddHealthActuator();
// or: builder.Services.AddAllActuators(); // enables to turn actuators on/off in configuration at runtime
```

To configure authentication (which always applies to all registered actuators), a separate extension method is provided:
```c#
builder.Services.ConfigureActuatorEndpoints(endpoints => endpoints.RequireAuthorization("TestAuth"));
```

To customize the CORS policy (which always applies to all registered actuators), use the following method:
```c#
builder.Services.ConfigureActuatorsCorsPolicy(policy => policy.WithOrigins("http://example.api.com"));
```

It doesn't matter anymore in which order the above extension methods are called, they can be used before or after adding actuators.

To control the middleware execution order, pass `false` to every `Add*Actuator` method and then configure the pipeline explicitly:
```c#
var builder = WebApplication.CreateBuilder();
builder.Services.AddEnvironmentActuator(false);
builder.Services.AddHealthActuator(false);
builder.Services.AddCloudFoundryActuator(false);
await using WebApplication app = builder.Build();

app.UseManagementPort();
app.UseCloudFoundrySecurity();
app.UseRouting();
app.UseActuatorsCorsPolicy();
app.UseActuatorEndpoints();

await app.StartAsync();
```

Aside from this, the public API has been adapted to support building third-party actuators, making the core building blocks publicly available.

All health contributors are changed from scoped to singleton, because otherwise it's not possible to obtain the `IEnumerable` of middlewares in tests: it would throw when fetched from the root container, and return two instances when fetched from a scoped container (which is wrong).

Bugfixes:
- Repaired broken CORS and added tests
- Corrections in middleware pipeline execution order
- Invalid JSON returned from thread dump actuator
- Thread safety issue in `AddSteeltoe()`

## Quality checklist


src/Management/src/Endpoint/Actuators/Health/EndpointServiceCollectionExtensions.cs
```diff
@@ -60,7 +95,7 @@ public static IServiceCollection AddHealthContributor<T>(this IServiceCollection
     {
         ArgumentNullException.ThrowIfNull(services);
 
-        services.TryAddEnumerable(ServiceDescriptor.Scoped<IHealthContributor, T>());
+        services.TryAddEnumerable(ServiceDescriptor.Singleton<IHealthContributor, T>());
```
TimHess, line 98: Do you think there is anything to be concerned about with the change from scoped to singleton? How likely is it that a custom `IHealthContributor` would be negatively affected?

src/Management/src/Endpoint/Actuators/ThreadDump/ThreadDumpV2Converter.cs
```diff
@@ -20,14 +20,14 @@ public override void Write(Utf8JsonWriter writer, IList<ThreadInfo> value, JsonS
 
         writer.WriteStartObject();
         writer.WritePropertyName("threads");
-        writer.WriteStartObject();
+        writer.WriteStartArray();
```
TimHess, line 23: This change definitely seems correct, but how did you discover the error?

src/Management/src/Endpoint/Actuators/CloudFoundry/EndpointServiceCollectionExtensions.cs
```diff
@@ -18,12 +22,59 @@ public static class EndpointServiceCollectionExtensions
...
+
+    /// <summary>
+    /// Adds the Cloud Foundry actuator to the service container.
+    /// </summary>
+    /// <param name="services">
+    /// The <see cref="IServiceCollection" /> to add services to.
+    /// </param>
+    /// <param name="configureMiddleware">
+    /// When <c>false</c>, skips configuration of the ASP.NET middleware pipeline. While this provides full control over the pipeline order, it requires to
+    /// manually add the appropriate middleware for actuators to work correctly.
```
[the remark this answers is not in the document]
bart-vmware: Thanks, please do provide suggestions to make the text easier to read. Or does it solely apply to this sentence, which occurs multiple times?

src/Management/src/Endpoint/Actuators/Health/EndpointServiceCollectionExtensions.cs
bart-vmware: I can't think of a scenario where this would be useful. However, if a health contributor needs it, an `IServiceProvider` can be injected to create a scope, or `IHttpContextAccessor` to obtain scoped services bound to the currently executing HTTP request.

Also, documentation at https://docs.steeltoe.io/api/v3/management/health.html mentions:
> The following example shows how to enable the `health` endpoint and add a custom `IHealthContributor` to the service container by adding `CustomHealthContributor` _as a singleton_.

which makes me think using scoped health contributors is not the recommended practice.

src/Management/src/Endpoint/Actuators/ThreadDump/ThreadDumpV2Converter.cs
bart-vmware: During the implementation of `ActuatorsHostBuilderTest`, I'd typically write the test, then put a breakpoint on the returned response body text, and then paste it in a tool to reformat the returned JSON, to determine what to assert on. The tool reported the pasted JSON was invalid.

src/Management/src/Endpoint/Actuators/CloudFoundry/EndpointServiceCollectionExtensions.cs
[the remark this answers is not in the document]
TimHess: I was referring to multiple occurrences of this sentence

TimHess (approved):
LGTM

PR #1402, opened 2024-11-01
Bugfix when using empty path in EndpointOptions

bart-vmware:
## Description

Fix: The path value is ignored when configuring endpoint options with an alternative ID and an empty path.

The hypermedia actuator uses an empty ID and an empty path by default, which makes it impossible to reference it in exposure settings. That's probably why Steeltoe skips the exposure check entirely when the ID is empty, effectively resulting in implicit exposure of the hypermedia actuator. To overcome that, a non-empty ID needs to be assigned. In that case, it should be possible to keep the path empty by explicitly configuring it.

```json
{
  "Management": {
    "Endpoints": {
      "Actuator": {
        "Id": "hypermedia",
        "Path": "",
        "Exposure": {
          "Exclude": [ "hypermedia" ]
        }
      }
    }
  }
}
```

## Quality checklist

### DIFF
--- a/src/Management/src/Abstractions/Configuration/EndpointOptions.cs
+++ b/src/Management/src/Abstractions/Configuration/EndpointOptions.cs
@@ -27,7 +27,7 @@ public abstract class EndpointOptions
     {
         get
         {
-            if (!string.IsNullOrEmpty(_path))
+            if (_path != null)
             {
                 return _path;
             }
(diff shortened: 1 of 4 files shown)


TimHess:
Restating to make sure I fully understand... you are saying there's currently a problem with:

1. disabling exposure of the hypermedia actuator 
2. erasing the base path used for other actuators (such as it's not possible to listen at `http://localhost/prometheus`)

What happens if you add an id for hypermedia and set the path to empty string but don't turn off it's exposure? In other words, could that create a situation where requests to the application root (`http://localhost/`) returns a response from the hypermedia actuator instead of the application itself?

bart-vmware:
Not entirely.

1. Correct.
2. No, erasing the base path is unrelated to this, which happens via management options (so `management:endpoints:path`, not endpoint options, such as hypermedia endpoint options here at `management:endpoints:actuator`). The Prometheus change is an unrelated correction: it doesn't make sense to use ID when building the path.

> What happens if...
Base path (in management options) is unrelated to all of this. I didn't want to change Steeltoe to use a non-empty ID, because the CF actuator does the same (and then we have two names for the same thing). By keeping it this way, my changes are contained and don't potentially break dependent code.

TimHess (approved):

PR #1413, opened 2024-12-05
Fix crash with latest SBA

### LINKED ISSUE #1414: Prior to adding full support for log groups, Steeltoe could return an empty node in the Loggers response

bart-vmware:
## Description

Fix crash with latest Spring Boot Admin by returning an empty loggers group from `/loggers` actuator. Changed internal response object from an unstructured dictionary to the structure expected. Also changed the response body to be empty instead of `{}` on POST. Improved test coverage.

image

image

Sidenote: It looks like Apps Manager is having problems with the levels: 7 levels in header, but only 6 circles.

Fixes #1414 

## Quality checklist


src/Management/test/Endpoint.Test/Actuators/Loggers/TestLoggerProvider.cs
```diff
@@ -8,34 +8,34 @@
...
 
-internal sealed class TestLogProvider : IDynamicLoggerProvider
+internal sealed class TestLoggerProvider : IDynamicLoggerProvider
 {
-    public string? Category { get; private set; }
-    public LogLevel MinLevel { get; private set; }
-    public bool GetLoggerConfigurationsCalled { get; private set; }
+    public bool HasCalledGetLogLevels { get; private set; }
+    public string? SetCategory { get; private set; }
+    public LogLevel SetMinLevel { get; private set; }
```
TimHess, line 15: Should this be `SetMinimumLevel` and (later in the file) `minimumLevel`?

src/Management/src/Endpoint/Actuators/Loggers/LoggersEndpointMiddleware.cs
```diff
@@ -92,17 +89,20 @@ private async Task<Dictionary<string, string>> DeserializeRequestAsync(Stream st
...
+        if (result is { HasError: true })
         {
             context.Response.StatusCode = (int)HttpStatusCode.BadRequest;
         }
         else
         {
-            context.HandleContentNegotiation(_logger);
-            JsonSerializerOptions options = ManagementOptionsMonitor.CurrentValue.SerializerOptions;
-            await JsonSerializer.SerializeAsync(context.Response.Body, result.Data, options, cancellationToken);
+            if (result != null)
```
TimHess, line 100: Can we return `204 No Content` when there is success and no response body? That would be more correct and match how Spring works

src/Management/test/Endpoint.Test/Actuators/Loggers/TestLoggerProvider.cs
bart-vmware: I don't think so. It would be inconsistent with the parameter coming from the `IDynamicLoggerProvider` interface:
```c#
public void SetLogLevel(string categoryName, LogLevel? minLevel)
{
    SetCategory = categoryName;
    SetMinLevel = minLevel ?? LogLevel.None;
}
```

bart-vmware: I did make the property nullable, though. Setting it to `None` is simply wrong.

TimHess (approved):
LGTM

PR #1416, opened 2024-12-09
Revisit Metrics and Tracing

### LINKED ISSUE #1279: Revisit Metrics and Tracing integrations
Follow ups from #1278:
- Is there significant value in the Steeltoe Tracing package? Compare with what Aspire does with regard to OpenTelemetry, maybe Steeltoe should exit this space or provide a slimmer way of helping users set up tracing
- Export to Jaeger has been deprecated in favor of OTLP, we should remove Jaeger references
- Wavefront integration does not behave like the other exporters (eg: private reference), it should align better 
  - The Wavefront exporter should be tested against Wavefront again. I noticed the [tracing sample](https://github.com/SteeltoeOSS/Samples/tree/main/Management/src/Tracing) will fail repeatedly sending metrics with the non-key that's configured, should we do more than check for the presence of a Uri when turning that on?

Work on adding Zipkin support for OTLP has begun [here](https://github.com/openzipkin-contrib/zipkin-otel), but does not yet look ready for the Zipkin exporting to go away

Additionally, with regard to metrics, specifically Prometheus:
- Does the actuator actually behave like an actuator? It looks like we're just configuring OpenTelemetry, I'm not sure it's actually integrated
- Does everything with the exporter actually work? While updating management samples, Prometheus scraping did not work. Encountered error: `invalid metric type "CLR memory usage gauge"`

TimHess:
## Description

- Removes Wavefront integration 
- Remove most of the methods for configuring OpenTelemetry (OpenTelemetry configuration suggestions will move to documentation), including configuration of the Zipkin exporter
- Enhance Prometheus exporter configuration to actually apply the actuator conventions
   - Note there are probably still deficiencies here, such as configuring the verb likely doesn't work

Resolves #1279 

## Quality checklist


src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -25,11 +30,39 @@ public static class PrometheusExtensions
...
+    /// <param name="configureBranchedPipeline">
+    /// Optional callback to configure the branched pipeline. Called before registration of the Prometheus middleware.
+    /// </param>
+    /// <returns>
+    /// The incoming <paramref name="services" /> so that additional calls can be chained.
+    /// </returns>
+    public static IServiceCollection AddPrometheusActuator(this IServiceCollection services, bool configureMiddleware,
+        Action<IApplicationBuilder>? configureBranchedPipeline)
     {
         ArgumentNullException.ThrowIfNull(services);
```
bart-vmware, line 56: Should throw `InvalidOperationException` when `configureMiddleware` is `false` and `configureBranchedPipeline` is not null (and have a test for it).

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -39,19 +72,80 @@ public static IServiceCollection AddPrometheusActuator(this IServiceCollection s
...
+    /// Adds the Steeltoe-configured OpenTelemetry Prometheus exporter to an <see cref="IApplicationBuilder" /> instance.
+    /// </summary>
+    /// <param name="builder">
+    /// The <see cref="IApplicationBuilder" />.
+    /// </param>
+    /// <returns>
+    /// The incoming <paramref name="builder" /> so that additional calls can be chained.
+    /// </returns>
     public static IApplicationBuilder UsePrometheusActuator(this IApplicationBuilder builder)
+    {
```
bart-vmware, line 85: Please add a test for this: `services.AddPrometheusActuator(configureMiddleware: false)` and set up the pipeline manually.

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -39,19 +72,80 @@ public static IServiceCollection AddPrometheusActuator(this IServiceCollection s
...
+                return builder;
+            }
+
+            string cloudFoundryPath = $"/{ConfigureManagementOptions.DefaultCloudFoundryPath}/{endpointPath}".Replace("//", "/", StringComparison.Ordinal);
+            builder.UseOpenTelemetryPrometheusScrapingEndpoint(null, null, cloudFoundryPath, ConfigureBranchedPipeline, null);
+        }
+
+        return builder;
+
+        void ConfigureBranchedPipeline(IApplicationBuilder branchedApplicationBuilder)
```
bart-vmware, line 135: It's a bit confusing to have both the `ConfigureBranchedPipeline` and the `configureBranchedPipeline` parameter, which are related but not the same.
What looks wrong is that when `configureBranchedPipeline` is null, it _still_ sets up the pipeline. I think the logic in `ConfigureBranchedPipeline` should move into `ConfigurePrometheusActuatorStartupFilter`, to make it possible to a) take all the defaults, or b) configure everything yourself.

Perhaps it's easier to use the same mechanism as `ActuatorConventionOptions` (store the callback(s) in the container), so users have an extra method to configure the pipeline instead of passing the full callback as a parameter. Such a method could be: `ConfigurePrometheusEndpoint` (the equivalent of `ConfigureActuatorEndpoints`)

bart-vmware (changes requested):

src/Management/src/Tracing/TracingLogProcessor.cs
```diff
@@ -26,41 +29,34 @@ public string Process(string message)
...
-            SpanContext context = currentSpan.Context;
-
-            var sb = new StringBuilder(" [");
-            sb.Append(options.Name);
-            sb.Append(',');
+            return message;
+        }
 
-            string traceId = context.TraceId.ToHexString();
+        SpanContext context = currentSpan.Context;
```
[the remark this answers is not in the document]
TimHess: `UseShortTraceIds` is available in the Zipkin exporter, but I think it's a pretty legacy option at this point and all it does is take a substring of the trace Id. Also, since we're not configuring the Zipkin exporter anymore, I think it's better to just log the whole id and keep it simpler

src/Management/src/Prometheus/ConfigurePrometheusActuatorStartupFilter.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+
+namespace Steeltoe.Management.Prometheus;
+
+internal sealed class ConfigurePrometheusActuatorStartupFilter(Action<IApplicationBuilder>? configureBranchedPipeline) : IStartupFilter
+{
+    public Action<IApplicationBuilder> Configure(Action<IApplicationBuilder> next)
```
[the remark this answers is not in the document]
bart-vmware: The `IStartupFilter.Configure` parameter is non-nullable, but it was changed here to become nullable. Is the interface definition wrong?

src/Management/src/Tracing/TracingLogProcessor.cs
```diff
@@ -2,70 +2,58 @@
...
 
-            sb.Append(traceId);
-            sb.Append(',');
+        var builder = new StringBuilder(" [");
+        builder.Append(_applicationInstanceInfo.ApplicationName);
+        builder.Append(',');
 
-            sb.Append(context.SpanId.ToHexString());
-            sb.Append(',');
+        string traceId = currentActivity.TraceId.ToHexString();
```
bart-vmware, line 41: Is this variable intentionally not inlined (as opposed to the other variables), to make debugging easier?

src/Management/src/Tracing/TracingServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+using Microsoft.Extensions.DependencyInjection.Extensions;
+using Steeltoe.Common.Extensions;
+using Steeltoe.Logging;
+
+namespace Steeltoe.Management.Tracing;
+
+public static class TracingServiceCollectionExtensions
+{
+    /// <summary>
+    /// Registers an <see cref="IDynamicMessageProcessor" /> that adds tracing details from <see cref="Activity.Current" /> (if found) to log messages.
```
bart-vmware, line 16: It would be nice to mention that it only works in combination with dynamic logging, i.e., `AddDynamicLogging`/`AddDynamicSerilog`.

src/Management/test/Tracing.Test/TracingServiceCollectionExtensionsTest.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Steeltoe.Management.Tracing.Test;
+
+public sealed class TracingServiceCollectionExtensionsTest
+{
+    [Fact]
+    public async Task AddTracingLogProcessor_RegistersTracingLogProcessor()
+    {
+        IServiceCollection services = new ServiceCollection();
+        services.AddSingleton<IConfiguration>(new ConfigurationManager());
```
bart-vmware, line 18:
```suggestion
        services.AddSingleton<IConfiguration>(new ConfigurationBuilder().Build());
```
I hadn't seen usage of `ConfigurationManager` in similar tests before. Perhaps it doesn't matter much, but I prefer to use the existing syntax for consistency, and because the implementation is simpler, ie no auto-reloads etc.

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -25,11 +30,64 @@ public static class PrometheusExtensions
...
+    /// <param name="services">
+    /// The <see cref="IServiceCollection" /> to add services to.
+    /// </param>
+    /// <param name="configureMiddleware">
+    /// When <c>false</c>, skips configuration of the ASP.NET middleware pipeline. While this provides full control over the pipeline order, it requires
+    /// manual addition of the appropriate middleware for the Prometheus exporter to work correctly.
+    /// </param>
+    /// <param name="configurePrometheusPipeline">
+    /// Optional callback to run additional middleware at the Prometheus endpoint path, before the Prometheus middleware runs. For example:
+    /// <code><![CDATA[builder => builder.UseAuthorization()]]></code>. Only used when is <c>true</c>.
```
bart-vmware, line 67:
```suggestion
    /// Optional callback to run additional middleware at the Prometheus endpoint path, before the Prometheus middleware runs. For example:
    /// <code><![CDATA[builder => builder.UseAuthorization()]]></code>. Only used when <paramref name="configureMiddleware" /> is <c>true</c>.
```

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -39,19 +97,111 @@ public static IServiceCollection AddPrometheusActuator(this IServiceCollection s
...
+    }
+
+    /// <summary>
+    /// Adds the Steeltoe-configured OpenTelemetry Prometheus exporter to an <see cref="IApplicationBuilder" /> instance.
+    /// </summary>
+    /// <param name="builder">
+    /// The <see cref="IApplicationBuilder" />.
+    /// </param>
+    /// <param name="configurePrometheusPipeline">
+    /// Optional callback to configure the branched pipeline. Called before registration of the Prometheus middleware.
```
bart-vmware, line 121:
```suggestion
    /// Optional callback to run additional middleware at the Prometheus endpoint path, before the Prometheus middleware runs. For example:
    /// <code><![CDATA[builder => builder.UseAuthorization()]]></code>.
```

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -39,19 +97,111 @@ public static IServiceCollection AddPrometheusActuator(this IServiceCollection s
...
+        }
+
+        if (applyActuatorConventions)
+        {
+            logger.LogWarning("Customizing endpoints is only supported when using endpoint routing.");
+        }
+
+        if (managementOptions.Port is null)
+        {
+            logger.LogWarning("The Prometheus endpoint may not be configured securely. Consider enabling EndpointRouting or using an alternate port.");
```
bart-vmware, line 179:
```suggestion
            logger.LogWarning("The Prometheus endpoint may not be configured securely. Consider enabling endpoint routing or using an alternate port.");
```

src/Management/src/Prometheus/PrometheusExtensions.cs
bart-vmware: The code could be cleaned up further and new cases aren't covered by tests, but I suppose this is good enough.

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -25,11 +30,39 @@ public static class PrometheusExtensions
...
 
+        if (configureMiddleware)
+        {
+            services.TryAddEnumerable(ServiceDescriptor.Transient<IStartupFilter, ConfigurePrometheusActuatorStartupFilter>(_ =>
+                new ConfigurePrometheusActuatorStartupFilter(configureBranchedPipeline)));
+        }
+
         services.AddOpenTelemetry().WithMetrics(builder =>
         {
             builder.AddMeter(SteeltoeMetrics.InstrumentationName);
```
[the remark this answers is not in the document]
bart-vmware: I can't oversee the implications of keeping or removing this line, but I trust your judgment, so please adapt to what makes the most sense.

bart-vmware (changes requested):
Please don't force-push when code review has started. It makes it harder to track how the code changed in response to review comments.

Also, please mention Zipkin removal to the PR description.

src/Management/src/Tracing/TracingLogProcessor.cs
TimHess: I just didn't inline it after removing the short trace id bit

TimHess:
> Please don't force-push when code review has started. It makes it harder to track how the code changed in response to review comments.

Huh. the only commit that should have changed in the force push was the last one, which was only added after your first review

bart-vmware:
> > Please don't force-push when code review has started. It makes it harder to track how the code changed in response to review comments.
> 
> Huh. the only commit that should have changed in the force push was the last one, which was only added after your first review

Ah yes, there was no force-push. But additional changes we added to the Github auto-generated commit, which I didn't expect.

src/Management/src/Tracing/TracingLogProcessor.cs
bart-vmware: Let's inline then.

src/Management/src/Prometheus/ConfigurePrometheusActuatorStartupFilter.cs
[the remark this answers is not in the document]
bart-vmware: Sounds good.

src/Management/src/Prometheus/PrometheusExtensions.cs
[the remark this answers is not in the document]
TimHess: Adding or removing this line does not impact `/metrics`. 

This line was doing something very similar to how users can (and IMO should) add OpenTelemetry metrics instrumentation:
```csharp
services.AddOpenTelemetry().WithMetrics(metrics =>
{
    metrics.AddAspNetCoreInstrumentation().AddHttpClientInstrumentation().AddRuntimeInstrumentation();
});
```

The difference being that `AddMeter(SteeltoeMetrics.InstrumentationName);` adds the Steeltoe observers, some of which monitor the exact same signals as their OpenTelemetry counterparts, leaving too much potential for collision/duplication.

src/Management/src/Prometheus/PrometheusActuatorStartupFilter.cs
```diff
@@ -0,0 +1,22 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.AspNetCore.Builder;
+using Microsoft.AspNetCore.Hosting;
+
+namespace Steeltoe.Management.Prometheus;
+
+internal sealed class PrometheusActuatorStartupFilter(Action<IApplicationBuilder>? configureBranchedPipeline) : IStartupFilter
```
bart-vmware, line 10: Please rename this parameter to `configurePrometheusPipeline` as well.

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -25,33 +29,174 @@ public static class PrometheusExtensions
...
+        {
+            services.TryAddEnumerable(ServiceDescriptor.Singleton<IStartupFilter, PrometheusActuatorStartupFilter>(_ =>
+                new PrometheusActuatorStartupFilter(configurePrometheusPipeline)));
+        }
+
         services.AddOpenTelemetry().WithMetrics(builder =>
         {
-            builder.AddMeter(SteeltoeMetrics.InstrumentationName);
             builder.AddPrometheusExporter();
         });
```
bart-vmware, line 93: Suggestion:
```c#
services.AddOpenTelemetry().WithMetrics(builder => builder.AddPrometheusExporter());
```

src/Management/src/Prometheus/PrometheusExtensions.cs
bart-vmware: Thanks, much better!

bart-vmware (approved):
LGTM

PR #1417, opened 2024-12-12
Enable heap dumps by default in all conditions

TimHess:
## Description

PTRACE is allowed on Linux in Cloud Foundry now

Fixes [integration test expectations](https://github.com/SteeltoeOSS/Samples/pull/343/files#diff-92a27874de2c3856c97366dfe0565034b4c6c50a14737797ea303975c6ebbec0R102)

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/Actuators/HeapDump/HeapDumpEndpointOptions.cs
+++ b/src/Management/src/Endpoint/Actuators/HeapDump/HeapDumpEndpointOptions.cs
@@ -2,16 +2,12 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
-using Steeltoe.Common;
 using Steeltoe.Management.Configuration;
 
 namespace Steeltoe.Management.Endpoint.Actuators.HeapDump;
 
 public sealed class HeapDumpEndpointOptions : EndpointOptions
 {
-    // Default to disabled on Linux + Cloud Foundry until PTRACE is allowed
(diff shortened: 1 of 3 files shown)


src/Management/src/Endpoint/Actuators/HeapDump/HeapDumpEndpointOptions.cs
```diff
@@ -2,16 +2,12 @@
...
 
-using Steeltoe.Common;
 using Steeltoe.Management.Configuration;
 
 namespace Steeltoe.Management.Endpoint.Actuators.HeapDump;
 
 public sealed class HeapDumpEndpointOptions : EndpointOptions
 {
-    // Default to disabled on Linux + Cloud Foundry until PTRACE is allowed
-    internal override bool DefaultEnabled { get; } = !(Platform.IsCloudFoundry && Platform.IsLinux);
```
bart-vmware, line 13: I'm okay with this change, but why not remove the property entirely?

TimHess: The property is still used in `EndpointOptionsExtensions`, but considering all it does it return `true`... I'm ok with just putting `true` _there_ and deleting the property entirely

TimHess: _Actually..._ before making that change, maybe it _is_ useful to have for custom actuator considerations? Sure, we only had one case in existing Steeltoe code, but it was useful for us to have. So maybe we should leave it?

bart-vmware (approved):

PR #1421, opened 2025-01-08
ShowDetails in /health

### LINKED ISSUE #1406: Health Endpoint default for ShowDetails should be Never
[Align with Spring](https://docs.spring.io/spring-boot/reference/actuator/endpoints.html#actuator.endpoints.health) and avoid exposing potentially sensitive information.

### LINKED ISSUE #1411: Bug with health actuator showDetails
## Describe the bug
`management:endpoints:health:showDetails` does not affect whether details of the health check are included in the response.

## Steps to reproduce
1. Find test `...Endpoint.Test.Actuators.Health.EndpointMiddlewareTest.HealthActuator_ReturnsOnlyStatusWhenAuthorized`
1. Observe details in the `health` object during debugging

## Expected behavior
The "details" property in the response should be empty when `showDetails` is `Never` or the user details in the request do not fit the configuration for `WhenAuthorized`

## Additional detail
Steeltoe v4 code (`HealthEndpointHandler.cs`):
```csharp
ShowDetails showDetails = endpointOptions.ShowDetails;
if (showDetails != ShowDetails.Never && (showDetails != ShowDetails.WhenAuthorized || healthRequest.HasClaim))
{
    foreach (string group in endpointOptions.Groups.Select(group => group.Key))
    {
        response.Groups.Add(group);
    }
}
```
Steeltoe v3 code:
```csharp
var showDetails = Options.ShowDetails;
if (showDetails == ShowDetails.Never || (showDetails == ShowDetails.WhenAuthorized && !securityContext.HasClaim(Options.Claim)))
{
    response.Details = new Dictionary<string, object>();
}
else
{
    response.Groups = Options.Groups.Select(g => g.Key);
}
```

TimHess:
## Description

Fix bug where health details aren't cleared out if they shouldn't be returned.
Change default value for ShowDetails to Never

Resolves #1406, Fixes #1411

Documentation updated in https://github.com/SteeltoeOSS/Documentation/pull/336

## Quality checklist


src/Management/src/Endpoint/Actuators/Health/HealthEndpointOptions.cs
```diff
@@ -8,7 +8,7 @@ namespace Steeltoe.Management.Endpoint.Actuators.Health;
 
 public sealed class HealthEndpointOptions : EndpointOptions
 {
-    public ShowDetails ShowDetails { get; set; }
+    public ShowDetails ShowDetails { get; set; } = ShowDetails.Never;
```
bart-vmware, line 11: I recommend reordering the enum members such that it defaults to Never, instead of explicitly setting it here.
Like this: `{ Never, WhenAuthorized, Always }`

The reason is that it's safer that way, in case there are more places (possibly in user code).

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
```diff
@@ -18,9 +18,9 @@ public sealed class EndpointMiddlewareTest : BaseTest
 {
     private readonly Dictionary<string, string?> _appSettings = new()
     {
-        ["management:endpoints:enabled"] = "true",
-        ["management:endpoints:path"] = "/cloudfoundryapplication",
-        ["management:endpoints:health:enabled"] = "true"
+        { "Management:Endpoints:Enabled", "true" },
+        { "Management:Endpoints:Path", "/cloudfoundryapplication" },
+        { "Management:Endpoints:Health:Enabled", "true" }
```
bart-vmware, line 23: Why switch back to the old-style dictionary syntax here?

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
```diff
@@ -298,8 +306,8 @@ public async Task GetStatusCode_OverrideUseStatusCodeFromResponseInConfiguration
         {
             var settings = new Dictionary<string, string?>(_appSettings)
             {
-                ["HealthCheckType"] = "down",
-                ["management:endpoints:UseStatusCodeFromResponse"] = "false"
+                { "HealthCheckType", "down" },
+                { "Management:Endpoints:UseStatusCodeFromResponse", "false" }
```
bart-vmware, line 310: Same there. There's no need to update all dictionary assignments to the new style in this PR, but let's not degrade the existing places.

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
```diff
@@ -313,10 +321,7 @@ public async Task GetStatusCode_OverrideUseStatusCodeFromResponseInConfiguration
         response.StatusCode.Should().Be(HttpStatusCode.OK);
 
         string json = await response.Content.ReadAsStringAsync();
-
-        json.Should().Contain("""
-            "status":"DOWN"
-            """);
+        json.Should().Be("{\"status\":\"DOWN\"}");
```
bart-vmware, line 324: I find the original assertion using a raw string easier to read, don't you agree?

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
```diff
@@ -361,19 +363,18 @@ public async Task GetStatusCode_MicrosoftAggregator_ReturnsExpected()
...
+            { "HealthCheckType", "default" }
         }));
 
         using IWebHost app = builder.Build();
         await app.StartAsync();
 
         using HttpClient client = app.GetTestClient();
         HttpResponseMessage unknownResult = await client.GetAsync(new Uri("http://localhost/cloudfoundryapplication/health"));
-        Assert.Equal(HttpStatusCode.OK, unknownResult.StatusCode);
+        unknownResult.StatusCode.Should().Be(HttpStatusCode.OK);
```
bart-vmware, line 374: Why is this called `unknownResult`? It implies the status is `Unknown`, which isn't actually the case. Should just be named `response`.

src/Management/src/Endpoint/Actuators/Health/HealthEndpointHandler.cs
```diff
@@ -85,6 +85,10 @@ public async Task<HealthEndpointResponse> InvokeAsync(HealthEndpointRequest heal
                 response.Groups.Add(group);
             }
         }
+        else
```
bart-vmware, line 88: I suspect the condition was inverted (compared to v3) to get rid of an empty branch block. But it wasn't actually correct, leading to this bug. Now that we have found both branches are needed, I would prefer to re-invert the condition to make it easier to read (and the same as v3).

So it becomes:
```c#
if (showDetails == ShowDetails.Never || (showDetails == ShowDetails.WhenAuthorized && !healthRequest.HasClaim))
{
    response.Details.Clear();
}
else
{
    foreach (string group in endpointOptions.Groups.Select(group => group.Key))
    {
        response.Groups.Add(group);
    }
}
```

bart-vmware (commented):
Thanks. Basically looks good, added some notes.

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
TimHess: I intended to make the file consistent, but I guess I did it backwards

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
TimHess: That's a question I pondered for a few minutes yesterday, but digging below the surface just got me more confused... Like how is the "default" HealthCheckType going to use the Microsoft Aggregator?? I didn't want to get more involved once I saw that and just left the rest for now

src/Management/src/Endpoint/Actuators/Health/HealthEndpointOptions.cs
bart-vmware: Was it intentional to use `{ Never, Always, WhenAuthorized }` instead of the suggested `{ Never, WhenAuthorized, Always }`?

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
bart-vmware: Code was changed from:
```c#
json.Should().Be("""
    "status":"DOWN"
    """);
```
to:
```c#
json.Should().Be("""{"status":"DOWN"}""");
```
I find the former easier to read, and easier to diff when pasting something over it.

TimHess: The original assertion was also `.Contain`. So the new version uses raw string for readability, and the full string so we don't need to separately assert the lack of other details

src/Management/src/Endpoint/Actuators/Health/HealthEndpointOptions.cs
TimHess: Whoops, I didn't connect the dots that you were suggesting reordering all 3, I just swapped the first two so the default would change.

src/Management/test/Endpoint.Test/Actuators/Health/EndpointMiddlewareTest.cs
bart-vmware: Whether it uses `.Be` or `.Contains`, my point is about the line breaks.

PR #1424, opened 2025-01-11
Move Diagnostics types from Management.Abstractions to Management.Endpoint

bart-vmware:
## Description

This PR moves all types in the namespace `Steeltoe.Management.Diagnostics` in the `Steeltoe.Management.Abstractions` project to the `Steeltoe.Management.Endpoint` project.

Comparing changes with Steeltoe 3.x, I found that some types moved from Endpoint to Abstractions in v4. Upon further inspection, it turns out all of these types are exclusively used by the Metrics actuator. In Steeltoe 3.x, they were shared between Core/Base projects, which required them to live in Abstractions.

## Quality checklist

### DIFF
--- a/src/Management/src/Abstractions/PublicAPI.Unshipped.txt
+++ b/src/Management/src/Abstractions/PublicAPI.Unshipped.txt
@@ -1,5 +1,4 @@
 #nullable enable
-abstract Steeltoe.Management.Diagnostics.DiagnosticObserver.ProcessEvent(string! eventName, object? value) -> void
 Steeltoe.Management.Configuration.EndpointOptions
 Steeltoe.Management.Configuration.EndpointOptions.AllowedVerbs.get -> System.Collections.Generic.IList<string!>!
 Steeltoe.Management.Configuration.EndpointOptions.EndpointOptions() -> void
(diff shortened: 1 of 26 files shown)


src/Management/src/Abstractions/PublicAPI.Unshipped.txt
```diff
@@ -9,42 +8,6 @@ Steeltoe.Management.Configuration.EndpointPermissions
 Steeltoe.Management.Configuration.EndpointPermissions.Full = 2 -> Steeltoe.Management.Configuration.EndpointPermissions
 Steeltoe.Management.Configuration.EndpointPermissions.None = 0 -> Steeltoe.Management.Configuration.EndpointPermissions
 Steeltoe.Management.Configuration.EndpointPermissions.Restricted = 1 -> Steeltoe.Management.Configuration.EndpointPermissions
-Steeltoe.Management.Diagnostics.DiagnosticObserver
```
TimHess, line 12: If we still need `DiagnosticObserver` then I think it (and `IDiagnosticObserver`) should stay in the Abstractions package for extensibility

src/Management/src/Abstractions/PublicAPI.Unshipped.txt
```diff
@@ -9,42 +8,6 @@ Steeltoe.Management.Configuration.EndpointPermissions
...
-Steeltoe.Management.Diagnostics.DiagnosticObserver.ListenerName.get -> string!
-Steeltoe.Management.Diagnostics.DiagnosticObserver.ObserverName.get -> string!
-Steeltoe.Management.Diagnostics.DiagnosticObserver.Subscribe(System.Diagnostics.DiagnosticListener! listener) -> void
-Steeltoe.Management.Diagnostics.IDiagnosticObserver
-Steeltoe.Management.Diagnostics.IDiagnosticObserver.ObserverName.get -> string!
-Steeltoe.Management.Diagnostics.IDiagnosticObserver.Subscribe(System.Diagnostics.DiagnosticListener! listener) -> void
-Steeltoe.Management.Diagnostics.IDiagnosticsManager
-Steeltoe.Management.Diagnostics.IDiagnosticsManager.Start() -> void
-Steeltoe.Management.Diagnostics.IDiagnosticsManager.Stop() -> void
-Steeltoe.Management.Diagnostics.IRuntimeDiagnosticSource
```
TimHess, line 24: I think there's an argument (though probably undocument) in there somewhere for making `IRuntimeDiagnosticSource` extensible, so maybe that should stay here

TimHess (approved):

src/Management/src/Abstractions/PublicAPI.Unshipped.txt
bart-vmware: Even if we still need them, they remain public and extensible. But it is only used by the metrics actuator, so close to that seems the better place to me.

TimHess: Keeping it in Abstractions would be better if (as an example) we wanted to add a `DiagnosticObserver` for Eureka. Sure it's only used by the endpoint, but another implementation likely wouldn't be in an application and might not want the heavier dependency on Endpoint

src/Management/src/Abstractions/PublicAPI.Unshipped.txt
TimHess: Basically the same as `DiagnosticObserver`, though I'm a little less confident on this one. Ultimately, we need to get to the bottom of whether we can remove these entirely

TimHess (approved):
Many of these will be deleted (or moved to support only `httpexchanges`) while addressing #1453, but they will also be internal to the Endpoints project at that point

PR #1438, opened 2025-01-16
Actuator content negotiation

### LINKED ISSUE #1335: Is MediaTypeVersion.V2 still the best version to use as the default for actuators
## Question
This is a follow-up task from [this pr](https://github.com/SteeltoeOSS/Steeltoe/pull/1334#discussion_r1718802728)

### LINKED ISSUE #1367: Revisit Actuator MediaTypeVersion handling
During recent [trace actuator work](https://github.com/SteeltoeOSS/Steeltoe/pull/1356#discussion_r1763102734), we discovered that the Content-Type header in actuator responses is not being handled correctly. It might make sense for individual actuator endpoints to be able to specify their media type (eg: /httpExchanges should only ever return v3) and not exclusively respond to an `Accepts` header and falling back to v2 if the header wasn't found.

We should investigate if any actuator front-ends we care about even pay attention to this header, and if so, how Spring determines the media type to use in the response.

TimHess:
## Description

* Improves content negotiation, provides per-request support for determining version
* Removes most v1/v2 remnants that aren't needed anymore.

Now includes changes from #1445 

Resolves #1335, Resolves #1367

## Quality checklist


bart-vmware (changes requested):
Looks great, happy to see all the startup classes are gone.

Although we decided to change Steeltoe to not perform any content negotiation anymore, I'm concerned adding it later is going to be a breaking change. Therefore I opened #1445, which builds upon this PR.

src/Management/src/Endpoint/Actuators/Health/HealthConverter.cs
```diff
@@ -30,36 +29,13 @@ public override void Write(Utf8JsonWriter writer, HealthEndpointResponse value,
 
         if (value.Details.Count > 0)
         {
-            writer.WritePropertyName("details");
+            writer.WritePropertyName("components");
             writer.WriteStartObject();
 
-            foreach ((string detailKey, object detailValue) in value.Details)
+            foreach (KeyValuePair<string, object> detail in value.Details)
```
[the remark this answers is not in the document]
TimHess: HealthConverter has been removed

src/Management/src/Endpoint/Actuators/Health/HealthConverter.cs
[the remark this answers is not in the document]
TimHess: HealthConverter has been removed

src/Management/src/Endpoint/Actuators/Health/HealthConverter.cs
```diff
@@ -30,36 +29,13 @@ public override void Write(Utf8JsonWriter writer, HealthEndpointResponse value,
 
         if (value.Details.Count > 0)
         {
-            writer.WritePropertyName("details");
+            writer.WritePropertyName("components");
```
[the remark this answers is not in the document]
TimHess: Yeah, I hadn't marked it ready for review yet for some of these reasons - most of these have now been addressed

src/Management/test/Endpoint.Test/Actuators/ThreadDump/EndpointMiddlewareTest.cs
```diff
@@ -21,8 +21,7 @@ public sealed class EndpointMiddlewareTest : BaseTest
     {
         ["management:endpoints:enabled"] = "true",
         ["management:endpoints:dump:enabled"] = "true",
-        ["management:endpoints:actuator:exposure:include:0"] = "threaddump",
-        ["management:endpoints:actuator:exposure:include:1"] = "dump"
+        ["management:endpoints:actuator:exposure:include:0"] = "threaddump"
```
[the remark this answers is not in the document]
TimHess: Yeah, I was planning to break things up a little, but this got big enough that it doesn't make sense to do separately so is now included

src/Management/test/Endpoint.Test/ActuatorsHostBuilderTest.cs
```diff
@@ -214,7 +214,7 @@ public async Task HealthActuatorWithoutDefaultContributors(HostBuilderType hostB
         response.StatusCode.Should().Be(HttpStatusCode.OK);
 
         string responseText = await response.Content.ReadAsStringAsync();
-        responseText.Should().Be("""{"status":"UNKNOWN"}""");
+        responseText.Should().Be("""{"status":"UNKNOWN","groups":["liveness","readiness"]}""");
```
TimHess, line 217: This is a side effect of the combination of serialization changes and a deficiency with group handling, I don't expect the change to stick but leaving it like this for now so we can finalize the other bits first.

src/Management/test/Endpoint.Test/ContentNegotiationTests.cs
```diff
@@ -0,0 +1,234 @@
...
+            }
+        };
+
+        using HttpClient client = host.GetTestClient();
+        HttpResponseMessage response = await client.SendAsync(requestMessage);
+
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+    }
+
+    // AcceptHeader values captured from Spring Boot Admin 3.4.1
```
bart-vmware, line 193: This is a good start, but some entries are missing. Please extract into a separate `ActuatorContentNegotiationTests` class, because these tests are actuator-specific while the other tests assert the generic mechanism.

I've created the table below from `InlineData` to indicate the missing information using `???` cells. Can you adapt the tests accordingly? A few notes on the table:
- If the goal is that Steeltoe behaves identically to a Spring Boot App, then the Spring Boot App column is redundant. I added it because I don't know _if_ that's the case (please let me know). For example, I find it surprising to see Steeltoe respond with V3 when asked for V2/V1/JSON; does SBA do that too?
- The same table can be produced for Cloud Foundry instead of Spring Boot Admin. If the values are the same, we don't need that table, but please verify and indicate if there are any differences. This is also why I've added the `GET /cloudfoundryapplication` entry. Basically all actuators are exposed at both base paths, so I'd also like to know if there are differences in cell values, depending on the base path.

| Endpoint                     | Spring Boot Admin               | Spring Boot App | Steeltoe App             |
| ---------------------------- | ------------------------------- | --------------- | ------------------------ |
| GET /actuator                | Accept: V3, V2, V1, JSON        | ???             | V3                       |
| GET /cloudfoundryapplication | Accept: ???                     | ???             | ???                      |
| GET beans                    | Accept: V2, V1, JSON            | ???             | V3                       |
| GET dbmigrations             | Accept: ???                     | ???             | ???                      |
| GET env                      | Accept: V2, V1, JSON            | ???             | V3                       |
| GET health                   | Accept: V3, V2, V1, JSON        | ???             | V3                       |
| GET heapdump                 | Accept: HTML, XHTML, ..., \*/\* | ???             | application/octet-stream |
| GET httpexchanges            | Accept: V2, V1, JSON            | ???             | V3                       |
| GET info                     | Accept: V3, V2, V1, JSON        | ???             | V3                       |
| GET loggers                  | Accept: V2, V1, JSON            | ???             | V3                       |
| POST loggers                 | Content-Type: ???               | ???             | ???                      |
|                              | Accept: ???                     | ???             | ???                      |
| GET mappings                 | Accept: V2, V1, JSON            | ???             | V2                       |
| GET metrics                  | Accept: V2, V1, JSON            | ???             | V3                       |
| POST refresh                 | Content-Type: ???               | ???             | ???                      |
|                              | Accept: ???                     | ???             | ???                      |
| GET threaddump               | Accept: V2, V1, JSON            | ???             | V3                       |

src/Management/test/Endpoint.Test/ContentNegotiationTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        foreach (string acceptValue in acceptHeader.Split(','))
+        {
+            requestMessage.Headers.Accept.Add(MediaTypeWithQualityHeaderValue.Parse(acceptValue));
+        }
+
+        using HttpClient client = host.GetTestClient();
+        HttpResponseMessage response = await client.SendAsync(requestMessage);
+
+        response.StatusCode.Should().Be(HttpStatusCode.OK);
+        response.Content.Headers.ContentType?.MediaType.Should().Be(expectedContentType);
```
bart-vmware, line 232: This is wrong. If no content type is returned, the test succeeds. Should null-check first instead.

src/Management/test/Endpoint.Test/ContentNegotiationTests.cs
```diff
@@ -0,0 +1,234 @@
...
+        "text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7",
+        "application/octet-stream")]
+    [InlineData("httpexchanges", SpringBootStandardAccept, ActuatorV3)]
+    [InlineData("info", $"{ActuatorV3},{SpringBootStandardAccept}", ActuatorV3)]
+    [InlineData("loggers", SpringBootStandardAccept, ActuatorV3)]
+    [InlineData("mappings", SpringBootStandardAccept, ActuatorV2)]
+    [InlineData("metrics", SpringBootStandardAccept, ActuatorV3)]
+    [InlineData("threaddump", SpringBootStandardAccept, ActuatorV3)]
+    [Theory]
+    public async Task Responses_for_SpringBootAdmin_match_expectations(string endpoint, string acceptHeader, string expectedContentType)
```
bart-vmware, line 208: Consider rename of `expectedContentType` to `responseContentType` to clarify. I imagine the tests will also get a `requestContentType` parameter for POSTs (probably easier to extract into separate `Theory`, where the combination of incoming `Content-Type` and `Accept` headers is tested).

src/Management/src/Endpoint/Actuators/ThreadDump/EndpointServiceCollectionExtensions.cs
```diff
@@ -38,58 +38,22 @@ public static IServiceCollection AddThreadDumpActuator(this IServiceCollection s
     /// The incoming <paramref name="services" /> so that additional calls can be chained.
     /// </returns>
     public static IServiceCollection AddThreadDumpActuator(this IServiceCollection services, bool configureMiddleware)
-    {
```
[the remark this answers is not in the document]
bart-vmware: Leaving this open, as I haven't seen a PR yet that includes this.

bart-vmware (changes requested):

src/Management/src/Endpoint/Middleware/EndpointMiddleware.cs
```diff
@@ -57,23 +57,68 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate? next)
...
+    }
+
     protected abstract Task<TResult> InvokeEndpointHandlerAsync(HttpContext context, CancellationToken cancellationToken);
 
     protected virtual async Task WriteResponseAsync(TResult result, HttpContext context, CancellationToken cancellationToken)
     {
         ArgumentNullException.ThrowIfNull(context);
 
-        context.HandleContentNegotiation(_logger);
+        context.Response.Headers.Append("Content-Type", ContentType);
```
bart-vmware, line 121: Beware that `CloudFoundrySecurityMiddleware.ReturnErrorAsync` _also_ sets the content type. It looks like this isn't correct anymore because `EndpointMiddleware<,>.WriteResponseAsync` overwrites that later.

src/Management/src/Endpoint/Actuators/RouteMappings/RouteMappingsEndpointMiddleware.cs
```diff
@@ -17,6 +17,8 @@ internal sealed class RouteMappingsEndpointMiddleware(
     IRouteMappingsEndpointHandler endpointHandler, IOptionsMonitor<ManagementOptions> managementOptionsMonitor, ILoggerFactory loggerFactory)
     : EndpointMiddleware<object?, RouteMappingsResponse>(endpointHandler, managementOptionsMonitor, loggerFactory)
 {
+    protected override string ContentType { get; } = "application/vnd.spring-boot.actuator.v2+json";
```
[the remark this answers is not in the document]
TimHess: No, but Spring Boot Admin will not request v3 and interprets the v3 response header as a v1 response... Apps Manager does not care. This is why I don't love hardcoding the values

src/Management/src/Endpoint/Middleware/EndpointMiddleware.cs
```diff
@@ -57,23 +57,68 @@ public async Task InvokeAsync(HttpContext context, RequestDelegate? next)
...
```
bart-vmware, line 121: It would be more correct to set the content type _after_ checking whether `result` is null. I've tried to write a test, but we have no built-in actuators for that:
- Loggers endpoint has POST, but overrides this method
- Refresh endpoint has POST, but returns `{}` when `ReturnConfiguration` is set to `false`

src/Management/test/Endpoint.Test/ContentNegotiationTests.cs
bart-vmware, line 1: The following tests fail, while they should succeed:
```c#
[Fact]
public async Task Cannot_use_incompatible_accept_header_in_heap_dump_actuator()
{
    var appSettings = new Dictionary<string, string?>(AppSettings)
    {
        ["management:endpoints:heapDump:heapDumpType"] = "gcdump"
    };

    WebApplicationBuilder builder = TestWebApplicationBuilderFactory.Create();
    builder.Configuration.AddInMemoryCollection(appSettings);
    builder.Services.AddHeapDumpActuator();

    await using WebApplication host = builder.Build();
    await host.StartAsync();

    using HttpClient client = host.GetTestClient();
    using var request = new HttpRequestMessage(HttpMethod.Get, new Uri("http://localhost/actuator/heapdump"));
    request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/json"));
    HttpResponseMessage response = await client.SendAsync(request);

    response.StatusCode.Should().Be(HttpStatusCode.NotAcceptable);
}

[Fact]
public async Task Can_use_compatible_accept_header_in_heap_dump_actuator()
{
    var appSettings = new Dictionary<string, string?>(AppSettings)
    {
        ["management:endpoints:heapDump:heapDumpType"] = "gcdump"
    };

    WebApplicationBuilder builder = TestWebApplicationBuilderFactory.Create();
    builder.Configuration.AddInMemoryCollection(appSettings);
    builder.Services.AddHeapDumpActuator();

    await using WebApplication host = builder.Build();
    await host.StartAsync();

    using HttpClient client = host.GetTestClient();
    using var request = new HttpRequestMessage(HttpMethod.Get, new Uri("http://localhost/actuator/heapdump"));
    request.Headers.Accept.Add(new MediaTypeWithQualityHeaderValue("application/octet-stream"));
    HttpResponseMessage response = await client.SendAsync(request);

}
```

This is because `HeapDumpEndpointMiddleware` doesn't override `ContentType`. Please add these tests and fix the bug.

src/Management/src/Endpoint/Middleware/EndpointMiddleware.cs
bart-vmware: Ah you're right, I didn't realize `CloudFoundrySecurityMiddleware` _doesn't_ derive from `EndpointMiddleware`.

src/Common/src/Common/HealthChecks/HealthCheckResult.cs
```diff
@@ -31,5 +36,6 @@ public sealed class HealthCheckResult
     /// <remarks>
     /// For parity with Spring Boot, repeat the status (in all-caps snake case) here.
     /// </remarks>
+    [JsonIgnoreEmptyCollection]
```
bart-vmware, line 39:
```suggestion
    [JsonIgnoreEmptyCollection]
```

bart-vmware (changes requested):

src/Common/src/Common/HealthChecks/HealthCheckResult.cs
TimHess: This change will be in the next PR, which will also remove all of those repeated statuses from health contributors

bart-vmware (approved):
LGTM, thanks

PR #1454, opened 2025-01-30
Remove metrics endpoint

### LINKED ISSUE #1423: Remove left-overs from .NET Framework observers
The [observer for legacy outgoing HTTP requests on .NET Framework](https://github.com/SteeltoeOSS/Steeltoe/blob/release/3.2/src/Management/src/EndpointBase/Metrics/Observer/HttpClientDesktopObserver.cs) was removed, but there are still references to it, such as [here](https://github.com/SteeltoeOSS/Steeltoe/blob/83d95dcf01e5b557de54be81e100a9c5282e130e/src/Management/src/Abstractions/Diagnostics/MetricsObserverOptions.cs#L32).

The `MetricsObserverOptions.HttpClientDesktop` property should be removed.

May also rename `MetricsObserverOptions.HttpClientCore` to just `.HttpClient` and update its comment. Doing so is a breaking change, though.

Also, this looks broken:
```c#
"HttpClientCoreObserver" => HttpClientCore,
```
Because the type is named `HttpClientObserver`. It looks like there aren't any tests for it.

### LINKED ISSUE #1453: Remove metrics endpoint
The metrics endpoint is pretty easy to use and is nice for diagnosing issues with metrics produced by an application, but because it is based on Steeltoe abstractions that are likely not widely used and there are no known clients for it, we should remove it and instead point to OpenTelemetry.

I am quite open to revisiting this issue is somebody can help identify actual use cases for it, but right now there is no connection between /metrics and /prometheus, and that's a recipe for confusion and an unnecessary maintenance burden.

TimHess:
## Description

Removes the metrics endpoint and alters the abstractions involved.
- Items needed by HttpExchanges to observe ASP.NET Core events have moved and are now internal
- Items not needed by HttpExchanges have been removed

Resolves #1453, Resolves #1423

## Quality checklist


bart-vmware:
Apologies, I squashed #1424 instead of creating a merge commit. As a result, conflicts appeared on this PR, which I resolved using force-push. No conflicts were reported during the rebase.

src/Management/src/Endpoint/Actuators/HttpExchanges/Diagnostics/DiagnosticObserver.cs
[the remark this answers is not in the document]
bart-vmware: Method `protected virtual void Dispose(bool disposing)` can be made non-virtual and private.

src/Management/src/Endpoint/Actuators/Metrics/SystemDiagnosticsMetrics/AggregationManager.cs
bart-vmware, line 1: Accompanied by these deletions, the following lines should be removed from `Steeltoe.All.sln.DotSettings`:
```
	<s:String x:Key="/Default/CodeInspection/ExcludedFiles/FilesAndFoldersToSkip2/=557AFE26_002DB89D_002D497D_002D92B6_002D8268D04396E5_002Fd_003AActuators_002Fd_003AMetrics_002Fd_003ASystemDiagnosticsMetrics/@EntryIndexedValue">ExplicitlyExcluded</s:String>
	<s:String x:Key="/Default/CodeInspection/ExcludedFiles/GeneratedFilesAndFolders/=557AFE26_002DB89D_002D497D_002D92B6_002D8268D04396E5_002Fd_003AActuators_002Fd_003AMetrics_002Fd_003ASystemDiagnosticsMetrics/@EntryIndexedValue">557AFE26-B89D-497D-92B6-8268D04396E5/d:Actuators/d:Metrics/d:SystemDiagnosticsMetrics</s:String>
```

src/Management/src/Endpoint/Actuators/All/EndpointServiceCollectionExtensions.cs
bart-vmware, line 1: Please remove the following lines from `Steeltoe.Management.Endpoint.csproj`:
```xml
<!-- The compiler constants below are required for the copied code in SystemDiagnosticsMetrics -->
<DefineConstants>$(DefineConstants);MEMORYMARSHAL_SUPPORT</DefineConstants>
```

src/Management/test/Endpoint.Test/ActuatorsHostBuilderTest.cs
bart-vmware, line 1: The following line can be removed from `Steeltoe\src\Management\test\Endpoint.Test\ActuatorRouteBuilderExtensionsTest.cs`:
```c#
path = path.Replace("metrics/{**_}", "metrics", StringComparison.Ordinal);
```

bart-vmware:
Can't we just use [`MeterListener`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meterlistener?view=net-8.0) instead of having our own abstractions? See example usage [here](https://stackoverflow.com/a/76898328), or look at the implementation of [`MetricCollector`](https://github.com/dotnet/extensions/blob/release/8.10/src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Metrics/MetricCollector.cs), which is designed for testing (it shows how to subscribe to `MeterListener`).

TimHess:
> Can't we just use [`MeterListener`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meterlistener?view=net-8.0) instead of having our own abstractions? See example usage [here](https://stackoverflow.com/a/76898328), or look at the implementation of [`MetricCollector`](https://github.com/dotnet/extensions/blob/release/8.10/src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Metrics/MetricCollector.cs), which is designed for testing (it shows how to subscribe to `MeterListener`).

If you're asking for the purpose of keeping `/metrics`, yeah, I think we might be able to re-implement it with `MeterListener`, but we don't have a use case for it outside of Spring Boot Admin, and we agreed that pointing to `/prometheus` makes more sense... I also don't think that would work at all for `HttpExchanges`, which needs access to `HttpContext` and not just the metric values, so I don't think it'd be worth the effort.

src/Management/src/Endpoint/Actuators/HttpExchanges/Diagnostics/DiagnosticObserver.cs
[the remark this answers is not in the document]
TimHess: > Method `protected virtual void Dispose(bool disposing)` can be made non-virtual and private.

This change results in an [S3881](https://rules.sonarsource.com/csharp/RSPEC-3881/) error.

src/Management/test/Endpoint.Test/ActuatorsHostBuilderTest.cs
TimHess: I changed the base of this PR to make that bit easier to manage

bart-vmware: That's unfortunate; now I'll need to do the full review again. Parking this PR until its base has been merged.

TimHess: > That's unfortunate; now I'll need to do the full review again. Parking this PR until its base has been merged.

I don't think it'd be much work to switch them around if that makes it easier for you, I think this would be the only conflict

bart-vmware: Yes, please do. This PR is much easier to complete than the other one.

src/Management/src/Endpoint/Actuators/HttpExchanges/Diagnostics/DiagnosticObserver.cs
```diff
@@ -3,27 +3,21 @@
...
 using Microsoft.Extensions.Logging;
 
-namespace Steeltoe.Management.Endpoint.Actuators.Metrics.Diagnostics;
+namespace Steeltoe.Management.Endpoint.Actuators.HttpExchanges.Diagnostics;
 
-public abstract class DiagnosticObserver : IDiagnosticObserver
+internal abstract class DiagnosticObserver : IObserver<KeyValuePair<string, object?>>, IDisposable
 {
+    private const string ListenerName = "Microsoft.AspNetCore";
+    private const string ObserverName = "HttpExchangesDiagnosticObserver";
```
bart-vmware, line 13: This makes the type unusable for anything _except_ HttpExchanges. It creates a dependency between base and derived that shouldn't exist.

bart-vmware (changes requested):

src/Management/src/Endpoint/Actuators/HttpExchanges/Diagnostics/DiagnosticObserver.cs
TimHess: I'm confused, you asked for these to be private fields... it doesn't matter to me either way, seems like something easily sorted if we come up with a need for another DiagnosticObserver

bart-vmware:
> > Can't we just use [`MeterListener`](https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.metrics.meterlistener?view=net-8.0) instead of having our own abstractions? See example usage [here](https://stackoverflow.com/a/76898328), or look at the implementation of [`MetricCollector`](https://github.com/dotnet/extensions/blob/release/8.10/src/Libraries/Microsoft.Extensions.Diagnostics.Testing/Metrics/MetricCollector.cs), which is designed for testing (it shows how to subscribe to `MeterListener`).
> 
> If you're asking for the purpose of keeping `/metrics`, yeah, I think we might be able to re-implement it with `MeterListener`, but we don't have a use case for it outside of Spring Boot Admin, and we agreed that pointing to `/prometheus` makes more sense...
> 
> I also don't think that would work at all for `HttpExchanges`, which needs access to `HttpContext` and not just the metric values, so I don't think it'd be worth the effort. I tried removing these abstractions, but the result is we still need something just like it. Since this PR makes it all internal we could change it up any time down the road if we work out a better way, but IMO these changes are an appropriate amount of effort for what we're trying to do right now

I was referring to the HttpExchanges case. But yeah, we need `HttpContext` instead of numbers, so it's probably not as easy as I assumed. I agree it's not a priority, so let's leave it as-is for now.

src/Management/src/Endpoint/Actuators/HttpExchanges/Diagnostics/DiagnosticObserver.cs
bart-vmware: Private fields, not constants. Their values should still be set by defined types.

bart-vmware: The problem is that tight coupling is being introduced here, which is a code smell. Another smell is that base types should not have knowledge of derived types. Sure, we can change it all later, but that doesn't mean introducing questionable code today is justified. The base and derived types could be unified into a single type, but I don't think that's worth the effort at this time.

src/Management/src/Endpoint/Actuators/HttpExchanges/Diagnostics/DiagnosticObserver.cs
```diff
@@ -3,27 +3,26 @@
...
 
-public abstract class DiagnosticObserver : IDiagnosticObserver
+internal abstract class DiagnosticObserver : IObserver<KeyValuePair<string, object?>>, IDisposable
 {
     private readonly ILogger _logger;
-    private IDisposable? _subscription;
 
-    public string ListenerName { get; }
-    public string ObserverName { get; }
+    private readonly string _observerName;
```
bart-vmware, line 14: This is what it looks like now:

image

Why the blank line after `_logger`, and why not follow the same order as the ctor parameters?

bart-vmware (approved):

PR #1408, opened 2024-11-21
Check exposure/enabled for HTTP verb match in management endpoints

bart-vmware:
## Description

Instead of mapping endpoints in ASP.NET routing per verb, always map all verbs and then filter inside middleware.

Advantages:
- Returns 404 when not exposed/enabled with invalid verb, instead of 405
- Can change verbs at runtime

Disadvantages:
- Verb information is lost in route mappings endpoint (always shows all verbs)
- Entries in route mappings with all verbs disabled are no longer hidden

Fixes #1410.

## Quality checklist

- [x] Your code complies with our [Coding Style](https://github.com/SteeltoeOSS/.github/blob/main/contributing-docs/contributing-code-style.md).
- [x] You've updated unit and/or integration tests for your change, where applicable.
- [ ] You've updated documentation for your change, where applicable.
      If your change affects other repositories, such as [Documentation](https://github.com/SteeltoeOSS/Documentation), [Samples](https://github.com/SteeltoeOSS/Samples) and/or [MainSite](https://github.com/SteeltoeOSS/MainSite), add linked PRs here.
- [ ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.
- [ ] You've added required license files and/or file headers (explaining where the code came from with proper attribution), where code is copied from StackOverflow, a blog, or OSS.

PR #1457, opened 2025-02-06
Revised route mappings actuator

### LINKED ISSUE #155: Mappings Actuator Should Return Razor Pages
/mappings currently doesn't know about razor pages, but it would be nice if it did

### LINKED ISSUE #1366: Better support for Minimal Api in /mappings
As identified during #1356, (specifically [here](https://github.com/SteeltoeOSS/Steeltoe/pull/1356#discussion_r1763108054)) minimal API mappings are not fully supported at this time. We should look deeper than just fixing the crash and see about a more complete solution.

### LINKED ISSUE #1409: Actuator endpoints don't show up in Spring Boot Admin Mappings View

### LINKED ISSUE #1410: Incorrect status code when actuator is disabled
> Interestingly, with default exposure settings, where `/refresh` is not exposed, requests to that actuator return 405 instead of 404. I don't know if it matters that we perform the verb check before an exposure check, but I wasn't expecting that.

See https://github.com/SteeltoeOSS/Steeltoe/pull/1408. Though I'm not sure if doing so is a good idea, there are tradeoffs.

_Originally posted by @bart-vmware in https://github.com/SteeltoeOSS/Documentation/pull/330#discussion_r1851854029_

bart-vmware:
## Description

Revised route mappings actuator to include endpoints from Minimal APIs and Razor Pages, with richer metadata and improved compatibility with Spring. Entries from conventional routing aren't shown anymore.

Incorporates the changes from #1408, special-casing actuator endpoints in the mappings actuator to overcome the mentioned disadvantages.

Centralized string manipulation around slashes in paths (see `EndpointOptions.GetEndpointPath()`) and made it more robust. Fixes the bug that the Prometheus actuator doesn't properly handle custom paths in configuration.

Fail app startup when running on Cloud Foundry without security configured.

### Apps Manager
image

### Spring Boot Admin
image

Fixes #1410, fixes #1409, fixes #1366, fixes #155.

## Quality checklist

  - https://github.com/SteeltoeOSS/Documentation/pull/348


src/Management/src/Endpoint/Configuration/ConfigureManagementOptions.cs
```diff
@@ -50,6 +51,7 @@ public void Configure(ManagementOptions options)
     private static void ConfigureSerializerOptions(ManagementOptions options)
     {
         options.SerializerOptions.DefaultIgnoreCondition = JsonIgnoreCondition.WhenWritingNull;
+        options.SerializerOptions.Encoder = JavaScriptEncoder.UnsafeRelaxedJsonEscaping;
```
TimHess, line 54: Why do we need this? Will we remember (or easily be able to discover) why in 6 months?

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
```diff
@@ -53,8 +53,9 @@ public async Task InvokeAsync(HttpContext context)
 
         _logger.LogDebug("InvokeAsync({RequestPath})", context.Request.Path.Value);
         CloudFoundryEndpointOptions endpointOptions = _endpointOptionsMonitor.CurrentValue;
+        ManagementOptions managementOptions = _managementOptionsMonitor.CurrentValue;
 
-        if (Platform.IsCloudFoundry && endpointOptions.IsEnabled(_managementOptionsMonitor.CurrentValue) &&
+        if (Platform.IsCloudFoundry && endpointOptions.IsEnabled(managementOptions) && managementOptions.IsCloudFoundryEnabled &&
```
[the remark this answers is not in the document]
TimHess: Noting that I've now seen [this condition](https://github.com/SteeltoeOSS/Steeltoe/pull/1457/files#diff-9688be13e85cd1b457188e89bfcb877acb0ad9ebd590b6723f01f701a913a145R58), so it seems like you've already considered the possibility, but would still like confirmation that it's locked down

src/Management/src/Endpoint/Actuators/RouteMappings/RoutingTypes/AspNetEndpoint.cs
```diff
@@ -0,0 +1,75 @@
...
+        HandlerMethod = handlerMethod;
+        Parameters = FilterParameters(parameters).ToList().AsReadOnly();
+        Headers = parameters.Where(parameter => parameter.Origin is AspNetEndpointParameterOrigin.Header).ToList().AsReadOnly();
+        Consumes = consumes.ToHashSet();
+        Produces = produces.ToHashSet();
+    }
+
+    private static IEnumerable<AspNetEndpointParameter> FilterParameters(List<AspNetEndpointParameter> parameters)
+    {
+        // Include parameters from query string and unknown origin for diagnostics purposes.
```
TimHess, line 49: Does `unknown origin for diagnostics purposes.` mean "we don't know where this came from, but you probably should know so here it is" or "for further development of /mappings" or something else?

src/Management/test/Endpoint.Test/Actuators/Hypermedia/HypermediaEndpointTest.cs
```diff
@@ -24,6 +24,14 @@ public async Task Invoke_ReturnsExpectedLinks()
             services.AddInfoActuator();
         };
 
+        testContext.AdditionalConfiguration = configuration =>
```
TimHess, line 27: How was this test passing before without changing the management path?

TimHess (changes requested):
Generally looks really good and is easier to follow than what it replaces, just the listed questions and requests and it should be good to go

src/Management/test/Endpoint.Test/Actuators/Hypermedia/HypermediaEndpointTest.cs
bart-vmware: The original code used the incoming HTTP request URL as input, ignoring what was configured.

src/Management/src/Endpoint/Actuators/RouteMappings/RoutingTypes/AspNetEndpoint.cs
bart-vmware: This means the parameter exists, but that ASP.NET cannot determine the binding origin. Remove the `or AspNetEndpointParameterOrigin.Unknown` clause from the code to see various test failures. For example, in the following MVC action method:

```c#
[HttpPatch("update/{id:guid}")]
[Consumes(typeof(Product), "application/json")]
public async Task UpdateAsync(
    Guid id, // bound from route
    [FromQuery] int? languageId,
    [FromHeader(Name = "X-Media-Version")] string mediaVersion,
    [FromHeader(Name = "X-Include-Details")] string? includeDetails,
    [FromBody] Product product,
    bool failOnError = true) // unknown origin
```
The parameter `failOnError` _possibly_ originates from the query string (in this case, because another parameter already binds from the request body), but there may as well be a custom model binder in place that pulls it from a header; we simply can't know.

Note: The above is how it is _supposed_ to work. However, I found bugs that I'm going to address.

In practice, we're missing out on a lot by _only_ including parameters from routes and headers in the response. This worsens in Razor Pages apps, where the route pattern is page-wide, and page methods likely take extra parameters. Or use bound properties, but I'm not including those because we can't determine which page methods use them.

 I've looked at what Spring does, which is pretty inconsistent. In Spring, for the following declaration:
```java
@RequestMapping(
            value = { "/greeting", "/greeting/{id:[\\d]+}" }
            ,method = { RequestMethod.GET, RequestMethod.POST }
            ,params = { "filter", "sort" } // caution: these are implicitly required now!
            ,headers = { "key1=val1", "key2=val2" }
            ,produces = { "text/plain", "application/special-json-type", "application/xml" }
            ,consumes = { "application/jsonp", "application/xhtml" }
    )
    String home(
            @PathVariable(required = false, name = "id") String greetingId,
            @RequestParam(value = "filter", required = false, defaultValue = "all") String filter,
            @RequestParam(value = "sort", required = false) String sort,
            @RequestHeader(value = "Authorization", required = false) String authToken) { }
```
only what's in `@RequestMapping` will be returned. The above declaration is nice for testing what the Spring actuator shows, but invoking the endpoint doesn't actually work because it has contradictions that never match.

So, for Steeltoe, I decided to favor including parameters that _may_ be relevant (this is a diagnostics tool, after all) in the `params` response field, instead of only the parameters that appear in the route pattern. Injected parameters from the service container are _always_ excluded for Minimal APIs and API controllers (because API Explorer runs a speculative model bind and detects them), and for other endpoints when decorated with `[FromServices]`.

src/Management/src/Endpoint/Configuration/ConfigureManagementOptions.cs
bart-vmware: I added this to make the output less unreadable. Without it,
```
"handlerMethod": {
  "className": "Steeltoe.Management.Endpoint.Test.Actuators.RouteMappings.AppTypes.MinimalApiTest\u002B\u003C\u003Ec",
  "name": "\u003CCan_get_routes_for_groups_using_inline_lambdas\u003Eb__8_0",
  "descriptor": "Void \u003CCan_get_routes_for_groups_using_inline_lambdas\u003Eb__8_0(System.Guid)"
},
```
is returned instead of:
```
"handlerMethod": {
  "className": "Steeltoe.Management.Endpoint.Test.Actuators.RouteMappings.AppTypes.MinimalApiTest+<>c",
  "name": "<Can_get_routes_for_groups_using_inline_lambdas>b__8_0",
  "descriptor": "Void <Can_get_routes_for_groups_using_inline_lambdas>b__8_0(System.Guid)"
},
```
and:
```
"mediaType": "application/xhtml\u002Bxml",
```
instead of:
```
"mediaType": "application/xhtml+xml",
```

I've added a comment to explain.

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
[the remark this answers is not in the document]
bart-vmware: `EndpointMiddleware.CanInvoke` blocks at `/cloudfoundryapplication` when `IsCloudFoundryEnabled` is `false` (unchanged code), but that same check was missing in `CloudFoundrySecurityMiddleware`. There's no point in executing the permissions check if the endpoint will block anyway, which is why I added it. The observable effect is that the user gets a 404 instead of 503.

While looking at `CloudFoundrySecurityMiddlewareTest`, I noticed many redundant configuration settings, which I cleaned up in a separate commit.

What's a bit odd though, is that `CloudFoundrySecurityMiddleware` tries to find the `EndpointOptions` for the current request (which it needs to check permissions), but fails with error "Endpoint is not available" when not found. It may as well just exist early, so that `EndpointMiddleware` catches it and returns a 404.

src/Management/src/Endpoint/Configuration/ConfigureManagementOptions.cs
TimHess: The detailed response and comment addition help a lot, thank you

src/Management/test/Endpoint.Test/Actuators/Hypermedia/HypermediaEndpointTest.cs
TimHess: Sorry, I don't see any related changes in this PR in BaseTest or TestContext that help me understand how it worked before.. where/when did that change?

src/Management/src/Endpoint/Actuators/RouteMappings/RoutingTypes/AspNetEndpoint.cs
TimHess: That is a lot of good context, thanks for sharing. I'm going to add a tiny suggestion that won't nearly do justice to that level of detail but should at least help disambiguate

src/Management/src/Endpoint/Actuators/RouteMappings/RoutingTypes/AspNetEndpoint.cs
```diff
@@ -0,0 +1,75 @@
...
```
TimHess, line 49:
```suggestion
        // Include parameters from query string and unknown origin (where ASP.NET Core can't identify the source) for diagnostics purposes.
```

This line is probably too long

src/Management/test/Endpoint.Test/Actuators/Hypermedia/HypermediaEndpointTest.cs
bart-vmware: The code in `HypermediaService` changed to read the base path from options instead of appending to the request URL. Formerly, it assumed that a request to `/foobar` meant the base path is `/foobar`. Therefore the unit test needs to provide compatible options now.

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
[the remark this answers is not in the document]
TimHess: > `EndpointMiddleware.CanInvoke` blocks at `/cloudfoundryapplication` when `IsCloudFoundryEnabled` is `false` (unchanged code), but that same check was missing in `CloudFoundrySecurityMiddleware`. There's no point in executing the permissions check if the endpoint will block anyway...

`CloudFoundrySecurityMiddleware` should execute for requests to any actuator under the `/cloudfoundryapplication` base path though, not just that exact path/endpoint. It would be odd to be sure, but I guess I'm wondering what happens if `IsCloudFoundryEnabled` is false and a request comes in for say `/cloudfoundryapplication/heapdump`?
 
> While looking at `CloudFoundrySecurityMiddlewareTest`, I noticed many redundant configuration settings, which I cleaned up in a separate commit.
Wow, no kidding... Great cleanup work!

> What's a bit odd though, is that `CloudFoundrySecurityMiddleware` tries to find the `EndpointOptions` for the current request (which it needs to check permissions), but fails with error "Endpoint is not available" when not found. It may as well just exi~s~t early, so that `EndpointMiddleware` catches it and returns a 404.
Are you saying here that we'd return 404 for `/actuator/not-an-endpoint` and 503 for `/cloudfoundryapplication/not-an-endpoint`? If so, I agree that would be wrong and should be fixed

src/Management/test/Endpoint.Test/Actuators/Hypermedia/HypermediaEndpointTest.cs
TimHess: Oh _wow_, yeah... Good catch there, the old behavior is silly

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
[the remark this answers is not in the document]
bart-vmware: > > `EndpointMiddleware.CanInvoke` blocks at `/cloudfoundryapplication` when `IsCloudFoundryEnabled` is `false` (unchanged code), but that same check was missing in `CloudFoundrySecurityMiddleware`. There's no point in executing the permissions check if the endpoint will block anyway...
> 
> `CloudFoundrySecurityMiddleware` should execute for requests to any actuator under the `/cloudfoundryapplication` base path though, not just that exact path/endpoint. It would be odd to be sure, but I guess I'm wondering what happens if `IsCloudFoundryEnabled` is false and a request comes in for say `/cloudfoundryapplication/heapdump`?

Then `CloudFoundrySecurityMiddleware` lets the request pass through, and then `EndpointMiddleware` produces a 404.

> > What's a bit odd though, is that `CloudFoundrySecurityMiddleware` tries to find the `EndpointOptions` for the current request (which it needs to check permissions), but fails with error "Endpoint is not available" when not found. It may as well just exi~s~t early, so that `EndpointMiddleware` catches it and returns a 404.
> 
> Are you saying here that we'd return 404 for `/actuator/not-an-endpoint` and 503 for `/cloudfoundryapplication/not-an-endpoint`? If so, I agree that would be wrong and should be fixed

Yes, that's currently the case.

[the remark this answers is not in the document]
TimHess: > > > `EndpointMiddleware.CanInvoke` blocks at `/cloudfoundryapplication` when `IsCloudFoundryEnabled` is `false` (unchanged code), but that same check was missing in `CloudFoundrySecurityMiddleware`. There's no point in executing the permissions check if the endpoint will block anyway...
> > 
> > 
> > `CloudFoundrySecurityMiddleware` should execute for requests to any actuator under the `/cloudfoundryapplication` base path though, not just that exact path/endpoint. It would be odd to be sure, but I guess I'm wondering what happens if `IsCloudFoundryEnabled` is false and a request comes in for say `/cloudfoundryapplication/heapdump`?
> 
> Then `CloudFoundrySecurityMiddleware` lets the request pass through, and then `EndpointMiddleware` produces a 404.

Sounds good

> 
> > > What's a bit odd though, is that `CloudFoundrySecurityMiddleware` tries to find the `EndpointOptions` for the current request (which it needs to check permissions), but fails with error "Endpoint is not available" when not found. It may as well just exi~s~t early, so that `EndpointMiddleware` catches it and returns a 404.
> > 
> > 
> > Are you saying here that we'd return 404 for `/actuator/not-an-endpoint` and 503 for `/cloudfoundryapplication/not-an-endpoint`? If so, I agree that would be wrong and should be fixed
> 
> Yes, that's currently the case.

OK, so we agree it's undesired behavior... address now or in a follow-up?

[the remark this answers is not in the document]
bart-vmware: I've pushed a commit to address this.

TimHess (approved):
Great work, thanks

PR #1467, opened 2025-02-12
Use DotNet.ReproducibleBuilds to manage packaging best practices

### LINKED ISSUE #1461: Review SourceLink and pdb configurations (embed pdb files?)

TimHess:
## Description

Let SDK manage SourceLink ref - [added in .NET 8](https://github.com/dotnet/sourcelink?tab=readme-ov-file#using-source-link-in-net-projects)
Adopt [`DotNet.ReproducibleBuilds`](https://github.com/dotnet/reproducible-builds?tab=readme-ov-file#dotnetreproduciblebuilds) for packaging best-practices, including embedding debug symbols, and various other settings (that we were mostly already doing)

Fixes #1461

## Quality checklist

### DIFF
--- a/shared-package.props
+++ b/shared-package.props
@@ -22,24 +22,8 @@
     <PackageReadmeFile>PackageReadme.md</PackageReadmeFile>
   </PropertyGroup>
 
-  <PropertyGroup>
-    <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
-  </PropertyGroup>
-
   <PropertyGroup Condition="'$(Configuration)' == 'Release'">
     <GeneratePackageOnBuild>True</GeneratePackageOnBuild>
-  </PropertyGroup>
-
-  <PropertyGroup Condition="'$(TF_BUILD)' == 'true'">
-    <!--
-      While deterministic builds are enabled by default in .NET SDK projects, there is an extra property, ContinuousIntegrationBuild,
-      to set on the build server to normalize stored file paths. These should not be enabled during local dev or the debugger won't
-      be able to find the local source files.
-    -->
-    <ContinuousIntegrationBuild>true</ContinuousIntegrationBuild>
   </PropertyGroup>
 
   <ItemGroup>
@@ -52,7 +36,7 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="$(SourceLinkGitHubVersion)" PrivateAssets="All" />
+    <PackageReference Include="DotNet.ReproducibleBuilds" Version="$(ReproducibleBuildsVersion)" PrivateAssets="All" />
     <PackageReference Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="$(PublicApiAnalyzersVersion)" PrivateAssets="All" />
   </ItemGroup>
 
--- a/shared.props
+++ b/shared.props
@@ -1,4 +1,5 @@
 <Project>
+  <Sdk Name="DotNet.ReproducibleBuilds.Isolated" Version="1.2.25" />
   <PropertyGroup>
     <LangVersion>latest</LangVersion>
     <Nullable>enable</Nullable>
--- a/versions.props
+++ b/versions.props
@@ -21,6 +21,7 @@
     <NewtonsoftJsonVersion>13.0.3</NewtonsoftJsonVersion>
     <PublicApiAnalyzersVersion>3.3.*</PublicApiAnalyzersVersion>
     <RabbitClientTestVersion>7.0.*</RabbitClientTestVersion>
+    <ReproducibleBuildsVersion>1.2.25</ReproducibleBuildsVersion>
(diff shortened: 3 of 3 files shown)


shared.props
```diff
@@ -1,4 +1,5 @@
 <Project>
+  <Sdk Name="DotNet.ReproducibleBuilds.Isolated" Version="1.2.25" />
```
TimHess, line 2: I tried to use the variable instead of duplicating the version number, but encountered `error MSB4236: The SDK 'DotNet.ReproducibleBuilds.Isolated' specified could not be found.` when I tried to use `$(ReproducibleBuildsVersion)` here, and moving the reference to versions.props above this line didn't make a difference

bart-vmware: See https://learn.microsoft.com/en-us/visualstudio/msbuild/how-to-use-project-sdk?view=vs-2022. Have you tried importing `versions.props` above this line?

TimHess: When I saw that project, I was reminded of the last couple rounds of new build/package features being a royal PITA to get working and thought it would be nice to remove an item from our pile of maintenance responsibilities... But you are right about it not really helping us that much today, especially this SDK bit. I'll take the refs back out and we'll keep managing it here, and we'll have this exchange as a record for where to look in the future if things change

shared-package.props
```diff
@@ -52,7 +51,6 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="Microsoft.SourceLink.GitHub" Version="$(SourceLinkGitHubVersion)" PrivateAssets="All" />
```
bart-vmware, line 55: This doesn't look right. Did you test SourceLink actually works?

TimHess: From [the link in the PR description](https://github.com/dotnet/sourcelink?tab=readme-ov-file#using-source-link-in-net-projects):

> If your project uses .NET SDK 8+ and is hosted by the above providers it does not need to reference any Source Link packages or set any build properties to include source control information in the PDB.

shared.props
TimHess: > Something is still not right with this suggestion:
> 
> > The only thing we truly need is embedded to replace IncludeSymbols and SymbolPackageFormat.
> 
> The build still produced .snupkg files, which I think was the reason you can't step into packages from the devops feed. But some combination of switches I tried yesterday did work, so I'll keep at it

Just kidding, I was apparently looking at the wrong build

TimHess:
With the packages built from these changes copied into a local nuget feed:
- With an offline machine and empty symbol cache, debugging into one of the packages resulted in file browser dialog, falling back to what I assume is disassembled code (still looks like our source, but no comments at all, experience was oddly seamless)
- With an online machine and empty symbol cache, debugging into one of the packages results in source code with comments

If we use EmbedAllSources instead of EmbedUntrackedSources, stepping into source works even when offline. Package sizes do increase from a total of 1.3 MB to 1.8 MB on disk (Management.Endpoint would be the largest at 402KB).

shared-package.props
```diff
@@ -24,9 +24,8 @@
 
   <PropertyGroup>
     <PublishRepositoryUrl>true</PublishRepositoryUrl>
-    <IncludeSymbols>true</IncludeSymbols>
-    <SymbolPackageFormat>snupkg</SymbolPackageFormat>
-    <EmbedUntrackedSources>true</EmbedUntrackedSources>
+    <DebugType>embedded</DebugType>
```
bart-vmware, line 27: I like this. Instead of adding .pdb files in the package (the default), the symbols now end up directly in the DLL. I've experienced hard-todiagnose issues in the past with coverage tools not properly resolving the absolute/relative paths (either only working locally, or only working in cibuild), and this solves the whole problem. Another nice side effect is that it avoids the annoying confirmation prompt when stepping into.

With this setting in place, `ContinuousIntegrationBuild` can be removed.

bart-vmware:
> With the packages built from these changes copied into a local nuget feed:
> 
> * With an offline machine and empty symbol cache, debugging into one of the packages resulted in file browser dialog, falling back to what I assume is disassembled code (still looks like our source, but no comments at all, experience was oddly seamless)
> * With an online machine and empty symbol cache, debugging into one of the packages results in source code with comments
> 
> If we use EmbedAllSources instead of EmbedUntrackedSources, stepping into source works even when offline. Package sizes do increase from a total of 1.3 MB to 1.8 MB on disk (Management.Endpoint would be the largest at 402KB).

Makes sense. With 1, adding breakpoints should still work (that's what the PDBs are for). With 2, sources are downloaded from GitHub using the embedded commit hash.

With `EmbedAllSources`, source files end up in the binary as well:

image

Once this is merged, please verify stepping into and adding breakpoints also works with NuGets in Azure DevOps. And check that code coverage still shows up.

shared-package.props
TimHess: Why remove `ContinuousIntegrationBuild`? I would think the pain around absolute/relative path differences was already resolved in all the tooling... I don't see what else it affects, but [the docs](https://learn.microsoft.com/en-us/dotnet/core/project-sdk/msbuild-props#continuousintegrationbuild) imply that isn't the only thing that changes

bart-vmware: Well, the reason for keeping it seems more of a theoretical concern (see https://github.com/dotnet/roslyn/issues/55860#issuecomment-905018564), but I'm fine leaving it in.

bart-vmware (approved):

PR #1478, opened 2025-02-24
Update content-type for mappings to v3

TimHess:
## Description

Remove the override that forced mappings label responses as v2.
Also throw an exception if SBA url wasn't set rather than trying to connect first.

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminClientHostedService.cs
+++ b/src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminClientHostedService.cs
@@ -137,7 +137,7 @@ internal sealed class SpringBootAdminClientHostedService : IHostedService
         }
         catch (Exception exception) when (!exception.IsCancellation())
         {
-            _logger.LogError(exception, "Error connecting to Spring Boot Admin Server: {Message}", exception.Message);
+            _logger.LogError(exception, "Error connecting to Spring Boot Admin Server.");
             return;
         }
 
--- a/src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminClientHostedService.cs
+++ b/src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminClientHostedService.cs
@@ -49,6 +49,11 @@ internal sealed class SpringBootAdminClientHostedService : IHostedService
     {
         SpringBootAdminClientOptions clientOptions = _clientOptionsMonitor.CurrentValue;
 
+        if (clientOptions.Url == null)
+        {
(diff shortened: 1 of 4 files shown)


TimHess:
Tested with SBA 3.4.3 built from https://github.com/SteeltoeOSS/Dockerfiles/pull/42
image

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminClientHostedService.cs
```diff
@@ -49,6 +49,11 @@ public async Task StartAsync(CancellationToken cancellationToken)
     {
         SpringBootAdminClientOptions clientOptions = _clientOptionsMonitor.CurrentValue;
 
+        if (clientOptions.Url == null)
+        {
+            throw new InvalidOperationException("Spring Boot Admin Server Url must be provided in options.");
+        }
```
bart-vmware, line 55: Is it safe to throw here? Running from a hosted service (background thread), would that crash the app or swallow the error? What's the experience when trying this from a sample?

TimHess: This change results in an app crash, which feels right (at least for a local dev) since we don't ever retry this operation and you need to restart to register.

Without the change the app runs but obviously never registers, and you see this in the logs:
```
fail: Steeltoe.Management.Endpoint.SpringBootAdminClient.SpringBootAdminClientHostedService[0]
      Error connecting to Spring Boot Admin Server: An invalid request URI was provided. Either the request URI must be an absolute URI or BaseAddress must be set.
      System.InvalidOperationException: An invalid request URI was provided. Either the request URI must be an absolute URI or BaseAddress must be set.
```

bart-vmware: The duplicate message is due to this line in `SpringBootAdminClientHostedService`:
```c#
_logger.LogError(exception, "Error connecting to Spring Boot Admin Server: {Message}", exception.Message);
```
It should just be:
```c#
_logger.LogError(exception, "Error connecting to Spring Boot Admin Server.");
```
Can you fix the logging in this file?

TimHess: Yeah, I'll fix the log message. What do you think about keeping the new exception?

bart-vmware: I'd have to try it out to see what happens. But you probably have a sample ready, so can we discuss it during standup?

PR #1484, opened 2025-02-27
Update build label to beta

TimHess:
## Description

Update build metadata

## Quality checklist

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "4.0-alpha",
+  "version": "4.0-beta",
   "publicReleaseRefSpec": [
     "^refs/heads/main$",
     "^refs/heads/release/\\d+(?:\\.\\d+)?$"


version.json
```diff
@@ -1,6 +1,6 @@
 {
   "$schema": "https://raw.githubusercontent.com/dotnet/Nerdbank.GitVersioning/master/src/NerdBank.GitVersioning/version.schema.json",
-  "version": "4.0-alpha",
+  "version": "4.0-beta",
```
bart-vmware, line 3: Should we use beta1 instead? My concern is that if we ever create a beta2, then "pkgname-beta" will be listed higher than "pkgname-beta2" on nuget.org.

TimHess: Right now this produces builds like `4.0.576-beta` -- if I understand correctly, we'll already have to use the override to get 4.0.0... I'm still new to nbgv so I'm unclear when that height resets, but if we already need to use the override for any release up to GA, maybe it's fine to leave it at beta in version.json and consider the `beta1`, `beta2` designations completely separately?

bart-vmware: It's been too long ago since I looked into height resets, I don't remember. Assuming we'll override for nuget.org, then this still affects Azure DevOps builds. Okay, let's go with this for now, but we definitely need to revisit how this works.

bart-vmware (approved):

PR #1508, opened 2025-04-16
Enhancements for Spring Boot Admin registration

### LINKED ISSUE #490: Spring Boot Admin Integration enhancements
[ConfigurationUrlHelpers](https://github.com/SteeltoeOSS/Steeltoe/blob/master/src/Discovery/src/Client/ConfigurationUrlHelpers.cs) should move to a more centralized location so they can also be used for registering with Spring Boot Admin.

Consider adding extension(s) for program.cs and/or including with general actuator setup for a more polished experience.

bart-vmware:
## Description

This PR enhances the registration process of Steeltoe apps with a Spring Boot Admin server.
- Periodically re-register (on by default), for resiliency against SBA server restarts
- New settings to override the app scheme/host/port/path to register with (for running behind a reverse proxy)
- Respond to configuration changes at runtime (unregister from previous server if URL changed)
- Improved logging of the registration process, improved test coverage
- Use the full Steeltoe version in the `User-Agent` header of outbound HTTP requests
- Enhanced detection of app URL for Spring Boot Admin
  - Take the settings for exposing management endpoints on an alternate port into account
  - Use detection logic for local host/IP similar to Eureka
  - Infer the scheme and port (not the hostname, if wildcard) from ASP.NET address bindings, prefer https if multiple
  - Works with ASP.NET dynamic port bindings

Breaking changes:
- `BasePath` configuration key was renamed to `BaseUrl` (new key `BasePath` means something else)
- Default value for `ConnectionTimeoutMs` setting reduced to 5 seconds

Fixes #490

## Quality checklist

      Documentation PR: https://github.com/SteeltoeOSS/Documentation/pull/371
      Samples PR: https://github.com/SteeltoeOSS/Samples/pull/387


src/Common/src/Common/Net/IDomainNameResolver.cs
```diff
@@ -2,9 +2,11 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
-namespace Steeltoe.Management.Endpoint.SpringBootAdminClient;
+namespace Steeltoe.Common.Net;
 
-internal sealed class RegistrationResult
+internal interface IDomainNameResolver
```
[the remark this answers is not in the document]
bart-vmware: Well, it's a weird thing whose logic I didn't want to change. The first method, `ResolveHostAddress` takes a `hostName` parameter and calls `Dns.GetHostAddresses(hostName`), which is documented as:

image

The second method, `ResolveHostName`, applies to the local host only.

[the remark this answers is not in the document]
bart-vmware: So it's not always local-only. But also not always a domain name. I struggled to find a good name, not sure what would be better.

src/Management/test/Endpoint.Test/SpringBootAdminClient/AppUrlCalculatorTest.cs
```diff
@@ -0,0 +1,383 @@
...
+        };
+
+        IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(appSettings).Build();
+        using ServiceProvider serviceProvider = BuildServiceProvider(configuration);
+
+        var calculator = serviceProvider.GetRequiredService<AppUrlCalculator>();
+        var options = serviceProvider.GetRequiredService<IOptions<SpringBootAdminClientOptions>>();
+        string? url = calculator.AutoDetectAppUrl(options.Value);
+
+        url.Should().Be($"https://{FakeDomainNameResolver.HostName}:{ListenSecurePort1}");
```
TimHess, line 54: Maybe I'm missing something, but I would think that since the value in `urls` is not a wildcard the hostname should be "dontcare"... Won't this cause additional certificate issues? -- We know SBA in a container already didn't trust the ASP.NET Core cert, and can't reach `localhost` anyway, so maybe that's not a great example, but now the host also won't match

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminRefreshRunner.cs
```diff
@@ -0,0 +1,184 @@
...
+            catch (Exception exception)
+            {
+                if (!exception.IsCancellation())
+                {
+                    _logger.LogWarning(exception, "Failed to unregister from Spring Boot Admin server at {Url}.", clientOptions.Url);
+                }
+            }
+            finally
+            {
+                _lastRegistrationId = null;
```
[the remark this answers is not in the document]
bart-vmware: That would introduce a bug. If `CleanupAsync` is called multiple times, and the first call fails (server temporarily unreachable), we still want to be able to unregister the second time. Setting `_lastGoodOptions` to `null` makes that impossible.

src/Management/test/Endpoint.Test/SpringBootAdminClient/AppUrlCalculatorTest.cs
bart-vmware: This test case is intentional, to express any hostname is meaningless. See the docs at https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-9.0#url-formats:

> Anything not recognized as a valid IP address or localhost is treated as a wildcard that binds to all IPv4 and IPv6 addresses. Some people like to use * or + to be more explicit.

The hostname has no meaning in ASP.NET, so it would be wrong if it meant something in Steeltoe. Users wouldn't expect that.

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminRefreshRunner.cs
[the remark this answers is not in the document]
bart-vmware: I've moved `_lastRegistrationId = null` outside the `finally` clause, although this is effectively unreachable now that `StopAsync` won't clean up multiple times.

src/Management/test/Endpoint.Test/SpringBootAdminClient/AppUrlCalculatorTest.cs
TimHess: Huh, based on the different docs it looks like that changed in 8.0, I either didn't know that or forgot about it...
In any case, it looks like localhost still gets special treatment in ASP.NET Core, but based on the tests we would register the actual hostname with SBA, that seems like it shouldn't work, but does it?

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        _logger.LogDebug("Starting cleanup.");
+        await _runner.CleanupAsync(cancellationToken);
+    }
+
+    public async ValueTask DisposeAsync()
+    {
+        if (!_isDisposed)
+        {
+            _isDisposed = true;
```
TimHess, line 98: Nothing jumps out at me as likely to break here, but shouldn't `_isDisposed` be set _after_ everything is disposed

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminRefreshRunner.cs
```diff
@@ -0,0 +1,184 @@
...
+        options.ApplicationName ??= _applicationInstanceInfo.ApplicationName;
+        options.BaseUrl ??= _appUrlCalculator.AutoDetectAppUrl(options);
+
+        if (string.IsNullOrEmpty(options.BaseUrl))
+        {
+            errors.Add($"{nameof(SpringBootAdminClientOptions.BaseUrl)} must be configured to register with Spring Boot Admin server");
+        }
+        else if (!Uri.IsWellFormedUriString(options.BaseUrl, UriKind.Absolute))
+        {
+            errors.Add($"{nameof(SpringBootAdminClientOptions.BaseUrl)} must be configured as a fully-qualified URL to register with Spring Boot Admin server");
```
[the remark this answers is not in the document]
bart-vmware: I'll change fully-qualified to absolute.

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
bart-vmware: The problem with that is when an exception is thrown half-way, we don't know which line caused it. So the first fields may already have been disposed, while the last lines haven't executed yet. Running dispose again would try to re-run code, potentially resulting in an exception because the object is already disposed. For example, the call to `await _timerTokenSource.CancelAsync()` throws `ObjectDisposedException`. So in that case, there's no way to recover. By setting the flag in advance, a second call to Dispose won't throw. The misery remains, but it's a little bit better because calling dispose multiple times shouldn't throw.

The guidelines are:
- Calling Dispose multiple times should not throw.
- Do not call Dispose multiple times.

src/Management/test/Endpoint.Test/SpringBootAdminClient/AppUrlCalculatorTest.cs
bart-vmware: ASP.NET has always worked like that, as far as I'm aware. The [documentation for .NET 5](https://learn.microsoft.com/en-us/aspnet/core/fundamentals/servers/kestrel/endpoints?view=aspnetcore-5.0#url-formats) uses slightly different wording:
> Host names, *, and +, aren't special. Anything not recognized as a valid IP address or localhost binds to all IPv4 and IPv6 IPs.

I'm going to change the logic so that `localhost` and IP addresses from bindings are preferred over resolving the local host/IP.

bart-vmware:
The Config Server flaky test fix is based on the following log fragment:
```
Steeltoe.Configuration.ConfigServer.Test.ConfigServerConfigurationProviderTest.Load_ConfigServerReturnsBadStatus_FailFastEnabled_RetryEnabled

Assert.Throws() Failure: Exception type was not an exact match
Expected: typeof(Steeltoe.Configuration.ConfigServer.ConfigServerException)
Actual:   typeof(System.Threading.Tasks.TaskCanceledException)


Captured logs:
DBUG Steeltoe.Configuration.ConfigServer.ConfigServerConfigurationProvider: Fetching configuration from server at: http://localhost:8888/
TRCE Steeltoe.Configuration.ConfigServer.ConfigServerConfigurationProvider: Processing label ''
TRCE Steeltoe.Configuration.ConfigServer.ConfigServerConfigurationProvider: Entered RemoteLoadAsync
TRCE Steeltoe.Configuration.ConfigServer.ConfigServerConfigurationProvider: Building HTTP request message
TRCE Steeltoe.Configuration.ConfigServer.ConfigServerConfigurationProvider: Sending HTTP request


---- Assert.Throws() Failure: Exception type was not an exact match
Expected: typeof(Steeltoe.Configuration.ConfigServer.ConfigServerException)
Actual:   typeof(System.Threading.Tasks.TaskCanceledException)
-------- System.Threading.Tasks.TaskCanceledException : The operation was canceled.
------------ System.Net.Http.HttpRequestException : Error while copying content to a stream.
---------------- System.IO.IOException :
-------------------- System.IO.IOException : The client aborted the request.
```

The new diagnostics are starting to prove helpful. A 10 ms timeout is way too short, so I changed it to 1 sec. The other test is supposed to trigger a timeout, so it can be shortened.

TimHess (approved):
LGTM!

PR #1520, opened 2025-05-28
Improve management test coverage

bart-vmware:
## Description

{A detailed description}

Fixes #{bug number}

## Quality checklist


src/Management/src/Endpoint/Actuators/Environment/EnvironmentEndpointHandler.cs
```diff
@@ -36,29 +36,33 @@ public EnvironmentEndpointHandler(IOptionsMonitor<EnvironmentEndpointOptions> op
...
+    private List<PropertySourceDescriptor> GetPropertySources()
     {
         List<PropertySourceDescriptor> results = [];
 
         foreach (IConfigurationProvider provider in _configuration.EnumerateProviders())
         {
             PropertySourceDescriptor descriptor = GetPropertySourceDescriptor(provider);
-            results.Add(descriptor);
+
+            if (descriptor.Properties.Count > 0)
```
TimHess: Should we add a switch to toggle the behavior here to allow returning the empty sources too? I'm thinking about a scenario where you're trying to diagnose why a config value is not what you expected, and the root cause is some provider is empty (and is now also missing from the response here)

src/Management/test/Endpoint.Test/garbage.git.properties
```diff
@@ -1,3 +1,7 @@
-﻿#Generated by Git-Commit-Id-Plugin
+#Generated by Git-Commit-Id-Plugin
 #Wed Jul 12 12:40:39 MDT 2017
-foobar.build.user.email=someone@testdomain.com
\ No newline at end of file
+foobar.build.user.email=someone@testdomain.com
```
TimHess: Is this change to line endings expected, and/or significant enough to warrant a change to `.gitattributes`?

src/Management/src/Endpoint/Actuators/Environment/EnvironmentEndpointHandler.cs
bart-vmware: The [Spring Boot env actuator](https://docs.spring.io/spring-boot/api/rest/actuator/env.html) includes sources without any properties, so we should probably revert this change entirely.

src/Management/test/Endpoint.Test/Actuators/All/AllActuatorsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+using Steeltoe.Management.Endpoint.ManagementPort;
+using Steeltoe.Management.Endpoint.Middleware;
+
+namespace Steeltoe.Management.Endpoint.Test.Actuators.All;
+
+public sealed class AllActuatorsTest
+{
+    [Theory]
+    [InlineData(false)]
+    [InlineData(true)]
```
TimHess, line 36: Hm... I don't think it functionally matters, but I've been putting the theory data above the `[Theory]` attribute so that Theory/Fact is always the line just above the method. Probably fits in the same conceptual bucket as whether or not to use underscores in test names and would be nice (but not essential) to be more consistent

src/Management/test/Endpoint.Test/Actuators/All/AllActuatorsTest.cs
```diff
@@ -0,0 +1,134 @@
...
+            // ReSharper disable once AccessToDisposedClosure
+            Func<object> action = () => serviceProvider.GetRequiredService(middlewareType);
+            action.Should().NotThrow();
+        }
+
+        if (platformIsCloudFoundry)
+        {
+            Action action = () =>
+            {
+                // ReSharper disable AccessToDisposedClosure
```
TimHess, line 74: Would it make sense to move this out/up a couple levels to allow remove the other suppression of this rule in this test?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundryActuatorTest.cs
```diff
@@ -0,0 +1,562 @@
...
+                  "href": "http://localhost/cloudfoundryapplication",
+                  "templated": false
+                }
+              }
+            }
+            """);
+    }
+
+    [Fact]
+    public async Task Returns_error_when_not_running_on_Cloud_Foundry()
```
TimHess, line 217:
```suggestion
    public async Task Returns_404_when_not_running_on_Cloud_Foundry()
```
Or something along these lines, to differentiate from server error or some other active-denial condition

src/Management/test/Endpoint.Test/Actuators/Loggers/LoggersActuatorTest.cs
```diff
@@ -0,0 +1,616 @@
...
+              "groups": {}
+            }
+            """);
+    }
+
+    [Fact]
+    public async Task Can_change_configuration_at_runtime()
+    {
+        var fileProvider = new MemoryFileProvider();
+        const string appSettingsJsonFileName = "appsettings.json";
```
TimHess: Should we consider centralizing constants like this in TestResources?

src/Management/test/Endpoint.Test/Actuators/Refresh/TestConfigurationSource.cs
```diff
@@ -0,0 +1,34 @@
...
+    private sealed class TestConfigurationProvider : ConfigurationProvider
+    {
+        private int _loadCount;
+
+        public override void Load()
+        {
+            base.Load();
+
+            int newLoadCount = Interlocked.Increment(ref _loadCount);
+            Data["FakeLoadCount"] = newLoadCount.ToString(CultureInfo.InvariantCulture);
```
TimHess, line 31: "FakeLoadCount" = "How many times this fake was loaded", right?
What about "ConfigurationLoadCount" or even just "LoadCount" since it is a real count of how many times Load was called, and the fact it's inside a fake is (probably?) not super important

TimHess (approved):
Generally LGTM, just a couple small items that could be addressed

src/Management/test/Endpoint.Test/garbage.git.properties
bart-vmware: I didn't notice. Pushed a commit to fix it. Downloaded the branch as a zip file from GitHub (so that auto conversion doesn't kick in), this is what it looks like there now:

image

src/Management/test/Endpoint.Test/Actuators/All/AllActuatorsTest.cs
bart-vmware: It reads more naturally to me this way: "Here's a compound test, and these are the variations..."

src/Management/test/Endpoint.Test/Actuators/All/AllActuatorsTest.cs
bart-vmware: There are only three occurrences, so it wouldn't make a significant difference. By moving it up, we lose the potentially useful diagnostic.

I did wonder if FluentAssertions can be expanded to indicate "this executes immediately" using `[InstantHandle]` (see [here](https://www.jetbrains.com/help/resharper/AccessToDisposedClosure.html), but I never followed up on that.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundryActuatorTest.cs
bart-vmware: I intentionally didn't do so, because the 404 is already in the test body. By duplicating it in the name, it easily gets out of sync. The essence of the test is that the outcome is non-successful.

src/Management/test/Endpoint.Test/Actuators/Loggers/LoggersActuatorTest.cs
bart-vmware: I think it would make more sense to expose it as a public constant on `MemoryFileProvider`. Are you okay with that?

TimHess: Sure. It's a pretty low-priority change IMO, just thinking there are probably a bunch strings that we could do this with for both tests and even the framework code (like all the CF env vars)

src/Management/test/Endpoint.Test/garbage.git.properties
TimHess: Heh, now the other one is changed - but at least things should be consistent now

src/Management/test/Endpoint.Test/Actuators/Loggers/LoggersActuatorTest.cs
bart-vmware: I've pushed the change on `MemoryFileProvider`, will leave other potential candidates to be addressed in the future.

src/Management/test/Endpoint.Test/Actuators/Refresh/TestConfigurationSource.cs
bart-vmware: It's the Fake provider, tracking how many times its Load method was called. This doesn't apply to every provider. For example, ConfigServer reloads itself periodically, so its count would be higher.

TimHess: I forgot to complete my original thought that it might be useful to use the same thing or kind of thing with Placeholder and/or other wrapping config providers, but there's no need to make additional changes now, this is fine

PR #1521, opened 2025-05-29
Align Cloud Controller response handling with Spring, improve tests

### LINKED ISSUE #1446: Is the 503 status code correct when the request to check permissions in CF fails
Follow-up from https://github.com/SteeltoeOSS/Steeltoe/pull/1444/files#r1920383965 - double check what Spring Boot does

TimHess:
## Description

To answer the original question: when using Spring Boot as the spec for these interactions, HTTP 503 can be the "correct" response.

This PR:
- Adds testing of various potential cloud controller responses
- Allows usage of 401 response code (all 4xx errors besides 403) | [relevant Spring code](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveCloudFoundrySecurityService.java#L103)
- Allows usage of 500 response code, with error message in response | [relevant Spring code](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/CloudFoundrySecurityInterceptor.java#L95)

Resolves #1446 

## Quality checklist

### DIFF
--- a/src/Common/src/Common/Extensions/ExceptionExtensions.cs
+++ b/src/Common/src/Common/Extensions/ExceptionExtensions.cs
@@ -66,7 +66,7 @@ internal static class ExceptionExtensions
     /// <param name="exception">
     /// The caught exception to inspect.
     /// </param>
-    public static bool IsTimeout(this Exception? exception)
+    public static bool IsHttpClientTimeout(this Exception? exception)
     {
         // See note in remarks at https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync.
         if (exception is OperationCanceledException && exception.InnerException?.GetType() == typeof(TimeoutException))
--- a/src/Common/src/Common/Extensions/ExceptionExtensions.cs
+++ b/src/Common/src/Common/Extensions/ExceptionExtensions.cs
@@ -59,4 +59,21 @@ internal static class ExceptionExtensions
 
         return false;
     }
+
+    /// <summary>
+    /// Determines whether the thrown exception results from an HTTP request timeout.
+    /// </summary>
+    /// <param name="exception">
+    /// The caught exception to inspect.
+    /// </param>
+    public static bool IsTimeout(this Exception? exception)
+    {
+        // See note in remarks at https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync.
+        if (exception is OperationCanceledException && exception.InnerException?.GetType() == typeof(TimeoutException))
+        {
+            return true;
+        }
+
+        return false;
+    }
 }
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
@@ -71,7 +71,7 @@ internal sealed class CloudFoundrySecurityMiddleware
 
             if (string.IsNullOrEmpty(endpointOptions.Api))
             {
-                await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.CloudfoundryApiMissing));
+                await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.CloudFoundryApiMissing));
                 return;
             }
 
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
@@ -164,7 +164,9 @@ internal sealed class CloudFoundrySecurityMiddleware
         _logger.LogError("Actuator Security Error: {Code} - {Message}", error.Code, error.Message);
         context.Response.Headers.Append("Content-Type", "application/json;charset=UTF-8");
 
-        // allowing override of 400-level errors is more likely to cause confusion than to be useful
+        // UseStatusCodeFromResponse was added to prevent IIS/HWC from blocking the response body on 500-level errors.
+        // Blocking 400-level error responses would be more likely to cause confusion than to be useful.
+        // See https://github.com/SteeltoeOSS/Steeltoe/issues/418 for more information.
         if (_managementOptionsMonitor.CurrentValue.UseStatusCodeFromResponse || (int)error.Code < 500)
         {
             context.Response.StatusCode = (int)error.Code;
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
@@ -10,6 +10,7 @@ using Microsoft.AspNetCore.Http;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Microsoft.Extensions.Primitives;
+using Microsoft.Net.Http.Headers;
 using Steeltoe.Common;
 using Steeltoe.Management.Configuration;
 using Steeltoe.Management.Endpoint.Actuators.Hypermedia;
@@ -64,13 +65,13 @@ internal sealed class CloudFoundrySecurityMiddleware
                 _logger.LogCritical(
                     "The Application Id could not be found. Make sure the Cloud Foundry Configuration Provider has been added to the application configuration.");
 
-                await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.ApplicationIdMissingMessage));
+                await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.ApplicationIdMissing));
                 return;
             }
 
             if (string.IsNullOrEmpty(endpointOptions.Api))
             {
-                await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.CloudfoundryApiMissingMessage));
+                await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.CloudfoundryApiMissing));
                 return;
             }
 
@@ -88,7 +89,7 @@ internal sealed class CloudFoundrySecurityMiddleware
 
                 if (targetEndpointOptions.RequiredPermissions > givenPermissions.Permissions)
                 {
-                    await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.Forbidden, PermissionsProvider.AccessDeniedMessage));
+                    await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.Forbidden, PermissionsProvider.Messages.AccessDenied));
                     return;
                 }
             }
@@ -102,7 +103,7 @@ internal sealed class CloudFoundrySecurityMiddleware
 
     internal string GetAccessToken(HttpRequest request)
     {
-        if (request.Headers.TryGetValue(PermissionsProvider.AuthorizationHeaderName, out StringValues authorizationHeaderValue))
+        if (request.Headers.TryGetValue(HeaderNames.Authorization, out StringValues authorizationHeaderValue))
         {
             string authorizationValue = authorizationHeaderValue.ToString();
 
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
@@ -19,13 +19,13 @@ namespace Steeltoe.Management.Endpoint.Actuators.CloudFoundry;
 
 internal sealed class CloudFoundrySecurityMiddleware
 {
+    private const string BearerTokenPrefix = "Bearer ";
     private readonly IOptionsMonitor<ManagementOptions> _managementOptionsMonitor;
     private readonly IOptionsMonitor<CloudFoundryEndpointOptions> _endpointOptionsMonitor;
     private readonly IEndpointOptionsMonitorProvider[] _endpointOptionsMonitorProviderArray;
     private readonly RequestDelegate? _next;
     private readonly ILogger<CloudFoundrySecurityMiddleware> _logger;
     private readonly PermissionsProvider _permissionsProvider;
-    private const string BearerTokenPrefix = "Bearer ";
 
     public CloudFoundrySecurityMiddleware(IOptionsMonitor<ManagementOptions> managementOptionsMonitor,
         IOptionsMonitor<CloudFoundryEndpointOptions> endpointOptionsMonitor, IEnumerable<IEndpointOptionsMonitorProvider> endpointOptionsMonitorProviders,
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
@@ -62,7 +62,7 @@ internal sealed class CloudFoundrySecurityMiddleware
         {
             if (string.IsNullOrEmpty(endpointOptions.ApplicationId))
             {
-                _logger.LogCritical(
+                _logger.LogError(
                     "The Application Id could not be found. Make sure the Cloud Foundry Configuration Provider has been added to the application configuration.");
 
                 await ReturnErrorAsync(context, new SecurityResult(HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.ApplicationIdMissing));
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
@@ -25,6 +25,7 @@ internal sealed class CloudFoundrySecurityMiddleware
     private readonly RequestDelegate? _next;
     private readonly ILogger<CloudFoundrySecurityMiddleware> _logger;
     private readonly PermissionsProvider _permissionsProvider;
+    private const string BearerTokenPrefix = "Bearer ";
 
     public CloudFoundrySecurityMiddleware(IOptionsMonitor<ManagementOptions> managementOptionsMonitor,
         IOptionsMonitor<CloudFoundryEndpointOptions> endpointOptionsMonitor, IEnumerable<IEndpointOptionsMonitorProvider> endpointOptionsMonitorProviders,
@@ -101,13 +102,13 @@ internal sealed class CloudFoundrySecurityMiddleware
 
     internal string GetAccessToken(HttpRequest request)
     {
-        if (request.Headers.TryGetValue(PermissionsProvider.AuthorizationHeaderName, out StringValues headerValue))
+        if (request.Headers.TryGetValue(PermissionsProvider.AuthorizationHeaderName, out StringValues authorizationHeaderValue))
         {
-            string header = headerValue.ToString();
+            string authorizationValue = authorizationHeaderValue.ToString();
 
-            if (header.StartsWith(PermissionsProvider.BearerHeaderNamePrefix, StringComparison.OrdinalIgnoreCase))
+            if (authorizationValue.StartsWith(BearerTokenPrefix, StringComparison.OrdinalIgnoreCase))
             {
-                return header[PermissionsProvider.BearerHeaderNamePrefix.Length..];
+                return authorizationValue[BearerTokenPrefix.Length..];
             }
         }
 
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
@@ -83,8 +83,6 @@ internal sealed class PermissionsProvider
         }
         catch (Exception exception) when (exception.IsHttpClientTimeout())
         {
-            _logger.LogWarning(exception, "Cloud Foundry request timed out while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
-
             return new SecurityResult(HttpStatusCode.ServiceUnavailable, Messages.CloudFoundryTimeout);
         }
     }
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
@@ -81,15 +81,8 @@ internal sealed class PermissionsProvider
             EndpointPermissions permissions = await ParsePermissionsResponseAsync(response, cancellationToken);
             return new SecurityResult(permissions);
         }
-        catch (Exception exception)
+        catch (Exception exception) when (!exception.IsCancellation())
         {
-            if (exception.IsCancellation())
-            {
-                _logger.LogTrace(exception, "Task cancelled or timed out while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
-
-                throw;
-            }
-
             _logger.LogError(exception, "Cloud Foundry returned exception while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
 
             return new SecurityResult(HttpStatusCode.InternalServerError, exception.Message);
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
@@ -19,16 +19,16 @@ namespace Steeltoe.Management.Endpoint.Actuators.CloudFoundry;
 
 internal sealed class PermissionsProvider
 {
-    private const string AuthorizationHeaderInvalid = "Authorization header is missing or invalid";
-    private const string CloudfoundryNotReachableMessage = "Cloud controller not reachable";
     private const string ReadSensitiveDataJsonPropertyName = "read_sensitive_data";
-    public const string HttpClientName = "CloudFoundrySecurity";
-    public const string ApplicationIdMissingMessage = "Application ID is not available";
-    public const string CloudfoundryApiMissingMessage = "Cloud controller URL is not available";
     public const string AccessDeniedMessage = "Access denied";
+    public const string ApplicationIdMissingMessage = "Application ID is not available";
     public const string AuthorizationHeaderName = "Authorization";
-    public const string BearerHeaderNamePrefix = "Bearer ";
-    private static readonly TimeSpan GetPermissionsTimeout = TimeSpan.FromMilliseconds(5000);
+    public const string AuthorizationHeaderInvalid = "Authorization header is missing or invalid";
+    public const string CloudfoundryApiMissingMessage = "Cloud controller URL is not available";
+    public const string CloudfoundryNotReachableMessage = "Cloud controller not reachable";
+    public const string HttpClientName = "CloudFoundrySecurity";
+    public const string InvalidTokenMessage = "Invalid token";
+    private static readonly TimeSpan GetPermissionsTimeout = TimeSpan.FromMilliseconds(5_000);
 
     private readonly IOptionsMonitor<CloudFoundryEndpointOptions> _optionsMonitor;
     private readonly IHttpClientFactory _httpClientFactory;
@@ -75,23 +75,34 @@ internal sealed class PermissionsProvider
                 _logger.LogInformation("Cloud Foundry returned status: {HttpStatus} while obtaining permissions from: {PermissionsUri}", response.StatusCode,
                     checkPermissionsUri);
 
-                return response.StatusCode is HttpStatusCode.Forbidden or HttpStatusCode.Unauthorized
-                    ? new SecurityResult(HttpStatusCode.Forbidden, AccessDeniedMessage)
+                if (response.StatusCode is HttpStatusCode.Forbidden)
+                {
+                    return new SecurityResult(HttpStatusCode.Forbidden, AccessDeniedMessage);
+                }
+
+                return (int)response.StatusCode is > 399 and < 500
+                    ? new SecurityResult(HttpStatusCode.Unauthorized, InvalidTokenMessage)
                     : new SecurityResult(HttpStatusCode.ServiceUnavailable, CloudfoundryNotReachableMessage);
             }
 
-            EndpointPermissions permissions = await GetPermissionsAsync(response, cancellationToken);
+            EndpointPermissions permissions = await ParsePermissionsAsync(response, cancellationToken);
             return new SecurityResult(permissions);
         }
         catch (Exception exception) when (!exception.IsCancellation())
         {
             _logger.LogError(exception, "Cloud Foundry returned exception while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
 
+            return new SecurityResult(HttpStatusCode.InternalServerError, exception.Message);
+        }
+        catch (Exception exception)
+        {
+            _logger.LogInformation(exception, "Task cancelled or timed out while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
+
             return new SecurityResult(HttpStatusCode.ServiceUnavailable, CloudfoundryNotReachableMessage);
         }
     }
 
-    public async Task<EndpointPermissions> GetPermissionsAsync(HttpResponseMessage response, CancellationToken cancellationToken)
+    public async Task<EndpointPermissions> ParsePermissionsAsync(HttpResponseMessage response, CancellationToken cancellationToken)
     {
         ArgumentNullException.ThrowIfNull(response);
 
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
@@ -94,12 +94,6 @@ internal sealed class PermissionsProvider
 
             return new SecurityResult(HttpStatusCode.InternalServerError, exception.Message);
         }
-        catch (Exception exception)
-        {
-            _logger.LogInformation(exception, "Task cancelled or timed out while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
-
-            return new SecurityResult(HttpStatusCode.ServiceUnavailable, CloudfoundryNotReachableMessage);
-        }
     }
 
     public async Task<EndpointPermissions> ParsePermissionsAsync(HttpResponseMessage response, CancellationToken cancellationToken)
--- a/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
+++ b/src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
@@ -20,14 +20,7 @@ namespace Steeltoe.Management.Endpoint.Actuators.CloudFoundry;
 internal sealed class PermissionsProvider
 {
     private const string ReadSensitiveDataJsonPropertyName = "read_sensitive_data";
-    public const string AccessDeniedMessage = "Access denied";
-    public const string ApplicationIdMissingMessage = "Application ID is not available";
-    public const string AuthorizationHeaderName = "Authorization";
-    public const string AuthorizationHeaderInvalid = "Authorization header is missing or invalid";
-    public const string CloudfoundryApiMissingMessage = "Cloud controller URL is not available";
-    public const string CloudfoundryNotReachableMessage = "Cloud controller not reachable";
     public const string HttpClientName = "CloudFoundrySecurity";
-    public const string InvalidTokenMessage = "Invalid token";
     private static readonly TimeSpan GetPermissionsTimeout = TimeSpan.FromMilliseconds(5_000);
 
     private readonly IOptionsMonitor<CloudFoundryEndpointOptions> _optionsMonitor;
@@ -55,7 +48,7 @@ internal sealed class PermissionsProvider
     {
         if (string.IsNullOrEmpty(accessToken))
         {
-            return new SecurityResult(HttpStatusCode.Unauthorized, AuthorizationHeaderInvalid);
+            return new SecurityResult(HttpStatusCode.Unauthorized, Messages.AuthorizationHeaderInvalid);
         }
 
         CloudFoundryEndpointOptions options = _optionsMonitor.CurrentValue;
@@ -77,15 +70,15 @@ internal sealed class PermissionsProvider
 
                 if (response.StatusCode is HttpStatusCode.Forbidden)
                 {
-                    return new SecurityResult(HttpStatusCode.Forbidden, AccessDeniedMessage);
+                    return new SecurityResult(HttpStatusCode.Forbidden, Messages.AccessDenied);
                 }
 
                 return (int)response.StatusCode is > 399 and < 500
-                    ? new SecurityResult(HttpStatusCode.Unauthorized, InvalidTokenMessage)
-                    : new SecurityResult(HttpStatusCode.ServiceUnavailable, CloudfoundryNotReachableMessage);
+                    ? new SecurityResult(HttpStatusCode.Unauthorized, Messages.InvalidToken)
+                    : new SecurityResult(HttpStatusCode.ServiceUnavailable, Messages.CloudfoundryNotReachable);
             }
 
-            EndpointPermissions permissions = await ParsePermissionsAsync(response, cancellationToken);
+            EndpointPermissions permissions = await ParsePermissionsResponseAsync(response, cancellationToken);
             return new SecurityResult(permissions);
         }
         catch (Exception exception)
@@ -103,7 +96,7 @@ internal sealed class PermissionsProvider
         }
     }
 
-    public async Task<EndpointPermissions> ParsePermissionsAsync(HttpResponseMessage response, CancellationToken cancellationToken)
+    public async Task<EndpointPermissions> ParsePermissionsResponseAsync(HttpResponseMessage response, CancellationToken cancellationToken)
     {
         ArgumentNullException.ThrowIfNull(response);
 
@@ -138,5 +131,15 @@ internal sealed class PermissionsProvider
         HttpClient httpClient = _httpClientFactory.CreateClient(HttpClientName);
         httpClient.ConfigureForSteeltoe(GetPermissionsTimeout);
         return httpClient;
+    }
+
+    internal static class Messages
+    {
+        public const string AccessDenied = "Access denied";
+        public const string ApplicationIdMissing = "Application ID is not available";
+        public const string AuthorizationHeaderInvalid = "Authorization header is missing or invalid";
+        public const string CloudfoundryApiMissing = "Cloud controller URL is not available";
+        public const string CloudfoundryNotReachable = "Cloud controller not reachable";
+        public const string InvalidToken = "Invalid token";
     }
 }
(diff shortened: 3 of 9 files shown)


src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -75,23 +75,34 @@ public async Task<SecurityResult> GetPermissionsAsync(string accessToken, Cancel
...
         catch (Exception exception) when (!exception.IsCancellation())
         {
             _logger.LogError(exception, "Cloud Foundry returned exception while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
 
+            return new SecurityResult(HttpStatusCode.InternalServerError, exception.Message);
+        }
+        catch (Exception exception)
+        {
+            _logger.LogInformation(exception, "Task cancelled or timed out while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
+
```
bart-vmware, line 100: This catch-all should be removed, so that the cancellation bubbles up and terminates the entire pipeline.

While it's useful to log a timeout, a client walking away because they closed their browser, or a consuming machine that restarts, should not pollute log files, as this is normal and expected behavior.

I would expect a true timeout while connecting to the auth server to bubble up and be caught and logged by ASP.NET itself, similar to any other exception type.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -365,6 +365,72 @@ public async Task Redacts_HTTP_headers()
...
+    [InlineData("forbidden", HttpStatusCode.Forbidden, PermissionsProvider.AccessDeniedMessage, "true")]
+    [InlineData("forbidden", HttpStatusCode.Forbidden, PermissionsProvider.AccessDeniedMessage, "false")]
+    [InlineData("timeout", HttpStatusCode.ServiceUnavailable, PermissionsProvider.CloudfoundryNotReachableMessage, "true")]
+    [InlineData("timeout", HttpStatusCode.OK, PermissionsProvider.CloudfoundryNotReachableMessage, "false")]
+    [InlineData("exception", HttpStatusCode.InternalServerError, @"Exception of type \u0027System.Net.Http.HttpRequestException\u0027 was thrown.", "true")]
+    [InlineData("exception", HttpStatusCode.OK, @"Exception of type \u0027System.Net.Http.HttpRequestException\u0027 was thrown.", "false")]
+    [InlineData("no_sensitive_data", HttpStatusCode.OK, null, "true")]
+    [InlineData("success", HttpStatusCode.OK, null, "true")]
+    [Theory]
+    public async Task CloudControllerScenarioTesting(string cloudControllerResponse, HttpStatusCode steeltoeStatusCode, string? errorMessage,
```
bart-vmware, line 383: Please add an assertion on what's logged, similar to the first test [here](https://github.com/SteeltoeOSS/Steeltoe/blob/d306c94e067a4a6df4e3aa5135c84464e1518a76/src/Management/test/Endpoint.Test/Actuators/ThreadDump/EventPipeThreadDumperTest.cs#L26).

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -19,16 +19,16 @@ namespace Steeltoe.Management.Endpoint.Actuators.CloudFoundry;
...
 {
-    private const string AuthorizationHeaderInvalid = "Authorization header is missing or invalid";
-    private const string CloudfoundryNotReachableMessage = "Cloud controller not reachable";
     private const string ReadSensitiveDataJsonPropertyName = "read_sensitive_data";
-    public const string HttpClientName = "CloudFoundrySecurity";
-    public const string ApplicationIdMissingMessage = "Application ID is not available";
-    public const string CloudfoundryApiMissingMessage = "Cloud controller URL is not available";
     public const string AccessDeniedMessage = "Access denied";
+    public const string ApplicationIdMissingMessage = "Application ID is not available";
     public const string AuthorizationHeaderName = "Authorization";
```
bart-vmware: This constant can be removed. Callers can use `Microsoft.Net.Http.Headers.HeaderNames.Authorization`.

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -19,16 +19,16 @@ namespace Steeltoe.Management.Endpoint.Actuators.CloudFoundry;
...
-    public const string BearerHeaderNamePrefix = "Bearer ";
-    private static readonly TimeSpan GetPermissionsTimeout = TimeSpan.FromMilliseconds(5000);
+    public const string AuthorizationHeaderInvalid = "Authorization header is missing or invalid";
```
bart-vmware: Should be renamed to `AuthorizationHeaderInvalidMessage`.

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -19,16 +19,16 @@ namespace Steeltoe.Management.Endpoint.Actuators.CloudFoundry;
...
+    public const string CloudfoundryApiMissingMessage = "Cloud controller URL is not available";
+    public const string CloudfoundryNotReachableMessage = "Cloud controller not reachable";
+    public const string HttpClientName = "CloudFoundrySecurity";
```
bart-vmware, line 23: Please move this one above the list of messages.

Even better would be to introduce a nested static class `Messages` to keep them together. Then drop the "Message" suffix from the constant names.

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -75,23 +75,35 @@ public async Task<SecurityResult> GetPermissionsAsync(string accessToken, Cancel
...
+
             _logger.LogError(exception, "Cloud Foundry returned exception while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
 
-            return new SecurityResult(HttpStatusCode.ServiceUnavailable, CloudfoundryNotReachableMessage);
+            return new SecurityResult(HttpStatusCode.InternalServerError, exception.Message);
         }
     }
 
-    public async Task<EndpointPermissions> GetPermissionsAsync(HttpResponseMessage response, CancellationToken cancellationToken)
+    public async Task<EndpointPermissions> ParsePermissionsAsync(HttpResponseMessage response, CancellationToken cancellationToken)
```
bart-vmware: Consider renaming this to `ParsePermissionsResponseAsync`

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -75,23 +75,35 @@ public async Task<SecurityResult> GetPermissionsAsync(string accessToken, Cancel
...
+            {
+                _logger.LogTrace(exception, "Task cancelled or timed out while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
+
+                throw;
+            }
+
             _logger.LogError(exception, "Cloud Foundry returned exception while obtaining permissions from: {PermissionsUri}", checkPermissionsUri);
 
-            return new SecurityResult(HttpStatusCode.ServiceUnavailable, CloudfoundryNotReachableMessage);
+            return new SecurityResult(HttpStatusCode.InternalServerError, exception.Message);
```
bart-vmware: Would it be better to just let this throw? Then, depending on the environment, the exception stack trace is included or not.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -24,6 +24,25 @@ namespace Steeltoe.Management.Endpoint.Test.Actuators.CloudFoundry;
...
+    private const string CFForbiddenLog =
+        "INFO Steeltoe.Management.Endpoint.Actuators.CloudFoundry.PermissionsProvider: Cloud Foundry returned status: Forbidden while obtaining permissions from: https://example.api.com/v2/apps/forbidden/permissions";
+
+    private const string CFExceptionLog =
+        "FAIL Steeltoe.Management.Endpoint.Actuators.CloudFoundry.PermissionsProvider: Cloud Foundry returned exception while obtaining permissions from: https://example.api.com/v2/apps/exception/permissions";
+
+    private const string CFTimeoutLog =
+        "TRCE Steeltoe.Management.Endpoint.Actuators.CloudFoundry.PermissionsProvider: Task cancelled or timed out while obtaining permissions from: https://example.api.com/v2/apps/timeout/permissions";
+
+    private const string MiddlewareForbiddenLog =
```
bart-vmware: Please use `typeof(T).FullName` instead of the explicit full type.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -365,6 +384,88 @@ public async Task Redacts_HTTP_headers()
...
+            if (!string.IsNullOrEmpty(errorMessage))
+            {
+                string errorText = await response.Content.ReadAsStringAsync(TestContext.Current.CancellationToken);
+                errorText.Should().Be($$"""{"security_error":"{{errorMessage}}"}""");
+            }
+            else
+            {
+                string responseBody = await response.Content.ReadAsStringAsync(TestContext.Current.CancellationToken);
+
+                responseBody.Should()
```
bart-vmware: Please use `.BeJson()` and format into a readable raw string with line breaks.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -24,6 +24,25 @@ namespace Steeltoe.Management.Endpoint.Test.Actuators.CloudFoundry;
...
+    private const string MiddlewareForbiddenLog =
+        "FAIL Steeltoe.Management.Endpoint.Actuators.CloudFoundry.CloudFoundrySecurityMiddleware: Actuator Security Error: Forbidden - Access denied";
+
+    private const string MiddlewareUnauthorizedLog =
+        "FAIL Steeltoe.Management.Endpoint.Actuators.CloudFoundry.CloudFoundrySecurityMiddleware: Actuator Security Error: Unauthorized - Invalid token";
+
+    private const string MiddlewareUnavailableLog =
+        "FAIL Steeltoe.Management.Endpoint.Actuators.CloudFoundry.CloudFoundrySecurityMiddleware: Actuator Security Error: ServiceUnavailable - Cloud controller not reachable";
+
+    private const string CFExceptionMessage = @"Exception of type \u0027System.Net.Http.HttpRequestException\u0027 was thrown.";
```
bart-vmware: Please replace the escape sequence with literal single quotes and use the following during assertion:
```c#
JsonValue.Create(errorMessage).ToJsonString()
```

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -24,6 +24,25 @@ namespace Steeltoe.Management.Endpoint.Test.Actuators.CloudFoundry;
 
 public sealed class CloudFoundrySecurityMiddlewareTest : BaseTest
 {
+    private const string CFForbiddenLog =
```
bart-vmware: All logs here are related to CF, so I think these prefixes should be replaced with "Provider" or a similar term.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -365,6 +384,88 @@ public async Task Redacts_HTTP_headers()
...
+    [InlineData("forbidden", HttpStatusCode.Forbidden, PermissionsProvider.AccessDeniedMessage, CFForbiddenLog, "false")]
+    [InlineData("timeout", null, null, CFTimeoutLog, "true")]
+    [InlineData("timeout", null, null, CFTimeoutLog, "false")]
+    [InlineData("exception", HttpStatusCode.InternalServerError, CFExceptionMessage, CFExceptionLog, "true")]
+    [InlineData("exception", HttpStatusCode.OK, CFExceptionMessage, CFExceptionLog, "false")]
+    [InlineData("no_sensitive_data", HttpStatusCode.OK, null, MiddlewareForbiddenLog, "true")]
+    [InlineData("success", HttpStatusCode.OK, null, null, "true")]
+    [Theory]
+    public async Task CloudControllerScenarioTesting(string cloudControllerResponse, HttpStatusCode? steeltoeStatusCode, string? errorMessage,
+        string? expectedLogs, string useStatusCodeFromResponse)
```
bart-vmware: `useStatusCodeFromResponse` should be `bool`

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/StartupWithSecurity.cs
```diff
@@ -22,4 +25,15 @@ public void ConfigureServices(IServiceCollection services)
     public void Configure(IApplicationBuilder app)
     {
     }
+
+    private sealed class MockHttpClientFactory(MockHttpMessageHandler mockHttpMessageHandler) : IHttpClientFactory
+    {
+        private readonly MockHttpMessageHandler _mockHttpMessageHandler =
+            mockHttpMessageHandler ?? throw new ArgumentNullException(nameof(mockHttpMessageHandler));
```
bart-vmware: Should use `ArgumentNullException.ThrowIfNull`

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -365,6 +384,88 @@ public async Task Redacts_HTTP_headers()
...
+        }
+
+        if (!string.IsNullOrEmpty(expectedLogs))
+        {
+            string logLines = loggerProvider.GetAsText();
+            logLines.Should().Contain(expectedLogs);
+        }
+    }
+
+    private static string MockAccessToken()
```
bart-vmware: Should be a static readonly field.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -365,6 +384,88 @@ public async Task Redacts_HTTP_headers()
...
+        if (cloudControllerResponse == "timeout")
+        {
+            await Assert.ThrowsAsync<TaskCanceledException>(() => client.SendAsync(httpRequestMessage, TestContext.Current.CancellationToken));
+        }
+        else
+        {
+            HttpResponseMessage response = await client.SendAsync(httpRequestMessage, TestContext.Current.CancellationToken);
+            response.StatusCode.Should().Be(steeltoeStatusCode);
+
+            if (!string.IsNullOrEmpty(errorMessage))
```
bart-vmware:
```suggestion
            if (errorMessage != null)
```

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
bart-vmware: It took me a while to understand why this works. The HTTP mock abuses the application name to register the different responses. It would be clearer to rename `cloudControllerResponse` to `scenario`.

As _every_ test verifies a scenario, the name `CloudControllerScenarioTesting` should be clarified. Suggestion: Returns expected response on permission check.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
```diff
@@ -41,8 +43,64 @@ public async Task GetPermissionsTest()
...
+        EndpointPermissions result = await permissionsProvider.ParsePermissionsAsync(response, TestContext.Current.CancellationToken);
+        result.Should().Be(EndpointPermissions.Full);
+    }
+
+    [InlineData("unavailable", HttpStatusCode.ServiceUnavailable, PermissionsProvider.CloudfoundryNotReachableMessage)]
+    [InlineData("not-found", HttpStatusCode.Unauthorized, PermissionsProvider.InvalidTokenMessage)]
+    [InlineData("unauthorized", HttpStatusCode.Unauthorized, PermissionsProvider.InvalidTokenMessage)]
+    [InlineData("forbidden", HttpStatusCode.Forbidden, PermissionsProvider.AccessDeniedMessage)]
+    [InlineData("timeout", HttpStatusCode.ServiceUnavailable, PermissionsProvider.CloudfoundryNotReachableMessage)]
+    [InlineData("exception", HttpStatusCode.InternalServerError, "Exception of type 'System.Net.Http.HttpRequestException' was thrown.")]
```
bart-vmware: Can use `typeof().FullName`

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
```diff
@@ -41,8 +43,64 @@ public async Task GetPermissionsTest()
...
+    [InlineData("no_sensitive_data", HttpStatusCode.OK, "")]
+    [InlineData("success", HttpStatusCode.OK, "")]
+    [Theory]
+    public async Task CloudControllerScenarioTesting(string cloudControllerResponse, HttpStatusCode steeltoeStatusCode, string? expectedMessage)
```
bart-vmware: Same comments about the test name and `cloudControllerResponse` parameter. Additionally, `expectedMessage` should not be nullable -or- tests should pass null instead of "".

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/StartupWithSecurity.cs
```diff
@@ -22,4 +25,15 @@ public void ConfigureServices(IServiceCollection services)
...
+
+        public HttpClient CreateClient(string name)
+        {
+            return _mockHttpMessageHandler.ToHttpClient();
```
bart-vmware: It would be cleaner to use `DelegateToMockHttpClientHandler`, see example usage [here](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Discovery/test/Eureka.Test/EurekaDiscoveryClientTest.cs#L289) (with HttpClientFactory).

The advantage is that the test doesn't need to know the name of the `HttpClient`, so the constant doesn't need to be public.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
TimHess: With that change, these can't be constants and then don't work as inline theory data

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
TimHess: That results in [CS0182](https://learn.microsoft.com/dotnet/csharp/language-reference/compiler-messages/parameter-argument-mismatch?#arguments-for-attributes), do you have a workaround that is worth the effort to implement?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
TimHess: N/A after other changes

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/StartupWithSecurity.cs
TimHess: N/A after other changes

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
TimHess: These names are attempting to indicate the source of the interaction that triggered the log message

eg: A Cloud Foundry "Forbidden" message triggered this log vs the later case where the Steeltoe middleware evaluation determined that the Http message from the test should be forbidden. I think using "Provider" here would be confusing. I considered trying to indicate Cloud Controller somehow, but I think "CF" is accurate enough for differentiating between logs due to Steeltoe code vs external interactions

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
TimHess: Just letting it throw would take us out of alignment with Spring and in some cases could make it harder or at least slower to track down exactly what went wrong.

I'd prefer to catch everything we can and provide the maximum (safe to share) level of detail that's available, including catching those timeouts so that we can clearly and concisely say that it's a problem with the underlying platform infrastructure (Cloud Controller) and is definitely not from the app.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
TimHess: I didn't spend time looking into this next bit, but there were also some cases where we _could_ assert that more than one log entry was written. 

I tried several variations of including line breaks but decided it likely wasn't worth pursuing since it would likely either complicate the inline data (what does it look like to optionally have more than one log entry parameter) or require more of these fields. If you have thoughts we could try something, but I'm of the opinion that what we have is enough and there would be diminishing returns on any additional work put in there (if any benefit at all).

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/StartupWithSecurity.cs
TimHess: The constant is in an internal class, so it already isn't _actually_ public... Regardless, this change does not have any impact on the accessor b/c it's still needed in `EndpointServiceCollectionExtensions.cs`, but it'll now use `DelegateToMockHttpClientHandler`

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
bart-vmware: Ah, that's unfortunate.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
bart-vmware: No, unfortunately.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
bart-vmware: Why make this so complicated? Now we have two vars that indirectly depend on each other, in such a way that the compiler can't infer nullability anymore, and so we need extra null conditionals. Instead, we should just keep the `if (errorMessage != null)` check, then declare the `jsonErrorValue` inside the `if` block, because it's not used anywhere else.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
bart-vmware: The change to `InvokeAsyncReturnsExpected` doesn't look right to me. There is no actual `InvokeAsync` call anywhere in the test method body, because it's not a unit test, but an integration test. Therefore, I think we should express the functional scenario being covered in the test name.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -417,40 +388,55 @@ public async Task CloudControllerScenarioTesting(string cloudControllerResponse,
...
         using IWebHost host = builder.Build();
+        host.Services.GetRequiredService<HttpClientHandlerFactory>().Using(CloudControllerPermissionsMock.GetHttpMessageHandler());
         await host.StartAsync(TestContext.Current.CancellationToken);
 
         using HttpClient client = host.GetTestClient();
-        var httpRequestMessage = new HttpRequestMessage(HttpMethod.Get, new Uri("http://localhost/cloudfoundryapplication"));
-        httpRequestMessage.Headers.Authorization = new AuthenticationHeaderValue("Bearer", MockAccessToken());
+        var hypermediaRequestMessage = new HttpRequestMessage(HttpMethod.Get, new Uri("http://localhost/cloudfoundryapplication"));
+        hypermediaRequestMessage.Headers.Authorization = new AuthenticationHeaderValue("Bearer", MockAccessToken);
+        var fullPermissionRequestMessage = new HttpRequestMessage(HttpMethod.Get, new Uri("http://localhost/cloudfoundryapplication/info"));
```
bart-vmware: Should this be named `infoRequestMessage`, given we also have `hypermediaRequestMessage`?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
```diff
@@ -15,34 +17,92 @@ namespace Steeltoe.Management.Endpoint.Test.Actuators.CloudFoundry;
...
+        unauthorized.Code.Should().Be(HttpStatusCode.Unauthorized);
+        unauthorized.Message.Should().Be(PermissionsProvider.Messages.AuthorizationHeaderInvalid);
     }
 
-    [Fact]
-    public async Task GetPermissionsTest()
+    [InlineData(false, true, EndpointPermissions.Restricted)]
+    [InlineData(true, true, EndpointPermissions.Full)]
+    [Theory]
+    public async Task ParsePermissionsAsyncReturnsExpected(bool readSensitive, bool readBasic, EndpointPermissions expectedPermissions)
```
bart-vmware:
```suggestion
    public async Task ParsePermissionsResponseAsyncReturnsExpected(bool readSensitive, bool readBasic, EndpointPermissions expectedPermissions)
```

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
```diff
@@ -28,40 +29,43 @@ public async Task EmptyTokenIsUnauthorized()
...
     {
         PermissionsProvider permissionsProvider = GetPermissionsProvider();
         var response = new HttpResponseMessage(HttpStatusCode.OK);
 
-        var permissions = new Dictionary<string, object>
+        var permissionsJson = new Dictionary<string, bool>
```
bart-vmware: But this is a dictionary, not JSON. I find the change confusing.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
TimHess: Probably not, the reason for two requests isn't really to test hypermedia and info... It's because we want one request that needs to be authenticated and one that needs to be authenticated and authorized. I'll rename again to indicate AuthN vs AuthZ

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
TimHess: The name of this thing should not matter at all; I'm just going to inline it.

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
```diff
@@ -101,13 +103,13 @@ public async Task InvokeAsync(HttpContext context)
...
+        if (request.Headers.TryGetValue(HeaderNames.Authorization, out StringValues authorizationHeaderValue))
         {
-            string header = headerValue.ToString();
+            string authorizationValue = authorizationHeaderValue.ToString();
 
-            if (header.StartsWith(PermissionsProvider.BearerHeaderNamePrefix, StringComparison.OrdinalIgnoreCase))
+            if (authorizationValue.StartsWith(BearerTokenPrefix, StringComparison.OrdinalIgnoreCase))
             {
-                return header[PermissionsProvider.BearerHeaderNamePrefix.Length..];
+                return authorizationValue[BearerTokenPrefix.Length..];
```
bart-vmware, line 112: In method `CloudFoundrySecurityMiddleware.ReturnErrorAsync`:

image

Is this a bug? The comment seems to contradict the code.

src/Common/src/Common/Extensions/ExceptionExtensions.cs
```diff
@@ -59,4 +59,21 @@ public static bool IsCancellation(this Exception? exception)
...
         return false;
     }
+
+    /// <summary>
+    /// Determines whether the thrown exception results from an HTTP request timeout.
+    /// </summary>
+    /// <param name="exception">
+    /// The caught exception to inspect.
+    /// </param>
+    public static bool IsTimeout(this Exception? exception)
```
bart-vmware: This doesn't look right; I would expect such a method to also return `true` when the incoming exception is a `TimeoutException`. Either rename to `IsHttpClientTimeout`, or adapt the code so it becomes generally usable.

Should validate other places in the codebase where `TimeoutException` appears.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
```diff
@@ -353,18 +325,88 @@ public async Task Redacts_HTTP_headers()
...
 
         _ = await httpClient.SendAsync(requestMessage, TestContext.Current.CancellationToken);
 
         string logMessages = string.Join(System.Environment.NewLine, capturingLoggerProvider.GetAll());
         logMessages.Should().Contain("Authorization: *");
     }
 
+    [ClassData(typeof(CloudFoundrySecurityMiddlewareTestScenarios))]
+    [Theory]
+    public async Task IntegrationTestReturnsExpected(string scenario, HttpStatusCode? steeltoeStatusCode, string? errorMessage, string[] expectedLogs,
```
bart-vmware: In case of a timeout, _two_ messages are logged at Error level. There should be only one at that level:

image

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
```diff
@@ -15,34 +17,97 @@ namespace Steeltoe.Management.Endpoint.Test.Actuators.CloudFoundry;
...
+        EndpointPermissions result = await permissionsProvider.ParsePermissionsResponseAsync(cloudControllerResponse, TestContext.Current.CancellationToken);
+        result.Should().Be(expectedPermissions);
+    }
+
+    [InlineData("unavailable", HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.CloudFoundryNotReachable)]
+    [InlineData("not-found", HttpStatusCode.Unauthorized, PermissionsProvider.Messages.InvalidToken)]
+    [InlineData("unauthorized", HttpStatusCode.Unauthorized, PermissionsProvider.Messages.InvalidToken)]
+    [InlineData("forbidden", HttpStatusCode.Forbidden, PermissionsProvider.Messages.AccessDenied)]
+    [InlineData("timeout", HttpStatusCode.ServiceUnavailable, PermissionsProvider.Messages.CloudFoundryTimeout)]
+    [InlineData("exception", null, "Exception of type 'System.Net.Http.HttpRequestException' was thrown.")]
```
bart-vmware: According to the docs at https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httpclient.sendasync?view=net-9.0:

> [HttpRequestException](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httprequestexception?view=net-9.0)
The request failed due to an issue getting a valid HTTP response, such as network connectivity failure, DNS failure, server certificate validation error, or invalid server response. On .NET 8 and later versions, the reason is indicated by [HttpRequestError](https://learn.microsoft.com/en-us/dotnet/api/system.net.http.httprequestexception.httprequesterror?view=net-9.0#system-net-http-httprequestexception-httprequesterror)

I would expect a `HttpStatusCode.ServiceUnavailable` here. Should also log the `HttpRequestError`, which indicates the reason for the failure.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTestScenarios.cs
```diff
@@ -0,0 +1,102 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Net;
+using Messages = Steeltoe.Management.Endpoint.Actuators.CloudFoundry.PermissionsProvider.Messages;
+
+namespace Steeltoe.Management.Endpoint.Test.Actuators.CloudFoundry;
+
+internal sealed class CloudFoundrySecurityMiddlewareTestScenarios : TheoryData<string, HttpStatusCode?, string?, string[], bool>
+{
+    private const string CFForbiddenLog =
```
bart-vmware: I still think CF is too generic, because it's all about Cloud Foundry. It's not CF producing these log entries; it's our code that's doing so. Can you rename them to indicate what produces them?

And `typeof` can now be used, because they no longer need to be constants.

TimHess: Does this name work fine for you? `_permissionsCheckForbiddenLog` 

I don't care if we want to change these names later, it doesn't seem like it's worth spending much time on IMO

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
TimHess: Now called `IntegrationTestReturnsExpected`

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
TimHess: Does this one look ok now?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTestScenarios.cs
bart-vmware: Why not just `ProviderForbiddenLog`, so it matches with `MiddlewareForbiddenLog`?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs
bart-vmware: That's similarly vague. Suggestion: Returns expected response on permission check.

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
TimHess: The code says "If we should `UseStatusCodeFromResponse` or the error code is less than 500, we should use the status code from the response". The comment is saying "Hi friend, that's nice that you think you don't want to see a 403, but we're going to show it to you anyway because you'd be super confused if we return 200" -- `UseStatusCodeFromResponse` is for IIS/HWC compatibility reasons (and the built-in server error pages that break our ability to return a response body)

src/Common/src/Common/Extensions/ExceptionExtensions.cs
TimHess: Yeah, that thought crossed my mind, but I didn't spend much time on it since it's all internal... Do you have a preference for making this generic vs specific?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTestScenarios.cs
TimHess: `typeof` requires changing from `const` to `readonly` and thus changes the casing/underscore conventions... this is what I have pending locally:
```csharp
private const string ExceptionLogStart = "FAIL Microsoft.AspNetCore.Server.Kestrel: Connection id";
private const string ExceptionLogEnd = ": An unhandled exception was thrown by the application.";

private const string SuccessLog =
    "INFO System.Net.Http.HttpClient.CloudFoundrySecurity.ClientHandler: Sending HTTP request GET https://example.api.com/v2/apps/success/permissions";

private readonly string _permissionsCheckForbiddenLog =
    $"INFO {typeof(PermissionsProvider)}: Cloud Foundry returned status: Forbidden while obtaining permissions from: https://example.api.com/v2/apps/forbidden/permissions";

private readonly string _permissionsCheckUnauthorizedLog =
    $"INFO {typeof(PermissionsProvider)}: Cloud Foundry returned status: Unauthorized while obtaining permissions from: https://example.api.com/v2/apps/unauthorized/permissions";

private readonly string _permissionsCheckNotFoundLog =
    $"INFO {typeof(PermissionsProvider)}: Cloud Foundry returned status: NotFound while obtaining permissions from: https://example.api.com/v2/apps/not-found/permissions";

private readonly string _permissionsCheckTimeoutLog =
    $"FAIL {typeof(PermissionsProvider)}: Cloud Foundry request timed out while obtaining permissions from: https://example.api.com/v2/apps/timeout/permissions";

private readonly string _middlewareForbiddenLog =
    $"FAIL {typeof(CloudFoundrySecurityMiddleware)}: Actuator Security Error: Forbidden - {Messages.AccessDenied}";

private readonly string _middlewareTimeoutLog =
    $"FAIL{typeof(CloudFoundrySecurityMiddleware)}: Actuator Security Error: ServiceUnavailable - {Messages.CloudFoundryTimeout}";

private readonly string _middlewareUnauthorizedLog =
    $"FAIL {typeof(CloudFoundrySecurityMiddleware)}: Actuator Security Error: Unauthorized - {Messages.InvalidToken}";

private readonly string _middlewareUnavailableLog =
    $"FAIL {typeof(CloudFoundrySecurityMiddleware)}: Actuator Security Error: ServiceUnavailable - {Messages.CloudFoundryNotReachable}";
 ```

File: src/Common/src/Common/Extensions/ExceptionExtensions.cs (hunk 23 above)
Comment (bart-vmware) [reply src/Common/src/Common/Extensions/ExceptionExtensions.cs]: I'm fine making it specific, we don't have other places in the codebase where it matters.

File: src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTestScenarios.cs (hunk 26 above)
Comment (bart-vmware) [reply src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTestScenarios.cs]: Looks good.

File: src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs (hunk 22 above)
Comment (bart-vmware) [reply src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs:112]: Can you clarify the comment in code, and link to the related issue?

File: src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs (hunk 17 above)
Comment (bart-vmware) [reply src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs]: Should align the name with the other conversation.

File: src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs (hunk 24 above)
Comment (TimHess) [reply src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs]: Oh right... I had a change pending in the middleware but scrapped it at some point. How about I just make the permissions provider a warning?

Comment (bart-vmware) [reply src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudFoundrySecurityMiddlewareTest.cs]: We should never produce duplicate logs when something is misbehaving. Operators tend to aggregate the number of errors and warnings to assess the stability of systems.

If we must log both, one should be at Info level max, but we may as well remove the deepest one, as it doesn't provide additional information to users.

File: src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs (hunk 25 above)
Comment (TimHess) [reply src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs]: The PermissionsProvider does not catch anything other than `TimeoutException` anymore... are you suggesting to catch, log and rethrow?

Comment (bart-vmware) [reply src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs]: No! Log or throw, never both -- https://rolf-engelhard.de/2013/04/logging-anti-patterns-part-ii/#:~:text=Log%20and%20Throw&text=Never%20do%20both%20at%20a,might%20log%20that%20exception%20too.

Catch `HttpRequestException`, then convert it into a returned message/status code pair (like most other ones). Log the exception at error level. If not provided already, include `HttpRequestError` in the log message.

File: src/Management/test/Endpoint.Test/Actuators/CloudFoundry/CloudControllerPermissionsMock.cs [hunk 27]
```diff
@@ -0,0 +1,42 @@
...
+        httpClientHandler.Mock.When(HttpMethod.Get, "https://example.api.com/v2/apps/unauthorized/permissions")
+            .Respond(HttpStatusCode.Unauthorized, "application/json", "{}");
+
+        httpClientHandler.Mock.When(HttpMethod.Get, "https://example.api.com/v2/apps/forbidden/permissions")
+            .Respond(HttpStatusCode.Forbidden, "application/json", "{}");
+
+        httpClientHandler.Mock.When(HttpMethod.Get, "https://example.api.com/v2/apps/timeout/permissions")
+            .Throw(new OperationCanceledException(null, new TimeoutException()));
+
+        httpClientHandler.Mock.When(HttpMethod.Get, "https://example.api.com/v2/apps/exception/permissions").Throw(new HttpRequestException());
```
bart-vmware: Should throw, for example, `new HttpRequestException(HttpRequestError.NameResolutionError)`, so we have something meaningful to assert on.

src/Management/src/Endpoint/Actuators/CloudFoundry/CloudFoundrySecurityMiddleware.cs
```diff
@@ -63,13 +65,13 @@ public async Task InvokeAsync(HttpContext context)
                 _logger.LogCritical(
```
bart-vmware: Please change `LogCritical` to `LogError` here, because the system recovers from this error.

From https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loglevel?view=net-9.0-pp#fields:

> Critical: Logs that describe an unrecoverable application or system crash, or a catastrophic failure that requires immediate attention.

bart-vmware (approved):
LGTM

PR #1523, opened 2025-06-05
Improve Common.Certificates internals and tests

### LINKED ISSUE #1506: Broken test for Common.Certificates
The test `ConfigureCertificateOptionsTest.CertificateOptionsUpdateOnFileLocationChange` contains the following line twice:
```c#
_ = changeToken.RegisterChangeCallback(_ => changeCalled = true, "state");
```

The second occurrence can be commented out, but the test still succeeds. Having two callbacks on the same token that execute identical logic is pointless. It doesn't prove that a change is triggered the second time.

Also, the return value of `RegisterChangeCallback` is disposable, so it should be wrapped in a `using` statement.

TimHess:
## Description

Handle exceptions when loading certificates in IOptions, fix named options bug, simplify related code.
Enhance test scenario coverage and implementations, ensure testing for both default and named options.

Fixes #1506 

## Quality checklist

### DIFF
--- a/coverlet.runsettings
+++ b/coverlet.runsettings
@@ -5,12 +5,10 @@
       <DataCollector friendlyName="XPlat Code Coverage">
         <Configuration>
           <Format>opencover</Format>
-          <Exclude>[*.Test*]*,[*]Microsoft.Diagnostics*</Exclude>
-          <Include>[Steeltoe.*]*</Include>
-          <ExcludeByAttribute>Obsolete,GeneratedCodeAttribute,CompilerGeneratedAttribute</ExcludeByAttribute>
+          <Exclude>[ConfigurationSchemaGenerator]*</Exclude>
+          <ExcludeByFile>**/test/**/*.*</ExcludeByFile>
+          <ExcludeByAttribute>ObsoleteAttribute,GeneratedCodeAttribute,CompilerGeneratedAttribute</ExcludeByAttribute>
           <UseSourceLink>true</UseSourceLink>
-          <IncludeTestAssembly>false</IncludeTestAssembly>
-          <SkipAutoProps>false</SkipAutoProps>
         </Configuration>
       </DataCollector>
     </DataCollectors>
--- a/src/Common/src/Common/ConfigurationSchemaAttributes.cs
+++ b/src/Common/src/Common/ConfigurationSchemaAttributes.cs
@@ -6,12 +6,15 @@
 // Steeltoe: This file was copied from the .NET Aspire Configuration Schema generator
 // at https://github.com/dotnet/aspire/tree/cb7cc4d78f8dd2b4df1053a229493cdbf88f50df/src/Tools/ConfigurationSchemaGenerator.
 
+using System.Diagnostics.CodeAnalysis;
+
 namespace Aspire;
 
 /// <summary>
 /// Attribute used to automatically generate a JSON schema for a component's configuration.
 /// </summary>
 [AttributeUsage(AttributeTargets.Assembly, AllowMultiple = true)]
+[ExcludeFromCodeCoverage(Justification = "Used while compiling Steeltoe, does not ship with Steeltoe.")]
 internal sealed class ConfigurationSchemaAttribute : Attribute
 {
     public ConfigurationSchemaAttribute(string path, Type type, string[]? exclusionPaths = null)
--- a/src/Management/src/Endpoint/Actuators/Health/Contributors/FileSystem/NetworkShareWrapper.cs
+++ b/src/Management/src/Endpoint/Actuators/Health/Contributors/FileSystem/NetworkShareWrapper.cs
@@ -1,7 +1,8 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
+// Licensed to the .NET Foundation under one or more agreements.
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics.CodeAnalysis;
 using System.Runtime.InteropServices;
 
 namespace Steeltoe.Management.Endpoint.Actuators.Health.Contributors.FileSystem;
@@ -26,6 +27,7 @@ internal sealed partial class NetworkShareWrapper : INetworkShareWrapper
             : null;
     }
 
+    [ExcludeFromCodeCoverage(Justification = "Workaround for https://github.com/coverlet-coverage/coverlet/issues/1762")]
     private static partial class NativeMethods
     {
         [LibraryImport("kernel32.dll", EntryPoint = "GetDiskFreeSpaceExW", StringMarshalling = StringMarshalling.Utf16, SetLastError = true)]
--- a/src/Management/src/Endpoint/Actuators/ThreadDump/EventPipeThreadDumper.cs
+++ b/src/Management/src/Endpoint/Actuators/ThreadDump/EventPipeThreadDumper.cs
@@ -77,7 +77,9 @@ internal sealed class EventPipeThreadDumper : IThreadDumper
             }
             finally
             {
+#pragma warning disable S1215 // "GC.Collect" should not be called
                 long totalMemory = GC.GetTotalMemory(true);
+#pragma warning restore S1215 // "GC.Collect" should not be called
                 _logger.LogDebug("Total memory: {Memory}.", totalMemory);
             }
         }, cancellationToken);


bart-vmware:
This is still wrong. Try debugging through it:

image

The `changeCalled` variable is set to `true` _before_ the files get written.

bart-vmware (approved):
Thanks!

bart-vmware:
Should probably dispose `X509Certificate2` instances in tests.

bart-vmware (changes requested):
Reverting PR approval due to failing tests in cibuild.

bart-vmware:
Wouldn't it be easier to test via `OptionsMonitor`, such as in the test [ChangeConfiguration_AppliesChanges](https://github.com/SteeltoeOSS/Steeltoe/blob/f7cee9f8b3cc146c644ce9a51dc80f84f467208a/src/Discovery/test/Eureka.Test/EurekaApplicationInfoManagerTest.cs#L112), combined with `MemoryFileProvider`, such as in the test [Reloads_options_on_change](https://github.com/SteeltoeOSS/Steeltoe/blob/f7cee9f8b3cc146c644ce9a51dc80f84f467208a/src/Configuration/test/Placeholder.Test/PlaceholderConfigurationTest.cs#L332)? Then the sleeps can be removed.

TimHess:
> Wouldn't it be easier to test via `OptionsMonitor`, such as in the test [ChangeConfiguration_AppliesChanges](https://github.com/SteeltoeOSS/Steeltoe/blob/f7cee9f8b3cc146c644ce9a51dc80f84f467208a/src/Discovery/test/Eureka.Test/EurekaApplicationInfoManagerTest.cs#L112), combined with `MemoryFileProvider`, such as in the test [Reloads_options_on_change](https://github.com/SteeltoeOSS/Steeltoe/blob/f7cee9f8b3cc146c644ce9a51dc80f84f467208a/src/Configuration/test/Placeholder.Test/PlaceholderConfigurationTest.cs#L332)? Then the sleeps can be removed.

I'm not sure it's worth the effort, or exactly how we would use `TestOptionsMonitor` in this situation when we are also depending on `IConfigureNamedOptions` to read the certificate from disk, and there's code in there to exit early if the certificate is not null (which it is unless we do something to null it out).

I do like the idea of using `MemoryFileProvider`, but the problem is that `FilePathInOptionsChangeTokenSource` depends on the private class `FileSystemChangeWatcher` which only works with `PhysicalFilesWatcher`. I see an option on `PhysicalFilesWatcher` to pass in a `FileSystemWatcher`, but that still seems to be tied to a physical file system (all the [tests](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.FileProviders.Physical/tests/PhysicalFilesWatcherTests.cs#L25) are using TempDirectory rather than any form of in-memory option), so I'm not sure how to change this without adding some additional abstraction to wrap `PhysicalFilesWatcher`

bart-vmware:
> I do like the idea of using `MemoryFileProvider`, but the problem is that `FilePathInOptionsChangeTokenSource` depends on the private class `FileSystemChangeWatcher` which only works with `PhysicalFilesWatcher`. I see an option on `PhysicalFilesWatcher` to pass in a `FileSystemWatcher`, but that still seems to be tied to a physical file system (all the [tests](https://github.com/dotnet/runtime/blob/main/src/libraries/Microsoft.Extensions.FileProviders.Physical/tests/PhysicalFilesWatcherTests.cs#L25) are using TempDirectory rather than any form of in-memory option), so I'm not sure how to change this without adding some additional abstraction to wrap `PhysicalFilesWatcher`

Good point, I didn't realize that limitation. It's unfortunate that we can't use `MemoryFileProvider` due to that, but I don't think it can be easily solved, so I'm okay with the original approach.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
-        string certificateFilePath = sandbox.CreateFile("cert", firstCertificateContent);
-        string privateKeyFilePath = sandbox.CreateFile("key", firstPrivateKeyContent);
+        var secondX509 = X509Certificate2.CreateFromPemFile("instance2.crt", "instance2.key");
+        string certificateFilePath = sandbox.CreateFile(Guid.NewGuid() + ".crt", firstCertificateContent);
+        string privateKeyFilePath = sandbox.CreateFile(Guid.NewGuid() + ".key", firstPrivateKeyContent);
+
+        if (TestContext.Current.IsRunningOnBuildServer())
+        {
+            await Task.Delay(2.Seconds(), TestContext.Current.CancellationToken);
+        }
```
bart-vmware: Can we replace this with polling until both files exist and their file sizes match the expected lengths? I'm concerned this becomes flaky.

But I don't understand why a delay would be needed at all. All writes are done synchronously, so at this point all data should have been written to disk (or to the OS-level cache, at least).

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
         await File.WriteAllTextAsync(certificateFilePath, secondCertificateContent, TestContext.Current.CancellationToken);
         await File.WriteAllTextAsync(privateKeyFilePath, secondPrivateKeyContent, TestContext.Current.CancellationToken);
-        await Task.Delay(2.Seconds(), TestContext.Current.CancellationToken);
 
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(secondX509);
+        SpinWait.SpinUntil(() =>
+        {
+            try
+            {
+                return optionsMonitor.Get(certificateName).Certificate!.Equals(secondX509);
```
bart-vmware: The null suppression can be removed when inverting left/right, or when using the static `Equals` method.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
 
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(firstX509);
+        optionsMonitor.Get(certificateName).Certificate.Should().BeEquivalentTo(firstX509);
 
         await File.WriteAllTextAsync(certificateFilePath, secondCertificateContent, TestContext.Current.CancellationToken);
         await File.WriteAllTextAsync(privateKeyFilePath, secondPrivateKeyContent, TestContext.Current.CancellationToken);
-        await Task.Delay(2.Seconds(), TestContext.Current.CancellationToken);
 
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(secondX509);
+        SpinWait.SpinUntil(() =>
```
bart-vmware: I believe polling would be more appropriate than spinning.

Spinning should be used when we can't afford to wait too long, which may happen with async/await when the thread pool is being starved. So when we need to wait for one second (but _not_ longer), and then assert something happened within that second. For example, proving that health contributors run in parallel.

Here, it doesn't matter how long it takes (aside from a timeout). It would be more efficient to poll with short intervals, which allows other tests to use the thread while we give it back to the pool during delays.

For example:
```c#
using Task pollTask = PollCertificate(TestContext.Current.CancellationToken);
await pollTask.WaitAsync(TimeSpan.FromSeconds(4), TestContext.Current.CancellationToken);

optionsMonitor.Get(certificateName).Certificate.Should().Be(secondX509);

async Task PollCertificate(CancellationToken cancellationToken)
{
    while (!Equals(optionsMonitor.Get(certificateName).Certificate, secondX509))
    {
        await Task.Delay(50, cancellationToken);
    }
}
```

Another advantage is that this stops the test immediately when the test is aborted.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
         string instance1PrivateKey = await File.ReadAllTextAsync("instance.key", TestContext.Current.CancellationToken);
         var firstX509 = X509Certificate2.CreateFromPemFile("instance.crt", "instance.key");
         string instance2Certificate = await File.ReadAllTextAsync("instance2.crt", TestContext.Current.CancellationToken);
         string instance2PrivateKey = await File.ReadAllTextAsync("instance2.key", TestContext.Current.CancellationToken);
-        var secondX509 = X509Certificate2.CreateFromPemFile("instance.crt", "instance.key");
-        string certificate1FilePath = sandbox.CreateFile("cert", instance1Certificate);
-        string privateKey1FilePath = sandbox.CreateFile("key", instance1PrivateKey);
-        string certificate2FilePath = sandbox.CreateFile("cert2", instance2Certificate);
-        string privateKey2FilePath = sandbox.CreateFile("key2", instance2PrivateKey);
+        var secondX509 = X509Certificate2.CreateFromPemFile("instance2.crt", "instance2.key");
```
bart-vmware: Please establish a consistent naming convention and avoid using a mixture of `1` and `first` for related data.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(secondX509);
+        SpinWait.SpinUntil(() =>
+        {
+            try
+            {
+                return optionsMonitor.Get(certificateName).Certificate!.Equals(secondX509);
+            }
+            catch
+            {
+                return false; // File(s) may not be readable yet. Swallow exceptions and keep spinning
```
bart-vmware: Why would this ever throw? `OptionsMonitor.Get/CurrentValue` is supposed to return the latest snapshot. Failure in creating a new snapshot should be unrelated to obtaining the latest snapshot.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
         await using ServiceProvider serviceProvider = services.BuildServiceProvider(true);
 
         var optionsMonitor = serviceProvider.GetRequiredService<IOptionsMonitor<CertificateOptions>>();
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(firstX509);
+        optionsMonitor.Get(certificateName).Certificate.Should().BeEquivalentTo(firstX509);
 
         IOptionsChangeTokenSource<CertificateOptions>[] tokenSources = [.. serviceProvider.GetServices<IOptionsChangeTokenSource<CertificateOptions>>()];
+
         tokenSources.OfType<FilePathInOptionsChangeTokenSource<CertificateOptions>>().Should().HaveCount(2);
         IChangeToken changeToken = tokenSources.OfType<ConfigurationChangeTokenSource<CertificateOptions>>().Single().GetChangeToken();
```
bart-vmware: This could be:
```c#
IChangeToken changeToken = tokenSources
    .OfType<ConfigurationChangeTokenSource<CertificateOptions>()
    .Should().ContainSingle().Which.GetChangeToken();
```

But why use `.Single()` here, but `.First()` later? It looks like a bug.

First, `changeToken` is obtained from `ConfigurationChangeTokenSource`, but later it is obtained from the first of two `FilePathInOptionsChangeTokenSource` instances.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
-        changeCalled.Should().BeTrue("file path information changed");
-
-        _ = changeToken.RegisterChangeCallback(_ => changeCalled = true, "state");
-        await File.WriteAllTextAsync(certificate2FilePath, instance1Certificate, TestContext.Current.CancellationToken);
-        await File.WriteAllTextAsync(privateKey2FilePath, instance1PrivateKey, TestContext.Current.CancellationToken);
-        await Task.Delay(2.Seconds(), TestContext.Current.CancellationToken);
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(firstX509);
-        changeCalled.Should().BeTrue("file contents changed");
+
+        using (changeToken.RegisterChangeCallback(_ => changeCalled = true, "changed-path"))
```
bart-vmware: The value `"changed-path"` is passed for the `object? state` parameter, but that state is never used. Should pass `null` instead, this is just raising questions. Applies multiple times.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
+            {
+                try
+                {
+                    return optionsMonitor.Get(certificateName).Certificate!.Equals(firstX509);
+                }
+                catch
+                {
+                    return false; // File(s) may not be readable yet. Swallow exceptions and keep spinning
+                }
+            }, 6.Seconds());
```
bart-vmware: 6 seconds is pretty extreme. This doesn't look very reliable.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,243 @@ public sealed class ConfigureCertificateOptionsTest
...
-        await Task.Delay(2.Seconds(), TestContext.Current.CancellationToken);
-        optionsMonitor.Get(CertificateName).Certificate.Should().BeEquivalentTo(firstX509);
-        changeCalled.Should().BeTrue("file contents changed");
+
+        using (changeToken.RegisterChangeCallback(_ => changeCalled = true, "changed-path"))
+        {
+            changeCalled.Should().BeFalse("nothing has changed yet");
+            configurationRoot[$"{GetConfigurationKey(certificateName, "CertificateFilePath")}"] = certificate2FilePath;
+            configurationRoot[$"{GetConfigurationKey(certificateName, "PrivateKeyFilePath")}"] = privateKey2FilePath;
+            configurationRoot.Reload();
```
bart-vmware: This is a pretty extreme action that doesn't accurately reflect what happens in an app. It would be better to write to an `appsettings.json` on disk. The system detects the file changes, a sequence of triggers fire, ultimately resulting in a changed value in the options monitor. We should assert that the options monitor value changed, not the underlying triggers, which is an implementation detail that doesn't really matter for what this test is trying to prove.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: Yeah... I don't understand it either, but I only added it because exceptions were happening here in Azure DevOps. I can try removing it again and see what happens now that we've got more GHA up and running too

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: Exception handling moved to ConfigureCertificateOptions

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: Any/all null suppression has been removed from this file

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: We're looking for two different kinds of sources, and whoever wrote the test didn't feel the need to assert that the second instance of the second type also fired (since we're writing to both files at the same time, it seems reasonable to me to assume that if one fires then the other _probably_ did as well...). Probably not a bug, but not as thorough as it could have been

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: Neither `.Reload()` nor `IConfigurationRoot` are used in these tests anymore

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: yeah, and the tests were still failing with that much time, so it clearly wasn't the right solution. This code is gone now

src/Common/src/Certificates/CertificateConfigurationExtensions.cs
```diff
@@ -152,7 +109,15 @@ public static IConfigurationBuilder AddAppInstanceIdentityCertificate(this IConf
 
         if (certificateFile != null && keyFile != null)
         {
-            builder.AddCertificate(AppInstanceIdentityCertificateName, certificateFile, keyFile);
+            const string keyPrefix = $"{CertificateOptions.ConfigurationKeyPrefix}:{AppInstanceIdentityCertificateName}:";
+
+            var keys = new Dictionary<string, string?>
+            {
+                [$"{keyPrefix}CertificateFilePath"] = certificateFile,
+                [$"{keyPrefix}PrivateKeyFilePath"] = keyFile
```
bart-vmware:
```suggestion
                [$"{keyPrefix}{nameof(CertificateSettings.CertificateFilePath)}"] = certificateFile,
                [$"{keyPrefix}{nameof(CertificateSettings.PrivateKeyFilePath)}"] = keyFile
```

src/Common/src/Certificates/CertificateServiceCollectionExtensions.cs
```diff
@@ -32,9 +32,9 @@ public static IServiceCollection ConfigureCertificateOptions(this IServiceCollec
             ? CertificateOptions.ConfigurationKeyPrefix
             : ConfigurationPath.Combine(CertificateOptions.ConfigurationKeyPrefix, certificateName);
 
-        services.AddOptions<CertificateOptions>().BindConfiguration(configurationKey);
-        services.WatchFilePathInOptions<CertificateOptions>(configurationKey, certificateName, "CertificateFileName");
-        services.WatchFilePathInOptions<CertificateOptions>(configurationKey, certificateName, "PrivateKeyFileName");
+        services.AddOptions<CertificateOptions>(certificateName).BindConfiguration(configurationKey);
+        services.WatchFilePathInOptions<CertificateOptions>(CertificateOptions.ConfigurationKeyPrefix, certificateName, "CertificateFilePath");
+        services.WatchFilePathInOptions<CertificateOptions>(CertificateOptions.ConfigurationKeyPrefix, certificateName, "PrivateKeyFilePath");
```
bart-vmware:
```suggestion
        services.WatchFilePathInOptions<CertificateOptions>(CertificateOptions.ConfigurationKeyPrefix, certificateName,
            nameof(CertificateSettings.CertificateFilePath));

        services.WatchFilePathInOptions<CertificateOptions>(CertificateOptions.ConfigurationKeyPrefix, certificateName,
            nameof(CertificateSettings.PrivateKeyFilePath));
```

src/Common/src/Certificates/ConfigureCertificateOptions.cs
```diff
@@ -16,12 +19,14 @@ internal sealed class ConfigureCertificateOptions : IConfigureNamedOptions<Certi
         TimeSpan.FromSeconds(1));
 
     private readonly IConfiguration _configuration;
+    private readonly ILogger<ConfigureCertificateOptions> _logger;
 
-    public ConfigureCertificateOptions(IConfiguration configuration)
+    public ConfigureCertificateOptions(IConfiguration configuration, ILogger<ConfigureCertificateOptions>? logger = null)
```
bart-vmware: This parameter should not be optional.

src/Common/src/Certificates/ConfigureCertificateOptions.cs
```diff
@@ -42,16 +47,29 @@ public void Configure(string? name, CertificateOptions options)
...
+            X509Certificate2[] certificateChain = CertificateRegex.Matches(File.ReadAllText(certificateFilePath))
+                .Select(x => new X509Certificate2(Encoding.ASCII.GetBytes(x.Value))).ToArray();
 
-        foreach (X509Certificate2 issuer in certificateChain.Skip(1))
+            foreach (X509Certificate2 issuer in certificateChain.Skip(1))
+            {
+                options.IssuerChain.Add(issuer);
+            }
+        }
+        catch (IOException ex)
```
bart-vmware: I can't add a suggestion, but this could just be:
```c#
        catch (Exception exception) when (exception is IOException or CryptographicException)
        {
            _logger.LogDebug(exception, "Failed to load certificate for '{CertificateName}' from '{Path}'. Will retry on next reload.", name,
                certificateFilePath);
        }
```

There's no need to name the exception type in the message; it's already in the stack trace.

To assert on it from a test (if you want to, not needed), you could use the following:
```c#
CapturingLoggerProvider loggerProvider = new() { IncludeStackTraces = true };
```

src/Common/src/Certificates/ConfigureCertificateOptions.cs
```diff
@@ -42,16 +47,29 @@ public void Configure(string? name, CertificateOptions options)
...
+        {
+            _logger.LogDebug(ex, "IOException while loading certificate for '{CertificateName}' from '{Path}'. Will retry on next reload.", name,
+                certificateFilePath);
+        }
+        catch (CryptographicException ex)
         {
-            options.IssuerChain.Add(issuer);
+            _logger.LogDebug(ex, "CryptographicException while parsing certificate for '{CertificateName}' from '{Path}'. Will retry on next reload.", name,
```
bart-vmware: Should this log at warning level? If an invalid certificate is deployed, an operator likely wants to know. But if this is a transient failure that auto-fixes itself within the next second, logging at Debug level would be appropriate.

src/Common/src/Certificates/ConfigureCertificateOptions.cs
```diff
@@ -42,16 +47,29 @@ public void Configure(string? name, CertificateOptions options)
...
 
-        foreach (X509Certificate2 issuer in certificateChain.Skip(1))
+            foreach (X509Certificate2 issuer in certificateChain.Skip(1))
+            {
+                options.IssuerChain.Add(issuer);
+            }
+        }
+        catch (IOException ex)
+        {
+            _logger.LogDebug(ex, "IOException while loading certificate for '{CertificateName}' from '{Path}'. Will retry on next reload.", name,
```
bart-vmware: Can you point me to a place where this actually happens? I've added the following code to find out, but all tests remain green. I scanned for failing pipelines in AzDO, but didn't find any.

```c#
if (exception is IOException)
{
    throw null;
}
```

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
 
-        Action action = () => configureOptions.Configure(CertificateName, options);
-        action.Should().Throw<CryptographicException>();
-
+        configureOptions.Configure(certificateName, options);
         options.Certificate.Should().BeNull();
+
+        loggerProvider.GetAll().Should()
+            .Contain(
+                $"DBUG Steeltoe.Common.Certificates.ConfigureCertificateOptions: CryptographicException while parsing certificate for '{certificateName}' from 'empty.crt'. Will retry on next reload.");
```
bart-vmware: The Resharper reformatter isn't perfect. It sometimes inserts too many like breaks. In this case, it needs a bit of help. Move the `.Contain(` after the end of the first line and reformat. That leaves us with two lines instead of three.

image

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
+        using var loggerFactory = new LoggerFactory([loggerProvider]);
+
+        IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string, string?>
         {
-            [$"{CertificateOptions.ConfigurationKeyPrefix}:{CertificateName}:certificateFilePath"] = "empty.crt"
+            [$"{GetConfigurationKey(certificateName, "CertificateFilePath")}"] = "empty.crt"
         }).Build();
 
-        var configureOptions = new ConfigureCertificateOptions(configurationRoot);
+        var configureOptions = new ConfigureCertificateOptions(configuration, loggerFactory.CreateLogger<ConfigureCertificateOptions>());
```
bart-vmware: Please break it up into a separate variable to make it easier to debug/track. Such as:
```c#
        ILogger<ConfigureCertificateOptions> logger = loggerFactory.CreateLogger<ConfigureCertificateOptions>();
        var configureOptions = new ConfigureCertificateOptions(configuration, logger);
        var options = new CertificateOptions();
```

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
+    [Theory]
+    [InlineData("")]
+    [InlineData(CertificateName)]
+    public void ConfigureCertificateOptions_EmptyFile_Crash_logged(string certificateName)
     {
-        IConfigurationRoot configurationRoot = new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string, string?>
+        CapturingLoggerProvider loggerProvider = new();
+        using var loggerFactory = new LoggerFactory([loggerProvider]);
+
+        IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(new Dictionary<string, string?>
```
bart-vmware: Lines like this are more readable when broken up, such as:
```c#
var appSettings = new Dictionary<string, string?>
{
    [$"{GetConfigurationKey(certificateName, "CertificateFilePath")}"] = "empty.crt"
};

IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(appSettings).Build();
```

I realize it was already like this, but I cleaned up many of them in Management. Here's an opportunity to do the same and improve the codebase.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
         var options = new CertificateOptions();
 
-        Assert.Throws<CryptographicException>(() => configureOptions.Configure(CertificateName, options));
-
+        configureOptions.Configure(certificateName, options);
         options.Certificate.Should().BeNull();
+
+        loggerProvider.GetAll().Should()
+            .Contain(
+                $"DBUG Steeltoe.Common.Certificates.ConfigureCertificateOptions: CryptographicException while parsing certificate for '{certificateName}' from 'instance.crt'. Will retry on next reload.");
```
bart-vmware: Should use `nameof` for resiliency.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
-
+        using var firstX509 = X509Certificate2.CreateFromPemFile("instance.crt", "instance.key");
+        string secondCertificateContent = await File.ReadAllTextAsync("secondInstance.crt", TestContext.Current.CancellationToken);
+        string secondPrivateKeyContent = await File.ReadAllTextAsync("secondInstance.key", TestContext.Current.CancellationToken);
+        using var secondX509 = X509Certificate2.CreateFromPemFile("secondInstance.crt", "secondInstance.key");
+        string certificateFilePath = sandbox.CreateFile(Guid.NewGuid() + ".crt", firstCertificateContent);
+        string privateKeyFilePath = sandbox.CreateFile(Guid.NewGuid() + ".key", firstPrivateKeyContent);
+
+        MemoryFileProvider fileProvider = new();
+        fileProvider.IncludeFile(MemoryFileProvider.DefaultAppSettingsFileName, BuildAppSettingsJson(certificateName, certificateFilePath, privateKeyFilePath));
```
bart-vmware: Please extract the result from `BuildAppSettingsJson` into a variable. It is impossible to debug this way.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
+
+        using Task pollTask = PollCertificateAsync(optionsMonitor, certificateName, firstX509, TestContext.Current.CancellationToken);
+        await pollTask.WaitAsync(TimeSpan.FromSeconds(4), TestContext.Current.CancellationToken);
+
+        certificateContentsChangeCalled.Should().BeTrue("certificate file contents changed");
+        keyContentsChangeCalled.Should().BeTrue("key file contents changed");
+        optionsMonitor.Get(certificateName).Certificate.Should().Be(firstX509);
+    }
+
+    private static string BuildAppSettingsJson(string certName, string certPath, string keyPath)
```
bart-vmware: No abbreviations, please.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
+        return $$"""
+            {
+              "Certificates": {
+                {{namedCertificateSection}}
+              }
+            }
+            """;
+    }
+
+    private static async Task PollCertificateAsync(IOptionsMonitor<CertificateOptions> optionsMonitor, string certificateName,
```
bart-vmware: Suggested name: `WaitUntilCertificateChangedTo`, with `expectedCertificate` as the first parameter.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +18,287 @@ public sealed class ConfigureCertificateOptionsTest
...
+
+        MemoryFileProvider fileProvider = new();
+
+        fileProvider.IncludeFile(MemoryFileProvider.DefaultAppSettingsFileName,
+            BuildAppSettingsJson(certificateName, firstCertificateFilePath, firstPrivateKeyFilePath));
 
         var configurationBuilder = new ConfigurationBuilder();
-        configurationBuilder.AddCertificate(CertificateName, certificate1FilePath, privateKey1FilePath);
-        IConfigurationRoot configurationRoot = configurationBuilder.Build();
+        configurationBuilder.AddJsonFile(fileProvider, MemoryFileProvider.DefaultAppSettingsFileName, false, true);
```
bart-vmware: I find this an odd combination. We're using true filesystem changes everywhere else in this test.

Ideally, this test and the one below it are identical, except for the 'act' phase: overwrite appsettings, versus overwrite certificate files.

src/Common/src/Certificates/ConfigureCertificateOptions.cs
TimHess: `IOException` happens unpredictably in `X509Certificate2.CreateFromPemFile`, usually when the private key file can't be open for reading. Unfortunately, it seems like all of the relevant failures haven't been retained, probably due to this setting 
image

I've now updated this setting to retain 50 runs. I was seeing this exception commonly in under 10 iterations when I did "Run Until Failure" with the VS test runner (on my fast machine)

src/Common/src/Certificates/ConfigureCertificateOptions.cs
TimHess: If we combine the exception handling, we'll lose the ability to differentiate between the two types. I've only seen `IOException` with the transient failures and only seen `CryptographicException` with the bad file content.

src/Common/src/Certificates/ConfigureCertificateOptions.cs
bart-vmware: That's nice; hopefully, that helps narrow it down. GHA defaults to 90 days.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: `nameof` isn't giving the same fully-qualified name that's in the logs, but `typeof(ConfigureCertificateOptions).FullName` is working

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
TimHess: I also removed the part of the second test that hit the same scenario as this first test, so there's less duplication as well

src/Common/src/Certificates/ConfigureCertificateOptions.cs
bart-vmware: It doesn't look like this was changed.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
bart-vmware: This also applies to the change in `ConfigServerConfigurationSource.cs`.

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
bart-vmware: Yes, that's what I meant.

src/Common/src/Certificates/ConfigureCertificateOptions.cs
bart-vmware: To reproduce the problem, the swallow of `IOException` must be removed first.

src/Common/src/Certificates/ConfigureCertificateOptions.cs
TimHess: I took somewhat of a middle road between what was here and what seemed like contradicting suggestions... The log message was updated as suggested, but I left the separate handling of exception types to keep the (probably) transient error logged at debug level and the (probably) permanent error logged as a warning

src/Common/test/Certificates.Test/ConfigureCertificateOptionsTest.cs
```diff
@@ -17,176 +17,232 @@ public sealed class ConfigureCertificateOptionsTest
...
-        IConfigurationRoot configurationRoot = new ConfigurationBuilder().AddCertificate(CertificateName, "instance.p12").Build();
-        configurationRoot[$"{CertificateOptions.ConfigurationKeyPrefix}:{CertificateName}:certificateFilePath"].Should().NotBeNull();
-        var configureOptions = new ConfigureCertificateOptions(configurationRoot);
+        var appSettings = new Dictionary<string, string?>
+        {
+            [$"{GetConfigurationKey(certificateName, "CertificateFilePath")}"] = "instance.p12"
+        };
+
+        IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(appSettings).Build();
+        configuration[$"{GetConfigurationKey(certificateName, "CertificateFilePath")}"].Should().NotBeNull();
```
bart-vmware:
```suggestion
        IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(appSettings).Build();
```

Issue #1524, opened 2025-06-11
Change to .NET Runtime breaks reverse proxy configuration for auth libraries


TimHess:
In [this change](https://github.com/dotnet/aspnetcore/pull/61530) that was backported to .NET 8.0.17 and 9.0.6, `X-Forwarded*` headers are no longer used if they came from an unknown proxy.

Because Steeltoe does not set any `KnownProxies` on `ForwardedHeadersOptions`, OpenIdConnect link generation and Certificate Authorization are impacted.

While not explicitly stated in its documentation, I've discovered that the IP address that needs to be configured as a known proxy in Cloud Foundry _may_ be identified by one of two env variables: [`CF-INSTANCE-IP`](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#CF-INSTANCE-IP) or [`CF-INSTANCE-INTERNAL-IP`](https://docs.cloudfoundry.org/devguide/deploy-apps/environment-variable.html#CF-INSTANCE-INTERNAL-IP).

For Certificate auth: it is trivial to register that address in `CertificateAuthorizationBuilderExtensions` when the env var is set.

For OpenIdConnect: Steeltoe [documents how](https://steeltoe.io/docs/v4/security/sso-open-id.html#add-and-use-openid-connect), but does not directly configure `ForwardedHeadersOptions`.

TimHess:
After further investigation, it looks like while the variables could be used in our test environment, they are not actually guaranteed to be in any way related to a valid reverse proxy address and are unlikely to provide a reliable solution

PR #1525, opened 2025-06-11
When present, add CF_INSTANCE_INTERNAL_IP to ASP.NET Core Known Proxies

TimHess:
## Description

Provides a fix for #1524 for certificate auth

## Quality checklist


src/Security/src/Authorization.Certificate/CertificateAuthorizationBuilderExtensions.cs
```diff
@@ -66,6 +68,14 @@ public static AuthorizationBuilder AddOrgAndSpacePolicies(this AuthorizationBuil
             }
         });
 
+        if (IPAddress.TryParse(Environment.GetEnvironmentVariable("CF_INSTANCE_INTERNAL_IP"), out IPAddress? address))
+        {
+            builder.Services.Configure<ForwardedHeadersOptions>(options =>
+            {
+                options.KnownProxies.Add(address);
+            });
```
bart-vmware:
```suggestion
            builder.Services.Configure<ForwardedHeadersOptions>(options => options.KnownProxies.Add(address));
```

src/Security/src/Authorization.Certificate/CertificateAuthorizationBuilderExtensions.cs
```diff
@@ -66,6 +68,14 @@ public static AuthorizationBuilder AddOrgAndSpacePolicies(this AuthorizationBuil
             }
         });
 
+        if (IPAddress.TryParse(Environment.GetEnvironmentVariable("CF_INSTANCE_INTERNAL_IP"), out IPAddress? address))
```
bart-vmware: We _could_ reuse `CloudFoundryEnvironmentSettingsReader.InstanceIP` here, which makes tracking the usage of this environment variable a bit easier. However, that requires taking a dependency on the `Configuration.CloudFoundry` package, which may not be desired.

bart-vmware (commented):
Thanks for diving into this. I haven't tested the changes, assuming you did.

src/Security/src/Authorization.Certificate/CertificateAuthorizationBuilderExtensions.cs
TimHess: Gray area for sure... I think it would be fine to add any abstraction package or any from the Common family. `ICloudFoundrySettingsReader` and the implementation _could_ move to the abstractions package, but maybe it's best to just duplicate the string for now

TimHess:
> Thanks for diving into this. I haven't tested the changes, assuming you did.

Yes, but also because https://github.com/SteeltoeOSS/Samples/pull/391 works with the same set of sample apps and I was hitting weird probably-environment issues during my first round of testing, I was already planning to leave this PR draft/open while wrapping up the other effort

src/Common/src/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// </exception>
+    public static IServiceCollection ConfigureForwardedHeadersOptionsForCloudFoundry(this IServiceCollection services)
+    {
+        ArgumentNullException.ThrowIfNull(services);
+
+        services.AddOptions();
+
+        services.AddOptions<ForwardedHeadersOptions>().Configure<IServiceProvider>((options, serviceProvider) =>
+        {
+            ILogger logger = serviceProvider.GetService<ILoggerFactory>()?.CreateLogger("Steeltoe.Common.Hosting.ServiceCollectionExtensions") ??
```
bart-vmware: I think this could follow the usual `IConfigureOptions` pattern, where we register our configurer in the container and inject dependencies the regular way. If that depends on `ILogger`, the registering method should call `services.AddLogging()` to ensure it's present.

src/Common/src/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+    /// <remarks>
+    /// Evaluates the environment variables CF_INSTANCE_IP and CF_INSTANCE_INTERNAL_IP, and if they contain valid IP addresses, adds their network to
+    /// <see cref="ForwardedHeadersOptions.KnownNetworks" />.
+    /// <para />
+    /// IMPORTANT: <see cref="ForwardedHeadersExtensions.UseForwardedHeaders(IApplicationBuilder)" /> must be called separately to activate these options.
+    /// </remarks>
+    /// <exception cref="ArgumentNullException">
+    /// Thrown if <paramref name="services" /> is <c>null</c>.
+    /// </exception>
+    public static IServiceCollection ConfigureForwardedHeadersOptionsForCloudFoundry(this IServiceCollection services)
```
bart-vmware: Can this be moved to the `Steeltoe.Configuration.CloudFoundry` project?

src/Common/src/Hosting/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,98 @@
...
+        return services;
+    }
+
+    private static void AddKnownNetworkFromEnv(string envVar, ForwardedHeadersOptions options, ILogger logger)
+    {
+        string? ipString = Environment.GetEnvironmentVariable(envVar);
+
+        if (IPAddress.TryParse(ipString, out IPAddress? address))
+        {
+            // Assume a /24 subnet mask (255.255.255.0), adjust if needed
```
bart-vmware: Is this assumption always correct? According to https://techdocs.broadcom.com/content/dam/broadcom/techdocs/us/en/pdf/vmware-tanzu/platform/tanzu-platform-for-cloud-foundry/10-0/tpcf/tpcf.pdf (search for "subnet") and https://techdocs.broadcom.com/us/en/vmware-tanzu/platform/tanzu-platform-for-cloud-foundry/4-0/tpcf/deploy-apps-cf-networking.html#cells, this can be configured in OPS Manager.

src/Common/test/Hosting.Test/ServiceCollectionExtensionsTest.cs
```diff
@@ -0,0 +1,111 @@
...
+        options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedProto);
+        AssertDefaultKnownNetworks(options);
+        AssertDefaultKnownProxies(options);
+    }
+
+    [Fact]
+    public void Adds_KnownNetworks_FromEnvironmentVariables()
+    {
+        const string instanceIPAddress = "192.168.7.100";
+        const string instanceInternalIPAddress = "10.11.12.255";
```
bart-vmware: This looks like a broadcast address, not the address of a single host.

src/Common/test/Hosting.Test/ServiceCollectionExtensionsTest.cs
```diff
@@ -0,0 +1,111 @@
...
+    private static readonly IPAddress LocalhostV4 = IPAddress.Parse("127.0.0.1");
+    private static readonly IPAddress LocalhostV6 = IPAddress.Parse("::1");
+
+    [Fact]
+    public void Adds_XForwardedHost_and_XForwardedProto_Headers()
+    {
+        var services = new ServiceCollection();
+
+        services.ConfigureForwardedHeadersOptionsForCloudFoundry();
+        ServiceProvider provider = services.BuildServiceProvider();
```
bart-vmware: Should pass `true` argument to validate the container on build.

bart-vmware: And wrap in `using` statement.

src/Common/test/Hosting.Test/ServiceCollectionExtensionsTest.cs
```diff
@@ -0,0 +1,111 @@
...
+        ForwardedHeadersOptions options = provider.GetRequiredService<IOptions<ForwardedHeadersOptions>>().Value;
+
+        AssertKnownNetworksContains(options, IPAddress.Parse(sameIP), 24);
+        AssertDefaultKnownProxies(options);
+    }
+
+    [Fact]
+    public void Does_not_add_invalid_EnvironmentVariables()
+    {
+        Environment.SetEnvironmentVariable("CF_INSTANCE_IP", "invalid-ip");
```
bart-vmware: This isn't safe. Use `EnvironmentVariableScope` instead.

src/Common/src/Hosting/ServiceCollectionExtensions.cs
TimHess: Probably not, removed handling of the vars entirely

src/Common/test/Hosting.Test/ServiceCollectionExtensionsTest.cs
TimHess: yeah, it probably wasn't the best test data. gone now

src/Configuration/src/CloudFoundry/ForwardedHeadersSettings.cs
```diff
@@ -0,0 +1,20 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace Steeltoe.Configuration.CloudFoundry;
+
+public sealed class ForwardedHeadersSettings
+{
+    internal const string ConfigurationKey = "Steeltoe:ForwardedHeaders";
+
+    /// <summary>
+    /// Gets or sets a value indicating whether to trust all networks (adds 0.0.0.0/0). WARNING: Use only behind a trusted ingress.
```
bart-vmware:
```suggestion
    /// Gets or sets a value indicating whether to trust all networks (adds 0.0.0.0/0). WARNING: Use only behind a trusted ingress. Default value: true.
```

src/Configuration/src/CloudFoundry/ForwardedHeadersSettings.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// </summary>
+    public bool TrustAllNetworks { get; set; } = true;
+
+    /// <summary>
+    /// Gets or sets known networks to trust, in CIDR notation. Example: "10.0.0.0/8,192.168.0.0/16".
+    /// </summary>
+    public string? KnownNetworks { get; set; }
```
bart-vmware: According to https://nestenius.se/net/configuring-asp-net-core-forwarded-headers-middleware, `KnownNetworks` can already be specified in appsettings and is bound automatically (when using a non-empty host builder, see [here](https://github.com/dotnet/aspnetcore/blob/ee050bd56bdf2b653d4bad75f26ba8802a4f58fa/src/DefaultBuilder/src/GenericHostBuilderExtensions.cs#L25) and [here](https://github.com/dotnet/aspnetcore/blob/ee050bd56bdf2b653d4bad75f26ba8802a4f58fa/src/DefaultBuilder/src/WebHost.cs#L269-L270)). So I don't think we need to introduce our own option for that. Based on `ASPNETCORE_FORWARDEDHEADERS_ENABLED`, the middleware is added automatically using a startup filter (when using a non-empty host builder, it defaults to `true`).

I verified that the following `appsettings.json` properly binds in a sample:
```json
"ForwardedHeaders": {
  "ForwardedHeaders": "XForwardedFor,XForwardedHost,XForwardedProto",
  "ForwardLimit": 1,
  "RequireHeaderSymmetry": true,
  "KnownProxies": [ "::1" ],
  "KnownNetworks": [ "127.0.0.1/8" ],
  "AllowedHosts": [ "nestenius.se" ]
}
```
image

src/Configuration/src/CloudFoundry/ForwardedHeadersSettings.cs
```diff
@@ -0,0 +1,20 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+namespace Steeltoe.Configuration.CloudFoundry;
+
+public sealed class ForwardedHeadersSettings
```
bart-vmware: The name *Settings is misleading, because Aspire uses that to indicate it's _not_ an `IOptions`.

If this type is going to stay, it should be added to the configuration schema.

src/Configuration/test/CloudFoundry.Test/CloudFoundryServiceCollectionExtensionsTest.cs
```diff
@@ -49,4 +54,285 @@ public async Task ConfigureCloudFoundryOptions_ConfiguresCloudFoundryOptions()
         Assert.Equal(16384, appOptions.Value.Limits?.FileDescriptor);
         Assert.Equal("playground", appOptions.Value.SpaceName);
     }
+
```
bart-vmware: We should have a test using `TestWebApplicationBuilderFactory.CreateDefault()` to verify settings are bound from configuration (using both our own configurer, and the one from ASP.NET), and that the startup filter adds the middleware.

src/Configuration/src/CloudFoundry/ForwardedHeadersSettings.cs
TimHess: I do not see anything that binds `ForwardedHeadersOptions` to configuration and have been unable to make anything like your suggestion work. The values you have there for `KnownProxies` and `KnownNetworks` are the defaults. 

Setting `ASPNETCORE_FORWARDEDHEADERS_ENABLED` clears out `KnownProxies` and `KnownNetworks` but is still not resulting in options bound from configuration for me.

src/Configuration/src/CloudFoundry/ForwardedHeadersSettings.cs
TimHess: This class has been removed entirely

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,101 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        using var vcapScope = new EnvironmentVariableScope("VCAP_APPLICATION", onCloudFoundry ? "{}" : null);
+
+        WebApplicationBuilder builder = TestWebApplicationBuilderFactory.CreateDefault();
+        builder.AddCloudFoundryConfiguration();
+        await using WebApplication host = builder.Build();
+        bool? isHttpRequest = null;
+
+        host.Map("/", context =>
+        {
+            isHttpRequest = context.Request.IsHttps;
```
bart-vmware: Should this be `isHttpsRequest`? (multiple places)
A better name would indicate what this functionally means, which is (as far as I understand): the middleware has kicked in and changed the `HttpRequest`, based on the incoming headers.

src/Security/src/Authorization.Certificate/CertificateApplicationBuilderExtensions.cs
```diff
@@ -31,7 +36,7 @@ public static IApplicationBuilder UseCertificateAuthorization(this IApplicationB
     /// The <see cref="IApplicationBuilder" /> to configure.
     /// </param>
     /// <param name="options">
-    /// Custom header forwarding policy. <see cref="ForwardedHeaders.XForwardedProto" /> is added to your <see cref="ForwardedHeadersOptions" />.
+    /// Custom header forwarding policy.
```
bart-vmware: We have no tests that call this overload. Should we?

I wonder why options aren't bound to configuration. Perhaps because they may vary in branched pipelines? Do you have any idea why it was designed this way in ASP.NET?

Interestingly, we _do_ have the test `ForwardedHeaders_configuration_can_be_overridden`, which adds a configurer to customize options. Do we need to keep this overload at all?

src/Security/src/Authorization.Certificate/CertificateApplicationBuilderExtensions.cs
```diff
@@ -21,7 +19,14 @@ public static class CertificateApplicationBuilderExtensions
     /// </returns>
     public static IApplicationBuilder UseCertificateAuthorization(this IApplicationBuilder builder)
     {
-        return UseCertificateAuthorization(builder, new ForwardedHeadersOptions());
+        ArgumentNullException.ThrowIfNull(builder);
+
+        builder.UseForwardedHeaders();
+        builder.UseCertificateForwarding();
+        builder.UseAuthentication();
+        builder.UseAuthorization();
```
bart-vmware: We can reduce code duplication by making both overloads call a shared private method that has a nullable `options` parameter.

src/Configuration/src/CloudFoundry/CloudFoundryConfigurationProvider.cs
```diff
@@ -73,6 +73,10 @@ private void Process()
 
             LoadData("vcap:application", applicationData.GetChildren(), data);
             AddDiegoVariables(data);
+
+            // Enable evaluation of X-Forwarded headers so that ASP.NET Core works automatically behind Gorouter.
+            // Equivalent to setting ASPNETCORE_FORWARDEDHEADERS_ENABLED to true.
+            data["FORWARDEDHEADERS_ENABLED"] = "true";
```
bart-vmware, line 79: According to [here](https://nestenius.se/net/configuring-asp-net-core-forwarded-headers-middleware/#enabledvariable), setting this activates a startup filter that automatically calls `app.UseForwardedHeaders()` very early in the middleware pipeline.

How does that relate to our `UseCertificateAuthorization` extension method? Does this setting effectively make the `builder.UseForwardedHeaders()` call there redundant?

src/Configuration/src/CloudFoundry/ForwardedHeadersSettings.cs
bart-vmware: > I do not see anything that binds `ForwardedHeadersOptions` to configuration and have been unable to make anything like your suggestion work. The values you have there for `KnownProxies` and `KnownNetworks` are the defaults.

> Setting `ASPNETCORE_FORWARDEDHEADERS_ENABLED` clears out `KnownProxies` and `KnownNetworks` but is still not resulting in options bound from configuration for me.

Ah, that explains it. I wasn't aware these were the defaults. You're right, it doesn't actually work.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,101 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        ForwardedHeadersOptions options = host.Services.GetRequiredService<IOptions<ForwardedHeadersOptions>>().Value;
+
+        options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedFor);
+        options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedProto);
+        options.KnownNetworks.Should().BeEmpty();
+        options.KnownProxies.Should().BeEmpty();
+    }
+
+    [Fact]
+    public async Task ForwardedHeaders_not_enabled_off_CloudFoundry()
```
bart-vmware: I find this and related test names hard to understand. Suggestion:
Header forwarding is restricted/unrestricted when (not) running on Cloud Foundry.
Header forwarding is applied depending on cloud platform

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,101 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        host.Map("/", context =>
+        {
+            isHttpRequest = context.Request.IsHttps;
+            return Task.CompletedTask;
+        });
+
+        await host.StartAsync(TestContext.Current.CancellationToken);
+        TestServer server = host.GetTestServer();
+
+        HttpContext context = await server.SendAsync(c =>
```
bart-vmware: Please don't use single-character variable names.

`context` should be renamed to `httpContext`, we also have `TestContext` here.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,101 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+    {
+        using var vcapScope = new EnvironmentVariableScope("VCAP_APPLICATION", "{}");
+        WebApplicationBuilder builder = TestWebApplicationBuilderFactory.CreateDefault();
+        builder.AddCloudFoundryConfiguration();
+        await using WebApplication host = builder.Build();
+
+        ForwardedHeadersOptions options = host.Services.GetRequiredService<IOptions<ForwardedHeadersOptions>>().Value;
+
+        options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedFor);
+        options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedProto);
```
bart-vmware: Should this assert on having exactly these two flags, as opposed to the current "any value is fine, as long as at least these two flags are present"?

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,101 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+
+        options.ForwardedHeaders.Should().Be(ForwardedHeaders.None);
+        options.KnownNetworks.Should().ContainSingle().Which.Should().BeEquivalentTo(IPNetwork.Parse("127.0.0.1/8"));
+        options.KnownProxies.Should().ContainSingle().Which.Should().Be(IPAddress.Parse("::1"));
+    }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task ForwardedHeadersMiddleware_enabled_as_expected(bool onCloudFoundry)
```
bart-vmware: `onCloudFoundry` should be either `isCloudFoundry` or `isRunningOnCloudFoundry`.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,101 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        bool? isHttpRequest = null;
+
+        host.Map("/", context =>
+        {
+            isHttpRequest = context.Request.IsHttps;
+            return Task.CompletedTask;
+        });
+
+        await host.StartAsync(TestContext.Current.CancellationToken);
+        TestServer server = host.GetTestServer();
```
bart-vmware: Why not use `using HttpClient httpClient = host.GetTestClient();`, which we do in most other places?

TimHess: With that approach, `Connection.RemoteIpAddress` is `null`. We _can_ use it when we don't care about `KnownProxies`, `KnownNetworks` or `X-Forwarded-For`, but I think we should do it this way for both of these tests.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
TimHess: It _could_, but we only care that these two are present and I'm not sure it matters to us either way if ASP.NET Core is changed to set more of them... I'm inclined to leave it like this

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
TimHess: Not _exactly_ what you asked for, but I combined two more `Fact`s into a `Theory` since they were so similar and updated all the names

src/Security/src/Authorization.Certificate/CertificateApplicationBuilderExtensions.cs
TimHess: All good questions... I'm not sure about the history or if there was any original _intent_ behind this behavior. 

I think we've now seen a few areas where ASP.NET Core just didn't (and sometimes still doesn't) do configuration binding, probably along the lines of "do it in code so you always know exactly what's going to happen at compile time" (all of the security libraries we're working with didn't do any configuration binding until ... idk, probably sometime after .NET 5).

I didn't remove this option at first because I didn't want to add yet another breaking change, but since the behavior is different, I think I am ok with removing it until/unless somebody says they need this option specifically.

src/Security/src/Authorization.Certificate/CertificateApplicationBuilderExtensions.cs
TimHess: Removing the other extension entirely should take care of the duplication

src/Configuration/src/CloudFoundry/CloudFoundryConfigurationProvider.cs
TimHess: Assuming some flavor of `.AddCloudFoundryConfiguration` is called, it is redundant after we also remove the variant that passes in an options instance.

`UseForwardedHeaders` will make sure the middleware isn't added twice though, so I don't think it hurts to keep it in case somebody doesn't want to add our CF config provider:
```
    public static IApplicationBuilder UseForwardedHeaders(this IApplicationBuilder builder)
    {
        ArgumentNullException.ThrowIfNull(builder);

        // Don't add more than one instance of this middleware to the pipeline using the options from the DI container.
        // Doing so could cause a request to be processed multiple times and the ForwardLimit to be exceeded.
        if (!builder.Properties.ContainsKey(ForwardedHeadersAdded))
        {
            builder.Properties[ForwardedHeadersAdded] = true;
            return builder.UseMiddleware<ForwardedHeadersMiddleware>();
        }

        return builder;
    }
```

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
bart-vmware: I agree, didn't know about that.

src/Configuration/src/CloudFoundry/CloudFoundryConfigurationProvider.cs
bart-vmware: Yes, I saw that. But I'm wondering if calling `UseCertificateAuthorization` when _not_ running on Cloud Foundry is a valid use case.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
         Assert.Equal("fb8fbcc6-8d58-479e-bcc7-3b4ce5a7f0ca", options.Version);
     }
 
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task ForwardedHeadersOptions_unrestricted_when_running_on_CloudFoundry(bool isRunningOnCloudFoundry)
+    {
+        using var vcapScope = new EnvironmentVariableScope("VCAP_APPLICATION", isRunningOnCloudFoundry ? "{}" : null);
```
bart-vmware:
```suggestion
        using IDisposable? scope = isRunningOnCloudFoundry ? new EnvironmentVariableScope("VCAP_APPLICATION", "{}") : null;

```
This is a bit safer. There are subtleties regarding environment variables having the value `null` or `""` versus not being defined.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        if (isRunningOnCloudFoundry)
+        {
+            options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedFor);
+            options.ForwardedHeaders.Should().HaveFlag(ForwardedHeaders.XForwardedProto);
+            options.KnownNetworks.Should().BeEmpty();
+            options.KnownProxies.Should().BeEmpty();
+        }
+        else
+        {
+            options.ForwardedHeaders.Should().Be(ForwardedHeaders.None);
```
bart-vmware:
```suggestion
            options.ForwardedHeaders.Should().NotHaveFlag(ForwardedHeaders.XForwardedFor);
            options.ForwardedHeaders.Should().NotHaveFlag(ForwardedHeaders.XForwardedProto);
```
Based on the other conversation, assuming we only care about these two flags.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        responseContext.Response.StatusCode.Should().Be(200);
+        forwardedHeadersWereEvaluated.Should().BeFalse("X-Forwarded-Proto should not be evaluated for unknown proxies");
+    }
+
+    [Theory]
+    [InlineData(true)]
+    [InlineData(false)]
+    public async Task ForwardedHeadersMiddleware_updates_connection_details_when_running_on_CloudFoundry(bool isRunningOnCloudFoundry)
+    {
+        using var vcapScope = new EnvironmentVariableScope("VCAP_APPLICATION", isRunningOnCloudFoundry ? "{}" : null);
```
bart-vmware:
```suggestion
        using IDisposable? scope = isRunningOnCloudFoundry ? new EnvironmentVariableScope("VCAP_APPLICATION", "{}") : null;

```

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        host.Map("/", context =>
+        {
+            forwardedHeadersWereEvaluated = context.Request.IsHttps;
+            return Task.CompletedTask;
+        });
+
+        await host.StartAsync(TestContext.Current.CancellationToken);
+        TestServer server = host.GetTestServer();
+
+        HttpContext responseContext = await server.SendAsync(requestContext =>
```
bart-vmware: I don't understand the names here. An `HttpContext` contains both request and response data. Why not name both of them `httpContext`?

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        }
+    }
+
+    [Fact]
+    public async Task ForwardedHeadersOptions_can_be_customized_when_running_on_CloudFoundry()
+    {
+        using var vcapScope = new EnvironmentVariableScope("VCAP_APPLICATION", "{}");
+        WebApplicationBuilder builder = TestWebApplicationBuilderFactory.CreateDefault();
+        builder.AddCloudFoundryConfiguration();
+        builder.Services.Configure<ForwardedHeadersOptions>(options => options.KnownProxies.Add(IPAddress.Parse("192.168.1.20")));
```
bart-vmware: We're setting an IP here, changing it later, but never assert on it. Should there be an assertion, or is this setup redundant entirely?

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+        await using WebApplication host = builder.Build();
+        bool? forwardedHeadersWereEvaluated = null;
+
+        host.Map("/", context =>
+        {
+            forwardedHeadersWereEvaluated = context.Request.IsHttps;
```
bart-vmware, line 270: I don't understand what this proves. An HTTPS request is never issued.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -216,6 +224,96 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
+            return Task.CompletedTask;
+        });
+
+        await host.StartAsync(TestContext.Current.CancellationToken);
+        TestServer server = host.GetTestServer();
+
+        HttpContext responseContext = await server.SendAsync(requestContext =>
```
bart-vmware: I'm still a bit unclear on how this works. Wouldn't it be more reliable to:
1. Define local variables for the relevant HttpContext parameters being captured
2. Assign them in `host.Map`
3. Send a regular request, passing the `X-*` header values
4. Assert on the response status and locals at the end

For example, it's unclear to me why `RemoteIpAddress` is being set to `Loopback`; it seems we're not really testing what would happen in a real app.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
bart-vmware: Oh, it comes from the header.

src/Configuration/src/CloudFoundry/CloudFoundryConfigurationProvider.cs
TimHess: `UseCertificateAuthorization` should still work for local development

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
TimHess: I had it that way at one point, but it seemed more confusing to me. If it's easier for you that way, I'll change it

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
TimHess: `ForwardedHeadersMiddleware` requires `RemoteIpAddress` to not be null to hit the logic change that brought us to this PR in the first place. 
Regular requests do not have a way to set `RemoteIpAddress`. For this test it doesn't matter what IP is used, just that _something_ is set.
`host.Map` is hit after the middleware runs, so it's too late to set anything relevant.

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
TimHess: We could, but the goal of this test is to prove that customizations to `ForwardedHeadersOptions` will have an impact on the header processing, so the test intends to look more closely at the impact of the change than the change itself. Here we set a known proxy and later simulate traffic through some other IP with the intent to see the x-forwarded headers ignored, whereas it would pass given the same setup as the test below. Basically, this test wants to prove that if you can lock things down if you want (and things will break if you do that wrong).

bart-vmware (approved):

PR #1530, opened 2025-06-17
Add GHA workflow for Code Style

bart-vmware:
## Description

Adds a GitHub Actions workflow for code style validation.

> [!NOTE]
> Running on a branch can't be tested until this is merged to the main branch.

## Quality checklist

### DIFF
--- a/.github/workflows/verify-code-style.yml
+++ b/.github/workflows/verify-code-style.yml
@@ -1,0 +1,61 @@
+name: Verify code style
+
+on:
+  push:
+    branches: [ 'main', 'release/*' ]
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  DOTNET_CLI_TELEMETRY_OPTOUT: 1
+  DOTNET_NOLOGO: true
+
+jobs:
+  verify-code-style:
+    runs-on: ubuntu-latest
+    steps:
+    - name: Setup .NET
+      uses: actions/setup-dotnet@v4
+      with:
+        dotnet-version: |
+          8.0.*
+          9.0.*
+
+    - name: Checkout repo
+      uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+
+    - name: Restore tools
+      run: dotnet tool restore --verbosity minimal
+
+    - name: Restore packages
+      run: dotnet restore src --verbosity minimal
+
+    - name: Set package version
+      run: nbgv cloud
+
+    - name: CleanupCode (on PR diff)
+      if: github.event_name == 'pull_request'
+      shell: pwsh
+      run: |
+        # Not using the environment variables for SHAs, because they may be outdated. This may happen on force-push after the build is queued, but before it starts.
+        # The below works because HEAD is detached (at the merge commit), so HEAD~1 is at the base branch. When a PR contains no commits, this job will not run.
+        $headCommitHash = git rev-parse HEAD
+        $baseCommitHash = git rev-parse HEAD~1
+
+        Write-Output "Running code cleanup on commit range $baseCommitHash..$headCommitHash in pull request."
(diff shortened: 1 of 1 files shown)


.github/workflows/verify-code-style.yml
```diff
@@ -0,0 +1,62 @@
+name: Verify code style
+
+on:
+  push:
+    branches: [ 'main', 'release/*' ]
```
TimHess, line 5: Does `'release/*'` really make sense here? I would think anything that matches that pattern would be pretty locked in

.github/workflows/verify-code-style.yml
```diff
@@ -0,0 +1,62 @@
...
+
+        Write-Output "Running code cleanup on commit range $baseCommitHash..$headCommitHash in pull request."
+        dotnet regitlint -s src/Steeltoe.All.sln --print-command --skip-tool-check --max-runs=5 --jb --dotnetcoresdk=$(dotnet --version) --jb-profile="Steeltoe Full Cleanup" --jb --properties:Configuration=Release --jb --properties:RunAnalyzers=false --jb --properties:NuGetAudit=false --jb --verbosity=WARN -f commits -a $headCommitHash -b $baseCommitHash --fail-on-diff --print-diff
+
+    - name: CleanupCode (on branch)
+      if: github.event_name == 'push' || github.event_name == 'release'
+      shell: pwsh
+      run: |
+        Write-Output "Running code cleanup on all files."
+        dotnet regitlint -s JsonApiDotNetCore.sln --print-command --skip-tool-check --jb --dotnetcoresdk=$(dotnet --version) --jb-profile="Steeltoe Full Cleanup" --jb --properties:Configuration=Release --jb --properties:RunAnalyzers=false --jb --properties:NuGetAudit=false --jb --verbosity=WARN --fail-on-diff --print-diff
```
TimHess:
```suggestion
        dotnet regitlint -s src/Steeltoe.All.sln --print-command --skip-tool-check --jb --dotnetcoresdk=$(dotnet --version) --jb-profile="Steeltoe Full Cleanup" --jb --properties:Configuration=Release --jb --properties:RunAnalyzers=false --jb --properties:NuGetAudit=false --jb --verbosity=WARN --fail-on-diff --print-diff
```

bart-vmware: Thanks!

.github/workflows/verify-code-style.yml
bart-vmware: We'd open a PR to fix a bug in a previous release, right? Then we'd like to run the usual checks.

TimHess: This is a push trigger though, so I think it doesn't fire till the PR merges

bart-vmware: Ah yes. It already triggers on any PR. But what's the downside of keeping it?

TimHess: Noise and wasted compute cycles if we're never going to act on the results... It's not a substantial problem, but what benefit is there to keeping it?

bart-vmware: I was thinking of adding a badge for it. On a PR, only the changed files are verified. It's nice to have a full run when other things change, such as the Resharper version, the .NET version, etc.

bart-vmware: Similar to running sonar, coverage reports, etc.

TimHess: I'm still not really seeing the benefit, but don't see a need to hold up the change on these grounds

bart-vmware: Thanks. What I did here matches the current Azure DevOps behavior. Let's reconsider next time we're updating pipelines and workflows.

TimHess: Sure, but we never got around to updating those pipelines after changing the 2.x and 3.x branch names to drop `release/`

bart-vmware: Hmm, good point.

PR #1542, opened 2025-07-10
Convert xUnit assertions to Fluent Assertions style and optimize assertions

### LINKED ISSUE #931: Use FluentAssertions instead of Xunit assertions
Refactor existing tests to use [FluentAssertions](https://fluentassertions.com/) instead of `Assert.*` methods, to get better feedback than "True should be False" when a test fails.

This can be automated using https://github.com/fluentassertions/fluentassertions.analyzers, which should be activated in `sharedtest.props`.

### LINKED ISSUE #1436: Drop FluentAssertions dependency
Because FluentAssertions v8 [requires a paid license](https://github.com/fluentassertions/fluentassertions/pull/2943) for closed-source commercial usage, we'll likely migrate away from it. Although Steeltoe itself is open-source, companies that fork Steeltoe privately would need to purchase a license (per developer). This requirement reduces the attractiveness of Steeltoe, which we'd like to avoid.

Potential future steps (and accompanying tools):
- Replace with [Shouldly](https://github.com/shouldly/shouldly)
  - https://github.com/shouldly/shouldly/issues/1034 (mentions PowerShell migration script)
- Replace with [NFluent](https://github.com/tpierrain/NFluent)
  - https://github.com/maximedezette/fluentassertions-to-nfluent-migrator
- Use xUnit built-in assertions
  - Migration tool at https://github.com/amoerie/fluentassertions-migrator
- Stick with FA v7 forever (or switch to a well-maintained fork, if one becomes available)
  - Pin dependency as `<PackageReference Include="FluentAssertions" Version="[7.0.0,8.0.0)" />` to block package updates

Please feel free to share your thoughts and concerns regarding the license change of this dependency.

bart-vmware:
## Description

Helpful analyzers used to convert syntax:
- [Meziantou.FluentAssertionsAnalyzers](https://www.nuget.org/packages/Meziantou.FluentAssertionsAnalyzers)
- [FluentAssertions.Analyzers](https://www.nuget.org/packages/FluentAssertions.Analyzers)

Fixes #931.
Closes #1436.

## Quality checklist


src/Discovery/test/Consul.Test/Discovery/PostConfigureConsulDiscoveryOptionsTest.cs
```diff
@@ -133,19 +135,39 @@ public async Task CanUseNetworkInterfacesWithoutReverseDnsOnIP()
...
-        Assert.Equal("abc1", PostConfigureConsulDiscoveryOptions.NormalizeForConsul("abc1", "name"));
-        Assert.Equal("ab-c1", PostConfigureConsulDiscoveryOptions.NormalizeForConsul("ab:c1", "name"));
-        Assert.Equal("ab-c1", PostConfigureConsulDiscoveryOptions.NormalizeForConsul("ab::c1", "name"));
+        Action action = () => PostConfigureConsulDiscoveryOptions.NormalizeForConsul("abc:", "name");
 
-        Assert.Throws<InvalidOperationException>(() => PostConfigureConsulDiscoveryOptions.NormalizeForConsul("9abc", "name"));
-        Assert.Throws<InvalidOperationException>(() => PostConfigureConsulDiscoveryOptions.NormalizeForConsul(":abc", "name"));
-        Assert.Throws<InvalidOperationException>(() => PostConfigureConsulDiscoveryOptions.NormalizeForConsul("abc:", "name"));
+        action.Should().ThrowExactly<InvalidOperationException>();
+    }
```
TimHess, line 164: This might be neater as a theory and easier to expand if we find other cases that should throw

src/Discovery/test/Consul.Test/Util/DateTimeConversionsTest.cs
```diff
@@ -9,20 +9,48 @@ namespace Steeltoe.Discovery.Consul.Test.Util;
...
+        action.Should().ThrowExactly<ArgumentException>();
+    }
+
+    [Fact]
+    public void ToTimeSpan_Invalid_Throws()
+    {
+        Action action = () => DateTimeConversions.ToTimeSpan("foobar");
+
+        action.Should().ThrowExactly<FormatException>();
+    }
```
TimHess, line 42: Another place we _could_ use a Theory

src/Discovery/test/Eureka.Test/AppInfo/ApplicationInfoCollectionTest.cs
```diff
@@ -91,19 +90,12 @@ public void Add_UpdatesExisting_ApplicationMap()
...
+        apps.ApplicationMap.Should().HaveCount(2);
+        ApplicationInfo? app = apps.ApplicationMap.Should().ContainKey("app1".ToUpperInvariant()).WhoseValue;
+        app.Instances.Should().ContainSingle(info => info.InstanceId == "id3");
+        app.Instances.Should().ContainSingle(info => info.InstanceId == "id4");
 
-        foreach (InstanceInfo instance in instances)
-        {
-            Assert.True(instance.InstanceId is "id3" or "id4");
-        }
+        apps.ApplicationMap.Should().ContainKey("app2".ToUpperInvariant());
```
TimHess: I'm assuming it would be excessive, but should we also be asserting that app2's instances weren't unexpectedly changed here?

src/Discovery/test/Eureka.Test/AppInfo/ApplicationInfoCollectionTest.cs
```diff
@@ -381,44 +356,43 @@ public void UpdateFromDelta_AddNewAppNewInstance_UpdatesCorrectly()
...
+        result.Should().Contain(app2.GetInstance("id1")!);
+        result.Should().Contain(app2.GetInstance("id2")!);
 
         result = apps.GetInstancesByVipAddress("vapp3");
-        Assert.NotNull(result);
-        Assert.Single(result);
-        Assert.Contains(app3.GetInstance("id1"), result);
+
+        result.Should().ContainSingle();
+        result.Should().Contain(app3.GetInstance("id1")!);
```
TimHess: Can these two lines be combined?

src/Discovery/test/Eureka.Test/AppInfo/ApplicationInfoCollectionTest.cs
```diff
@@ -582,38 +550,32 @@ public void UpdateFromDelta_ExistingAppWithRemovedInstance_UpdatesCorrectly()
...
 
         result = apps.GetInstancesByVipAddress("vapp2");
-        Assert.NotNull(result);
-        Assert.Single(result);
-        Assert.Contains(app2.GetInstance("id1"), result);
-        Assert.DoesNotContain(app2.GetInstance("id2"), result);
+
+        result.Should().ContainSingle();
+        result.Should().Contain(app2.GetInstance("id1")!);
+        result.Should().NotContain(app2.GetInstance("id2")!);
```
TimHess: If there's only one result and it contains `app2.GetInstance("id1")` then can't it already not contain `app2.GetInstance("id2")`? If so, maybe it could be a one-liner?

src/Discovery/test/Eureka.Test/EurekaApplicationsHealthContributorTest.cs
```diff
@@ -36,18 +36,16 @@ public void GetApplicationsFromConfiguration_ReturnsExpected()
...
-        Assert.NotNull(apps);
-        Assert.NotEmpty(apps);
-        Assert.Equal(3, apps.Count);
-        Assert.Contains("foo", apps);
-        Assert.Contains("bar", apps);
-        Assert.Contains("boo", apps);
+        apps.Should().HaveCount(3);
+        apps.Should().Contain("foo");
+        apps.Should().Contain("bar");
+        apps.Should().Contain("boo");
```
TimHess: Inconsequential, but given the context I would have expected "baz" here, wonder why it's "boo" 👻

src/Discovery/test/Eureka.Test/AppInfo/InstanceInfoTest.cs
```diff
@@ -289,42 +288,40 @@ public void FromInstanceConfiguration_DefaultInstanceOptions_Correct()
...
+        instance.SecureVipAddress.Should().BeNull();
+        instance.CountryId.Should().BeNull();
+        instance.DataCenterInfo.Name.ToString().Should().Be("MyOwn");
+        instance.Status.Should().Be(InstanceStatus.Up);
+        instance.OverriddenStatus.Should().BeNull();
+        instance.LeaseInfo.Should().NotBeNull();
+        instance.LeaseInfo.RenewalInterval.Should().NotBeNull();
+        instance.LeaseInfo.RenewalInterval.Value.TotalSeconds.Should().Be(30);
+        instance.LeaseInfo.Duration.Should().NotBeNull();
+        instance.LeaseInfo.Duration.Value.TotalSeconds.Should().Be(90);
```
TimHess, line 315: Are these 4 lines combine-able into 2? Question applies elsewhere in this file, but I'll avoid making a mess of conversations in case they aren't

TimHess (dismissed):
Thanks for doing this work, the end result is frequently clearer and much easier to understand! Comments should all be treated as optional

src/Discovery/test/Consul.Test/Discovery/PostConfigureConsulDiscoveryOptionsTest.cs
bart-vmware: I've tried, but this is not possible because there is no non-generic overload of `ThrowExactly`.

image

src/Discovery/test/Consul.Test/Util/DateTimeConversionsTest.cs
bart-vmware: This isn't possible for the same reason. The type of exception varies.

src/Discovery/test/Eureka.Test/AppInfo/ApplicationInfoCollectionTest.cs
bart-vmware: Good catch!

src/Discovery/test/Eureka.Test/EurekaApplicationsHealthContributorTest.cs
bart-vmware: I assume "baz" is expected here only because it's the next item in the [well-known sequence](https://en.wikipedia.org/wiki/Foobar)? Otherwise, the test appears to be fine to me.

src/Discovery/test/Eureka.Test/AppInfo/InstanceInfoTest.cs
bart-vmware: Great question. The short answer is: no. Because `NotBeNull()` returns a non-generic `ObjectAssertions`, so we lose the compile-time type being asserted. This is a consequence of limitations in overload resolution.

The long answer is that it's actually possible by adding our own variation(s) of `Should`, which _do_ preserve the type. But each one must be named differently to prevent conflicts in overload resolution. The caller needs to know when to use which variation, for the proper extension methods to be offered. See the [implementation](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/267876985c5457ed1f55514d14c5ea27ade1e916/test/TestBuildingBlocks/FluentExtensions.cs#L74-L81) and [usage](https://github.com/json-api-dotnet/JsonApiDotNetCore/blob/267876985c5457ed1f55514d14c5ea27ade1e916/test/JsonApiDotNetCoreTests/UnitTests/ModelStateValidation/ModelStateValidationTests.cs#L57) for doing it for reference types. We could possibly introduce something similar for value types, such as `.ValueShould()` and `.NullableValueShould()`.

I didn't want to go that far here, to keep things simple. Adding this requires consumers to understand when to use which variation.

src/Discovery/test/Eureka.Test/EurekaApplicationsHealthContributorTest.cs
TimHess: Yeah, I think I've seen baz used in this context for Spring tests/examples, but not boo. I thought it was funny more than anything else

TimHess (approved):

PR #1556, opened 2025-07-30
Updates for Steeltoe v3.3

bart-vmware:
## Description

This PR updates Steeltoe 3.x code with the following changes:
- Added target for `net8.0`, removed target for `netcoreapp3.1` (`net6.0` is still targeted)
  - **Exception:** Integration/Messaging/Stream are incompatible with .NET 8, so they only target .NET 6
- Bumped the version of vulnerable dependent NuGet packages
  - The oldest non-vulnerable version of OpenTelemetry requires .NET 8 versions of core packages like `Microsoft.Extensions.Logging`, so we had to bump those as well.
  - Because only vulnerable versions of `OpenTelemetry.Exporter.Jaeger` exist, Jaeger support has been removed.
- Changed Steeltoe packages author from VMware to Broadcom
- Marked APIs obsolete [whose features are no longer available in Steeltoe v4](https://github.com/SteeltoeOSS/Steeltoe/issues/1244) (see commit messages for affected directories)

> [!IMPORTANT]
> The changes in this PR represent the _minimal effort_ to make all existing tests succeed. This means that Steeltoe 3.x still references very old third-party packages that are likely no longer supported. This is done to minimize the risk associated with existing users updating to this version. For any new developments, please use Steeltoe v4.

Related PR to add the redirect to the documentation website: https://github.com/SteeltoeOSS/Documentation/pull/399.

## Quality checklist

- [x] Your code complies with our [Coding Style](https://github.com/SteeltoeOSS/.github/blob/main/contributing-docs/contributing-code-style.md).
- [x] You've updated unit and/or integration tests for your change, where applicable.
- [ ] You've updated documentation for your change, where applicable.
      If your change affects other repositories, such as [Documentation](https://github.com/SteeltoeOSS/Documentation), [Samples](https://github.com/SteeltoeOSS/Samples) and/or [MainSite](https://github.com/SteeltoeOSS/MainSite), add linked PRs here.
- [ ] There's an open issue for the PR that you are making. If you'd like to propose a new feature or change, please open an issue to discuss the change or find an existing issue.
- [x] You've added required license files and/or file headers (explaining where the code came from with proper attribution), where code is copied from StackOverflow, a blog, or OSS.

PR #1562, opened 2025-08-06
Fix transitive vulnerable dependencies 3x

bart-vmware:
## Description

Adds explicit package references for transient packages that are reported as vulnerable.
Includes a GitHub Actions workflow to detect future vulnerabilities.

The higher versions mostly affect test dependencies, except for the following:
- `Microsoft.IdentityModel.JsonWebTokens` (replacement for `System.IdentityModel.Tokens.Jwt`): Bumped both from 5.2.2 to 7.7.1

The warnings caused by `KubernetesClient` are suppressed, because there's no non-breaking solution. The lowest version without transitive vulnerabilities no longer targets `netstandard2.0`, which breaks upstream Steeltoe packages. Furthermore, that version has breaking API changes that are non-trivial to fix. The suppression works in the .NET CLI, but Visual Studio Package Manager ignores it.

## Quality checklist

### DIFF
--- a/.github/workflows/scan-vulnerable-dependencies.yml
+++ b/.github/workflows/scan-vulnerable-dependencies.yml
@@ -1,0 +1,40 @@
+name: Scan vulnerable dependencies
+
+on:
+  workflow_dispatch:
+  push:
+    branches:
+    - '3.x'
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+env:
+  DOTNET_CLI_TELEMETRY_OPTOUT: 1
+  DOTNET_NOLOGO: true
+  SOLUTION_FILE: 'src/Steeltoe.All.sln'
+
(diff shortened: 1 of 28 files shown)


bart-vmware:
> We should not be introducing explicit dependencies for things available natively from the framework. If necessary they could be conditional so that they are only added on the older TFM'S where necessary or another dependency has it as a dependency for all TFM'S.
> 
> At the same time transitive pinning in conjunction with central package packagement also helps simplify this.

This is not possible in most cases. *Base projects in Steeltoe 3.x reference netstandard, which doesn't provide these packages. The goal here is to ensure non-vulnerable dependencies with _minimal customer impact_, rather than redesigning how dependencies flow in 3.x. It's either this or dropping netstandard and net6.0 entirely, which has way more impact.

See also the explanation at #1556. Our plan is to release a minor 3.x version that works with .NET 6 and .NET 8 (except for Stream/Messaging/Integration) and will be supported for one year. This should enable customers to transition to Steeltoe 4.0, where dependencies are handled in a different manner.

PR #1459, opened 2025-02-10
Run all tests against .NET 8 and 9

bart-vmware:
## Description

Keep Steeltoe targeting .NET 8, but run all tests against both .NET 8 and .NET 9.
Update NuGet packages to latest versions, except for `DiagnosticsTracingVersion`, which has breaking changes.

Changes in Sonar rules between v9.25.0.90414 and v10.6.0.109712:

| Rule                                                              | Title                                                                                     | Sonar v9.25 | Sonar v10.6 | Change   | Steeltoe main | Main uses default? | Steeltoe PR | PR uses default? | Unchanged in PR?      | Comments                                                                                                                                                |
| ----------------------------------------------------------------- | ----------------------------------------------------------------------------------------- | ----------- | ----------- | -------- | ------------- | ------------------ | ----------- | ---------------- | --------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------- |
| [S1694](https://sonarsource.github.io/rspec/#/rspec/S1694/csharp) | An abstract class should have both abstract and concrete methods                          | None        | Warning     | Severity | None          | Yes                | Warning     | Yes              | No (severity changed) | Use new default in Debug/Release, no existing violations                                                                                                |
| [S2053](https://sonarsource.github.io/rspec/#/rspec/S2053/csharp) | Password hashing functions should use an unpredictable salt                               | Warning     |             | Removed  | Warning       | Yes                |             |                  | No (removed rule)     |                                                                                                                                                         |
| [S2222](https://sonarsource.github.io/rspec/#/rspec/S2222/csharp) | Locks should be released on all paths                                                     | Warning     |             | Removed  | Warning       | Yes                |             |                  | No (removed rule)     |                                                                                                                                                         |
| [S2259](https://sonarsource.github.io/rspec/#/rspec/S2259/csharp) | Null pointers should not be dereferenced                                                  | Warning     |             | Removed  | Warning       | Yes                |             |                  | No (removed rule)     |                                                                                                                                                         |
| [S2325](https://sonarsource.github.io/rspec/#/rspec/S2325/csharp) | Methods and properties that don't access instance data should be static                   | None        | Warning     | Severity | None          | Yes                | None        | No               | Yes                   | Keep turned off in Debug/Release, silly rule                                                                                                            |
[...]

PR #1569, opened 2025-08-15
Fixed crash in RabbitMQ connector

bart-vmware:
## Description

Fixed crash during RabbitMQ connect:

> System.InvalidCastException: 'Unable to cast object of type 'AsyncStateMachineBox`1[System.IDisposable,Steeltoe.Connectors.RabbitMQ.RabbitMQServiceCollectionExtensions+<CreateConnectionAsync>d__5]' to type 'RabbitMQ.Client.IConnection'.'

This happened because the connect callback returned a `Task<IConnection>`, whereas the connector is registered in the service container with `IConnection`. There are two solutions:

1. Change the service registration from `ConnectorFactory<RabbitMQOptions, IConnection>` to `ConnectorFactory<RabbitMQOptions, Task<IConnection>>`, such that the call to `connectorFactory.Get().GetConnection()` needs to be awaited.
2. Stick with the current API and synchronously wait for the task to complete before returning it.

This PR implements solution 2, because that doesn't require breaking the public API. I'm also unsure whether it's safe when the task is awaited concurrently.

## Quality checklist

### DIFF
--- a/src/Connectors/src/Connectors/RabbitMQ/DynamicTypeAccess/ConnectionFactoryInterfaceShim.cs
+++ b/src/Connectors/src/Connectors/RabbitMQ/DynamicTypeAccess/ConnectionFactoryInterfaceShim.cs
@@ -11,12 +11,34 @@ internal sealed class ConnectionFactoryInterfaceShim(RabbitMQPackageResolver pac
 {
     private readonly RabbitMQPackageResolver _packageResolver = packageResolver;
 
+    public ConnectionInterfaceShim CreateConnection(CancellationToken cancellationToken)
+    {
+        Task task = InvokeCreateConnectionAsync(cancellationToken);
+
+#pragma warning disable S4462 // Calls to "async" methods should not be blocking
+        // Justification: The service container needs to contain IConnection, not Task<IConnection>.
+        task.Wait(cancellationToken);
+#pragma warning restore S4462 // Calls to "async" methods should not be blocking
+
+        return GetTaskResult(task);
+    }
+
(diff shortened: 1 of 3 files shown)


bart-vmware:
This bug was introduced in #1459.

TimHess:
> 1. Change the service registration from `ConnectorFactory<RabbitMQOptions, IConnection>` to `ConnectorFactory<RabbitMQOptions, Task<IConnection>>`, such that the call to `connectorFactory.Get().GetConnection()` needs to be awaited.
> 2. Stick with the current API and synchronously wait for the task to complete before returning it.
> 
> This PR implements solution 2, because that doesn't require breaking the public API. I'm also unsure whether it's safe when the task is awaited concurrently.

Sticking with the current API seems like the right thing to do, this approach seems right given where we're at with 4.0 and nothing jumps out at me as blockers for merge, but I have a couple questions before approving:

- Is it awkward/unexpected to return the RMQ `ConnectionFactory` and have the caller open the connection on their own?
- Would it be unreasonable to add a `GetConnectionAsync` variant for connectors that have an async way to open the connection?

bart-vmware:
> * Is it awkward/unexpected to return the RMQ `ConnectionFactory` and have the caller open the connection on their own?

Yes, that's an anti-pattern. See https://www.rabbitmq.com/dotnet-api-guide.html#connection-and-channel-lifespan.

> * Would it be unreasonable to add a `GetConnectionAsync` variant for connectors that have an async way to open the connection?

That's basically solution 1, but with a different method name. I'm not sure that awaiting concurrently is thread-safe.

PR #1580, opened 2025-08-28
Enhance JWT issuer validation

TimHess:
## Description

Allow a variation on the authority (that seems to show up only occasionally?) when using the SSO Tile on Cloud Foundry

Note: Steeltoe V3 performed issuer validation [based strictly on the presence of `uaa` in the issuer](https://github.com/SteeltoeOSS/Steeltoe/blob/3.x/src/Security/src/Authentication.CloudFoundryBase/CloudFoundryTokenValidator.cs#L30) 

## Quality checklist

### DIFF
--- a/src/Security/src/Authentication.JwtBearer/PostConfigureJwtBearerOptions.cs
+++ b/src/Security/src/Authentication.JwtBearer/PostConfigureJwtBearerOptions.cs
@@ -5,6 +5,7 @@
 using Microsoft.AspNetCore.Authentication.JwtBearer;
 using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.Options;
+using Steeltoe.Common;
 
 namespace Steeltoe.Security.Authentication.JwtBearer;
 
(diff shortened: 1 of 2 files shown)


src/Security/src/Authentication.JwtBearer/PostConfigureJwtBearerOptions.cs
```diff
@@ -42,7 +43,18 @@ public void PostConfigure(string? name, JwtBearerOptions options)
             return;
         }
 
-        options.TokenValidationParameters.ValidIssuer = $"{options.Authority}/oauth/token";
+        if (Platform.IsCloudFoundry && options.Authority.Contains(".login", StringComparison.OrdinalIgnoreCase))
+        {
+            options.TokenValidationParameters.ValidIssuers =
+            [
+                $"{options.Authority}/oauth/token",
+                $"{options.Authority.Replace(".login", ".uaa", StringComparison.OrdinalIgnoreCase)}/oauth/token"
```
bart-vmware, line 51: Why case-insensitive?

TimHess: It's evaluating a hostname, which wouldn't be case sensitive. I suspect it would always be lowercase, so maybe it doesn't need to be, but I don't know that for sure and figured this wouldn't hurt

bart-vmware (approved):

PR #1589, opened 2025-09-05
Delete nuget.config

bart-vmware:


bart-vmware:
Azure DevOps requires this file to exist, so superseded by #1591.

PR #1598, opened 2025-09-24
Exclude Aspire resolver from Eureka internal HttpClient

bart-vmware:
## Description

Exclude Aspire service discovery resolver from internal HttpClient used by Eureka.

## Quality checklist

### DIFF
--- a/src/Discovery/test/HttpClients.Test/ResolvingHttpDelegatingHandler.cs
+++ b/src/Discovery/test/HttpClients.Test/ResolvingHttpDelegatingHandler.cs
@@ -1,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.ServiceDiscovery.Http;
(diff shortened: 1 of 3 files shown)


src/Discovery/test/HttpClients.Test/ResolvingHttpDelegatingHandler.cs
```diff
@@ -0,0 +1,16 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+#pragma warning disable IDE0130 // Namespace does not match folder structure
+// ReSharper disable once CheckNamespace
+namespace Microsoft.Extensions.ServiceDiscovery.Http;
+#pragma warning restore IDE0130 // Namespace does not match folder structure
+
+internal sealed class ResolvingHttpDelegatingHandler : DelegatingHandler
```
TimHess, line 10: I wondered, but then assumed (and confirmed) the real type is `internal` and that's why this copy is here... We could add a note to eliminate future guessing, but that probably isn't _necessary_

TimHess (approved):

src/Discovery/test/HttpClients.Test/ResolvingHttpDelegatingHandler.cs
bart-vmware: Even if public, we'd need to take a dependency on Aspire, which I tried to avoid.

PR #1604, opened 2025-10-01
Expose secure/non-secure URIs on service instance, optimize Eureka

bart-vmware:
## Description

This PR adds two new nullable properties to `IServiceInstance`: `SecureUri` and `NonSecureUri`.
They are populated for Eureka, because such an instance can contain both.
Furthermore, internal storage for secure/non-secure Eureka VIP addresses has been combined into a single collection, which reduces the need for copying collections. Locking was removed to improve throughput, which comes at the cost of a small increase in memory usage: when all instances of a service are removed, an empty instance collection is kept in memory.

## Quality checklist

### DIFF
--- a/src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
+++ b/src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
@@ -4,6 +4,7 @@
 
 using System.Collections;
 using System.Collections.Concurrent;
+using System.Collections.ObjectModel;
 using System.Text;
 using System.Text.Json;
 using Steeltoe.Common;
@@ -17,11 +18,8 @@ namespace Steeltoe.Discovery.Eureka.AppInfo;
 /// </summary>
 public sealed class ApplicationInfoCollection : IReadOnlyCollection<ApplicationInfo>
 {
-    private readonly object _addRemoveInstanceLock = new();
-
     internal ConcurrentDictionary<string, ApplicationInfo> ApplicationMap { get; } = new();
     internal ConcurrentDictionary<string, ConcurrentDictionary<string, InstanceInfo>> VipInstanceMap { get; } = new();
-    internal ConcurrentDictionary<string, ConcurrentDictionary<string, InstanceInfo>> SecureVipInstanceMap { get; } = new();
 
     public string? AppsHashCode { get; internal set; }
     public long? Version { get; private set; }
@@ -51,18 +49,25 @@ public sealed class ApplicationInfoCollection : IReadOnlyCollection<ApplicationI
         return ApplicationMap.GetValueOrDefault(appName.ToUpperInvariant());
     }
 
-    internal List<InstanceInfo> GetInstancesBySecureVipAddress(string secureVipAddress)
-    {
-        ArgumentException.ThrowIfNullOrWhiteSpace(secureVipAddress);
-
-        return GetByVipAddress(secureVipAddress, SecureVipInstanceMap);
-    }
-
-    internal List<InstanceInfo> GetInstancesByVipAddress(string vipAddress)
+    internal ReadOnlyCollection<InstanceInfo> GetInstancesByVipAddress(string vipAddress)
     {
         ArgumentException.ThrowIfNullOrWhiteSpace(vipAddress);
 
-        return GetByVipAddress(vipAddress, VipInstanceMap);
+        List<InstanceInfo> result = [];
+        string addressUpper = vipAddress.ToUpperInvariant();
+
+        if (VipInstanceMap.TryGetValue(addressUpper, out ConcurrentDictionary<string, InstanceInfo>? instancesById))
+        {
+            foreach (InstanceInfo instance in instancesById.Values)
+            {
+                if ((ReturnUpInstancesOnly && instance.EffectiveStatus == InstanceStatus.Up) || !ReturnUpInstancesOnly)
+                {
+                    result.Add(instance);
+                }
+            }
+        }
+
+        return result.AsReadOnly();
     }
 
     /// <inheritdoc />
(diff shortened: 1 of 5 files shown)


src/Discovery/src/Eureka/EurekaServiceInstance.cs
```diff
@@ -27,24 +46,23 @@ public EurekaServiceInstance(InstanceInfo instance)
...
-        Port = GetPort(instance);
-        IsSecure = instance.IsSecurePortEnabled;
-        Uri = new Uri($"{(IsSecure ? "https" : "http")}://{Host}:{Port}");
         Metadata = instance.Metadata;
-    }
 
-    private static int GetPort(InstanceInfo instance)
-    {
-        if (instance.IsSecurePortEnabled)
+        if (instance is { IsNonSecurePortEnabled: true, NonSecurePort: > 0 })
```
TimHess, line 51: Is there potential for this to conflict with dynamic port assignment? IIRC that would involve configuring the app to listen on `:0`, but I don't know offhand how the timing of this code aligns with that turning into a real port

src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
```diff
@@ -92,79 +97,51 @@ internal void Add(ApplicationInfo app)
...
-                instances = new ConcurrentDictionary<string, InstanceInfo>();
-                dictionary[addressUpper] = instances;
-            }
+        HashSet<string> addresses = [];
 
-            instances[instance.InstanceId] = instance;
+        if (instance.SecureVipAddress != null)
+        {
+            string[] secureAddresses = instance.SecureVipAddress.Split(',', StringSplitOptions.TrimEntries | StringSplitOptions.RemoveEmptyEntries);
+            addresses.UnionWith(secureAddresses);
```
TimHess, line 122: Is there an advantage of using `.UnionWith` here when we know `addresses` is empty, or is it more for alignment with line 128?

src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
```diff
@@ -92,79 +97,51 @@ internal void Add(ApplicationInfo app)
...
         {
-            string addressUpper = address.ToUpperInvariant();
+            string addressUpper = vipAddress.ToUpperInvariant();
 
-            if (dictionary.TryGetValue(addressUpper, out ConcurrentDictionary<string, InstanceInfo>? instances))
+            if (VipInstanceMap.TryGetValue(addressUpper, out ConcurrentDictionary<string, InstanceInfo>? instancesById))
             {
-                _ = instances.TryRemove(instance.InstanceId, out _);
-
-                if (instances.IsEmpty)
```
TimHess, line 164: It looks like we will end up with zombie apps in the internal dictionary that have no instances ([example in the tests](https://github.com/SteeltoeOSS/Steeltoe/pull/1604/files#diff-180e7a90b742496441a438262e20c27594b7bb4dc89857a53bbbbc10cc928aa8R171)) after this change. Is that intended, inconsequential or am I just missing something?

src/Discovery/test/Eureka.Test/EurekaDiscoveryClientTest.cs
```diff
@@ -433,13 +435,13 @@ public async Task GetInstancesByVipAddress_ReturnsExpected()
 
         discoveryClient.Applications = new ApplicationInfoCollection([
             new ApplicationInfo("app1", [
-                new InstanceInfo("id1", "app1", "localhost", "192.168.56.1", new DataCenterInfo(), TimeProvider.System)
+                new InstanceInfo("id11", "app1", "localhost", "192.168.56.1", new DataCenterInfo(), TimeProvider.System)
```
TimHess, line 438: Why change the id of these instances? I'm not super concerned about it, just curious

bart-vmware: Scroll down a bit. The IDs for app2 are named id21/id22, so this change aligns with that pattern.

src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
bart-vmware: Yes, that's what this text in the PR description is about:

> Locking was removed to improve throughput, which comes at the cost of a small increase in memory usage: when all instances of a service are removed, an empty instance collection is kept in memory.

`GetInstancesByVipAddress` loops over the entries, so it doesn't affect the outcome.

src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
bart-vmware: The latter. `HashSet<>.UnionWith` is the equivalent of `List<>.AddRange`. We could lazily initialize, saving an allocation when both addresses are empty, but the common case is for at least one of them to be non-empty, so I don't think such an optimization is useful. If the common case is that only one address is populated, we might possibly save a collection resize; however, I have no evidence to suggest that's the case.

src/Discovery/src/Eureka/EurekaServiceInstance.cs
bart-vmware: I don't expect Eureka to return instances with a zero port, so this only applies to the local instance.
Then yes, but _only_ if all of the following conditions are met:

1. Dynamic port assignment is used, and the app binds to https or http+https
2. `SecurePortEnabled` is explicitly set to `true` in configuration (it defaults to `false`)
3. `EurekaDiscoveryClient.GetLocalServiceInstance()` is called _before_ the app has fully started (so before ASP.NET assigns the dynamic port number)
4. The app doesn't use Aspire (the DCP assigns dynamic port numbers via environment variables upfront)

Before this PR, it would return: `IsSecure: true, Port: 0`.
After this PR, it would return: `IsSecure: false, Port: 0`.

I believe this is an extreme corner case, because why would anyone explicitly configure `SecurePortEnabled: true` (2), knowing that it'll be overwritten by ASP.NET bindings once the app has started. I also don't see the point of calling `EurekaDiscoveryClient.GetLocalServiceInstance()` before the app has fully started (3) in this case, because the port number will always be zero (regardless of whether the changes from this PR are taken or not).

In my opinion, when using dynamic ports, requesting the local instance _before_ the dynamic port number(s) have been assigned should be considered garbage output.

src/Discovery/src/Eureka/AppInfo/ApplicationInfoCollection.cs
```diff
@@ -261,22 +238,4 @@ internal static ApplicationInfoCollection FromJson(JsonApplications? jsonApplica
...
         return apps;
     }
-
-    private List<InstanceInfo> GetByVipAddress(string name, ConcurrentDictionary<string, ConcurrentDictionary<string, InstanceInfo>> dictionary)
-    {
-        List<InstanceInfo> result = [];
-
-        if (dictionary.TryGetValue(name.ToUpperInvariant(), out ConcurrentDictionary<string, InstanceInfo>? instances))
-        {
-            foreach (InstanceInfo instance in instances.Values.ToArray())
```
bart-vmware, line 271: Note I removed the `.ToArray()` call, which is redundant. I've verified that adding/removing items to a concurrent dictionary while enumerating it does not throw an exception (as opposed to a regular dictionary).

```c#
using System.Collections.Concurrent;

var dictionary = new ConcurrentDictionary<string, int>
{
    ["a"] = 1,
    ["b"] = 2,
    ["c"] = 3
};

foreach (var value in dictionary.Values)
{
    dictionary.TryRemove("c", out _);
    dictionary.TryAdd("x", 9);
}
```

PR #1611, opened 2025-10-30
Update dump dependencies, fix gcdump

bart-vmware:
## Description

Updates the dependent packages for taking heap and thread dumps to the latest versions. This fixes https://github.com/dotnet/diagnostics/issues/5508.

This PR also fixes apps that consume Steeltoe and fail to take a GC dump. Steeltoe employs a trick to download the gcdump assembly without referencing it (because adding a PackageReference to a tool package is not possible). The same needs to be done in consuming apps. Tested with both .NET 8 and 9.

image

## Quality checklist

### DIFF
--- a/src/Management/src/Endpoint/Build/Steeltoe.Management.Endpoint.props
+++ b/src/Management/src/Endpoint/Build/Steeltoe.Management.Endpoint.props
@@ -1,0 +1,30 @@
+<Project>
+  <PropertyGroup>
+    <DotnetGCDumpVersion>9.0.652701</DotnetGCDumpVersion>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="dotnet-gcdump" Version="$(DotnetGCDumpVersion)" IncludeAssets="None" ExcludeAssets="All" PrivateAssets="None"
+      GeneratePathProperty="true">
+      <!--
+          The dotnet-gcdump global tool can't be referenced, but we can download it.
+          GeneratePathProperty makes the path to the package available in $(Pkgdotnet-gcdump), so we can add an assembly reference.
+          See https://docs.microsoft.com/en-us/nuget/consume-packages/package-references-in-project-files#generatepathproperty.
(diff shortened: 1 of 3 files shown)


src/Management/src/Endpoint/Build/Steeltoe.Management.Endpoint.props
```diff
@@ -0,0 +1,30 @@
+<Project>
+  <PropertyGroup>
+    <DotnetGCDumpVersion>9.0.652701</DotnetGCDumpVersion>
```
TimHess, line 3: Duplicated rather than referencing versions.props because this piece needs to be included in the shipped package and versions.props isn't? And I suppose also including the maintenance note about multiple locations here would be confusing for package consumers

src/Management/src/Endpoint/Build/Steeltoe.Management.Endpoint.props
```diff
@@ -0,0 +1,30 @@
...
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="dotnet-gcdump" Version="$(DotnetGCDumpVersion)" IncludeAssets="None" ExcludeAssets="All" PrivateAssets="None"
```
TimHess, line 7: Would [PackageDownload](https://learn.microsoft.com/en-us/nuget/consume-packages/packagedownload-functionality) be relevant here?

bart-vmware: It might work, I haven't tried. However, it comes with several limitations. I'd rather stick with the widely-supported `PackageReference` construct.

src/Management/src/Endpoint/Build/Steeltoe.Management.Endpoint.props
bart-vmware: Yes, both are correct.

TimHess (approved):

PR #1615, opened 2025-11-12
Update Steeltoe for .NET 10

bart-vmware:
## Description

Updates the codebase to build using the .NET 10 SDK, multi-target Steeltoe against .NET 8 and 10, and run tests against .NET 8, 9 and 10.

Updates the Steeltoe Spring Boot Configuration Provider to adapt to the breaking change in loading configuration from JSON files, which is documented at https://learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/10.0/configuration-null-values-preserved.

## Quality checklist

### DIFF
--- a/.github/workflows/Steeltoe.All.yml
+++ b/.github/workflows/Steeltoe.All.yml
@@ -27,7 +27,7 @@ env:
 jobs:
   analyze:
     name: Build and Test
-    timeout-minutes: 30
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/sonarcube.yml
+++ b/.github/workflows/sonarcube.yml
@@ -28,7 +28,7 @@ env:
 jobs:
   analyze:
     name: Analyze
-    timeout-minutes: 30
+    timeout-minutes: 60
     runs-on: ubuntu-latest
 
     services:


src/Configuration/src/SpringBoot/SpringBootEnvironmentVariableProvider.cs
```diff
@@ -61,11 +61,8 @@ public override void Load()
             {
                 string? value = Data[key];
 
-                if (value != null)
-                {
-                    string newKey = key.Contains('.') ? key.Replace('.', ':') : key;
-                    data[newKey] = value;
-                }
+                string newKey = key.Contains('.') ? key.Replace('.', ':') : key;
+                data[newKey] = value;
```
TimHess, line 65: This seems like it could be behavioral change. I'm not sure how I feel about it, is it fixing a bug?
Assuming we have something like `My.key.with.null.value=null`...
With the original behavior it looks like the key/value would disappear, but now `My:key:with:null:value=null` will make it into the final configuration?

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -467,14 +481,4 @@ public void Loads_VCAP_SERVICES_from_stream()
         provider.TryGet("p-mysql:1:credentials:uri", out value).Should().BeTrue();
         value.Should().Be("mysql://gxXQb2pMbzFsZQW8:lvMkGf6oJQvKSOwn@192.168.0.97:3306/cf_b2d83697_5fa1_4a51_991b_975c9d7e5515?reconnect=true");
     }
-
-    private sealed class VcapApp
```
TimHess, line 471: Yeah, seems like the right idea to be removing this

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
```diff
@@ -219,12 +225,12 @@ public void Load_VCAP_APPLICATION_Allows_Reload_Without_Throwing_Exception()
...
         {
-            options = configurationRoot.GetSection("vcap:application").Get<VcapApp>();
+            options = configurationRoot.GetSection("vcap:application").Get<CloudFoundryApplicationOptions>();
         }
 
         options.Should().NotBeNull();
-        options.Name.Should().Be("my-app");
-        options.Version.Should().Be("fb8fbcc6-8d58-479e-bcc7-3b4ce5a7f0ca");
+        options.ApplicationVersion.Should().Be("fb8fbcc6-8d58-479e-bcc7-3b4ce5a7f0ca");
+        options.SpaceName.Should().Be("test-space");
```
TimHess, line 233: Any particular reason for using `SpaceName` instead of `ApplicationName` here?

src/Configuration/test/SpringBoot.Test/SpringBootEnvironmentVariableProviderTest.cs
```diff
@@ -63,14 +63,16 @@ public void TryGet_Tree()
         value.Should().Be("q");
 
         provider.TryGet("r", out value).Should().BeTrue();
-        value.Should().BeEmpty();
+        value.Should().BeNull();
```
TimHess, line 66: Is this a side effect of the `FoundationalVersion` change?

src/Connectors/src/Connectors/DynamicTypeAccess/ConnectorFactoryShim.cs
```diff
@@ -60,7 +60,7 @@ private static TypeAccessor MakeGenericTypeAccessor(Type connectionType)
 
     public ConnectorShim<TOptions> Get(string serviceBindingName)
     {
-        object instance = InstanceAccessor.InvokeMethodOverload(nameof(ConnectorFactory<TOptions, object>.Get), true, [typeof(string)], serviceBindingName)!;
+        object instance = InstanceAccessor.InvokeMethodOverload(nameof(ConnectorFactory<,>.Get), true, [typeof(string)], serviceBindingName)!;
```
TimHess, line 63: Cool that the compiler knows how to figure out these types, but it looks a bit mysterious to me now (looking from the github UI). No strong objection, but this one definitely forced a double-take

src/Configuration/test/CloudFoundry.Test/CloudfoundryConfigurationProviderTest.cs
bart-vmware: Yes, because it doesn't work. It's a bug in the .NET configuration binder, see https://github.com/SteeltoeOSS/Steeltoe/blob/3cc666a49bcee54e17416bf0239ab1e747bd84a8/src/Configuration/src/CloudFoundry/ConfigureCloudFoundryApplicationOptions.cs?plain=1#L34-L35.

src/Connectors/src/Connectors/DynamicTypeAccess/ConnectorFactoryShim.cs
bart-vmware: I think you misunderstand. This is not about the compiler infering types. The goal here is to pass the string "Get" via nameof. It's pointless for the compiler to require type arguments, because whatever they are, the method name "Get" isn't affected by that. That compiler limitation has been lifted now.

```c#
// before
int lengthPropertyName1 = nameof(Dictionary<int, string>).Length;
// after
int lengthPropertyName2 = nameof(Dictionary<,>).Length;

```

src/Configuration/src/SpringBoot/SpringBootEnvironmentVariableProvider.cs
bart-vmware: Yes, it's a breaking change in .NET 10, documented [here](https://learn.microsoft.com/en-us/dotnet/core/compatibility/extensions/10.0/configuration-null-values-preserved). I initially made it conditional on the TFM in [this commit](https://github.com/SteeltoeOSS/Steeltoe/pull/1615/commits/49ec3f790e77b1da41bd3a2ce2d37ad58e9a8282). In a later commit, I bumped `FoundationalVersion`, so now it applies unconditionally.

src/Configuration/test/SpringBoot.Test/SpringBootEnvironmentVariableProviderTest.cs
bart-vmware: Yes, it's a result of the breaking change in `Microsoft.Extensions.Configuration.Json.JsonStreamConfigurationProvider` v10.

src/Configuration/src/SpringBoot/SpringBootEnvironmentVariableProvider.cs
bart-vmware: For Steeltoe, the odd behavior in earlier .NET versions required special code on our side to compensate for the oddity. That special code is no longer needed now.

bart-vmware:
I'm going to resolve the open conversations and merge this PR to unblock other work. The raised concerns are intentional, but please let me know if I missed something or misunderstood, so we can revisit later.

PR #1633, opened 2025-12-09
Various test fixes

bart-vmware:
Backported changes from experimenting with running tests on MTPv2.

- Fixes test failures when running in WSL2 on Ubuntu
  Make assertions ignore differences in line endings, originating from multi-line raw string literals. The code is _compiled_ using the Windows newline separator, which differs from the newline separator used when _executing_ on Ubuntu.
- Explicitly target VSTest mode in xUnit
  ~Reference `xunit.v3.mtp-off`, so we don't need `DisableTestingPlatformServerCapability=false` anymore.~
  Edit: Reverted back to reference `xunit.v3`, because `xunit.v3.mtp-off` [breaks running tests in WSL](https://github.com/xunit/xunit/issues/3457#issuecomment-3645960475).
- Adds a workaround to prevent taking a dump on hanging/crashing test on macOS from timing out
- Reduce test hang timeout from 3 to 1 minutes
  A hanging test on all three frameworks took 9 minutes, leaving insufficient time to upload the dump in component builds (which time out after 15 minutes).
- Always write test output to a subdirectory, instead of the checkout root
- Run tests for all frameworks in a single command (the bug ignoring `TestTfmsInParallel` has been fixed)
- Replace `skipFilter` with a boolean, which separates intent from implementation
- Run memory dump tests only for `Management.Endpoint.Test` project to speed up runs
- Add diagnostics to investigate intermittent hostname lookup failures on macOS
  Guarded by an environment variable that is set in GHA, to not influence existing production behavior.

### DIFF
--- a/shared-test.props
+++ b/shared-test.props
@@ -17,7 +17,7 @@
     <PackageReference Include="Microsoft.Extensions.TimeProvider.Testing" Version="$(FoundationalVersion)" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestSdkVersion)" />
     <PackageReference Include="Moq" Version="$(MoqVersion)" />
-    <PackageReference Include="xunit.v3.mtp-off" Version="$(XunitVersion)" />
+    <PackageReference Include="xunit.v3" Version="$(XunitVersion)" />
     <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitVisualStudioVersion)" />
   </ItemGroup>
 


src/Common/src/Common/Net/DomainNameResolver.cs
```diff
@@ -9,6 +9,8 @@ namespace Steeltoe.Common.Net;
 
 internal sealed class DomainNameResolver : IDomainNameResolver
 {
+    private static readonly bool IsInDiagnosticsMode = Environment.GetEnvironmentVariable("STEELTOE_MACOS_DIAGNOSE_HOSTNAME_LOOKUP") == "true";
```
TimHess, line 12: This was added for diagnosing issues on Mac, but it looks like it is set for and potentially evaluated on any OS... Should it be named `STEELTOE_DIAGNOSE_HOSTNAME_LOOKUP`? Maybe not worth changing if we're only actually hitting it on mac, especially if you expect this to be temporary

src/Management/test/Endpoint.Test/Actuators/RouteMappings/RazorPagesWebApplicationFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+
+        string absoluteContentRoot = System.Environment.CurrentDirectory;
+        absoluteContentRoot = absoluteContentRoot.Replace($"/{testAssemblyName}/", $"/{appAssemblyName}/", StringComparison.Ordinal);
+        absoluteContentRoot = absoluteContentRoot.Replace($@"\{testAssemblyName}\", $@"\{appAssemblyName}\", StringComparison.Ordinal);
+
+        // Workaround for https://github.com/dotnet/aspnetcore/issues/55867.
+        builder.UseContentRoot(absoluteContentRoot);
+
+        // Workaround for https://github.com/dotnet/aspnetcore/issues/55867#issuecomment-3046941805.
+        builder.UseEnvironment("Production");
```
TimHess, line 27: I don't see the connection with that issue, and I thought "Production" is the default? I don't see a problem with this line, but also don't see the point

TimHess (approved):
Just a couple nits, otherwise looks good to me

src/Common/src/Common/Net/DomainNameResolver.cs
bart-vmware: I chose this name to express intent, and yes, it is temporary.

src/Management/test/Endpoint.Test/Actuators/RouteMappings/RazorPagesWebApplicationFactory.cs
bart-vmware: It ensures we don't hit [this code path](https://github.com/dotnet/aspnetcore/blob/release/10.0/src/DefaultBuilder/src/WebHost.cs#L225-L228) during test execution. Without it, the test fails with the following stack trace:

```
  Message: 
System.ArgumentException : The path must be absolute. (Parameter 'root')

  Stack Trace: 
PhysicalFileProvider.ctor(String root, ExclusionFilters filters)
<>c.<UseStaticWebAssetsCore>b__1_0(String contentRoot)
ManifestStaticWebAssetFileProvider.ctor(StaticWebAssetManifest manifest, Func`2 fileProviderFactory)
StaticWebAssetsLoader.UseStaticWebAssetsCore(IWebHostEnvironment environment, Stream manifest)
StaticWebAssetsLoader.UseStaticWebAssets(IWebHostEnvironment environment, IConfiguration configuration)
BootstrapHostBuilder.RunDefaultCallbacks()
WebApplicationBuilder.InitializeHosting(BootstrapHostBuilder bootstrapHostBuilder)
WebApplicationBuilder.ctor(WebApplicationOptions options, Action`1 configureDefaults)
WebApplication.CreateBuilder(String[] args)
Program.<Main>$(String[] args) line 7
<6 more frames...>
DeferredHostBuilder.Build()
WebApplicationFactory`1.CreateHost(IHostBuilder builder)
WebApplicationFactory`1.ConfigureHostBuilder(IHostBuilder hostBuilder)
WebApplicationFactory`1.StartServer()
WebApplicationFactory`1.CreateDefaultClient(DelegatingHandler[] handlers)
WebApplicationFactory`1.CreateDefaultClient(Uri baseAddress, DelegatingHandler[] handlers)
WebApplicationFactory`1.CreateClient(WebApplicationFactoryClientOptions options)
WebApplicationFactory`1.CreateClient()
RazorPagesExternalAppTest.Can_get_routes_for_razor_pages() line 17
--- End of stack trace from previous location ---
```

Debugging without this workaround, the environment turns out to be Development:

image

The link to the issue describes _what_ we're trying to workaround. I just found a different (easier) way to work around the problem detailed there.

bart-vmware: I've found a way to make tests in WSL work; pushed another commit for that.

bart-vmware: To make all tests succeed in WSL2, https://github.com/SteeltoeOSS/Samples/pull/429 is needed.

TimHess: I was initially confused because ASP.NET Core uses `Production` by default - after some searching I found that `WebApplicationFactory` sets the environment to `Development`, so now I understand

TimHess (approved):

PR #1644, opened 2026-01-27
Enable skipping ASP.NET health checks at actuator endpoint

bart-vmware:
## Description

Adds detection of the "SkipFromHealthActuator" tag in ASP.NET health checks to enable skipping them from the `actuator/health` endpoint.

This enables adding a health check to verify that an app is registered as healthy in Spring Boot Admin. Because SBA queries the health actuator to determine whether the app is healthy, we'd otherwise end up in a chicken-and-egg problem: SBA observes an unhealthy app because the app hasn't been registered as healthy in SBA yet.

## Quality checklist

### DIFF
--- a/src/Common/src/Common/HealthChecks/HealthAggregator.cs
+++ b/src/Common/src/Common/HealthChecks/HealthAggregator.cs
@@ -69,7 +69,7 @@ internal sealed class HealthAggregator : IHealthAggregator
         ICollection<HealthCheckRegistration> healthCheckRegistrations, IServiceProvider serviceProvider, CancellationToken cancellationToken)
     {
         HealthCheckRegistration[] activeHealthCheckRegistrations =
-            healthCheckRegistrations.Where(registration => !registration.Tags.Contains("SkipFromHealthActuator")).ToArray();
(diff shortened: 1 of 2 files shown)


src/Common/src/Common/HealthChecks/HealthAggregator.cs
```diff
@@ -68,16 +68,18 @@ await Parallel.ForEachAsync(contributors, cancellationToken, async (contributor,
     private static async Task<IDictionary<string, SteeltoeHealthCheckResult>> AggregateMicrosoftHealthChecksAsync(ICollection<IHealthContributor> contributors,
         ICollection<HealthCheckRegistration> healthCheckRegistrations, IServiceProvider serviceProvider, CancellationToken cancellationToken)
     {
-        if (healthCheckRegistrations.Count == 0)
+        HealthCheckRegistration[] activeHealthCheckRegistrations =
+            healthCheckRegistrations.Where(registration => !registration.Tags.Contains("SkipFromHealthActuator")).ToArray();
```
TimHess: What do you think about using `ExcludeFromHealthActuator` instead of Skip?

bart-vmware: I'm fine with that; pushed.

Issue #970, opened 2022-08-01
Use System.Text.Json source generator


TimHess:

TimHess:
There may be some maintenance difficulties with this idea, we need to do more research on the level of IDE support for this feature and whether perf gains are even significant. This might be one of those things that's better in theory than reality

bart-vmware:
In .NET 9, the usage of source generators causes errors in code analyzers, which is explained at https://github.com/dotnet/roslyn/issues/75483#issuecomment-2502565487. To summarize: the experience currently is crippled, in an attempt to address performance issues during build.

bart-vmware:
Reopened because #1652 incorrectly referenced this issue.

Issue #969, opened 2022-08-01
Use Compile-time logging source generation


TimHess:
For high performance logging
https://docs.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

bart-vmware:
Should also turn on the source generator for regular expressions.

bart-vmware:
In .NET 9, the usage of source generators causes errors in code analyzers, which is explained at https://github.com/dotnet/roslyn/issues/75483#issuecomment-2502565487. To summarize: the experience currently is crippled, in an attempt to address performance issues during build.

bart-vmware:
Fixed in #1652.

PR #1652, opened 2026-02-26
Use source generators for logging and regular expressions

bart-vmware:
## Description

This PR updates logging and regular expressions to use the source generators provided by the .NET runtime.
- Log messages are cleaner now, but this PR does not change their levels nor the locations where logging happens.
- Optimized the flags provided to regular expressions.

Additionally, I updated the Sonar analyzer because new violations were only reported in Cloud (which always uses the latest version). This was caused by false-negative fixes in existing rules; no new rules were added. Also, updated R# to the latest version.

Reviewing per commit is recommended.

Closes ~#970~ #969.

## Quality checklist

### DIFF
--- a/src/Bootstrap/src/AutoConfiguration/BootstrapScanner.cs
+++ b/src/Bootstrap/src/AutoConfiguration/BootstrapScanner.cs
@@ -38,7 +38,7 @@ using Steeltoe.Management.Tracing;
 
 namespace Steeltoe.Bootstrap.AutoConfiguration;
 
-internal sealed class BootstrapScanner
+internal sealed partial class BootstrapScanner
 {
     private readonly HostBuilderWrapper _wrapper;
     private readonly AssemblyLoader _loader;
@@ -93,21 +93,21 @@ internal sealed class BootstrapScanner
     {
         _wrapper.AddConfigServer(_loggerFactory);
 
-        _logger.LogInformation("Configured Config Server configuration provider");
+        LogConfigServerConfigured();
     }
 
     private void WireCloudFoundryConfiguration()
     {
         _wrapper.AddCloudFoundryConfiguration(_loggerFactory);
 
-        _logger.LogInformation("Configured Cloud Foundry configuration provider");
+        LogCloudFoundryConfigured();
     }
 
     private void WireRandomValueProvider()
     {
         _wrapper.ConfigureAppConfiguration(configurationBuilder => configurationBuilder.AddRandomValueSource(_loggerFactory));
 
-        _logger.LogInformation("Configured random value configuration provider");
+        LogRandomValueConfigured();
     }
 
     private void WireSpringBootProvider()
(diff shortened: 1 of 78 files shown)


src/Bootstrap/src/AutoConfiguration/BootstrapScanner.cs
```diff
@@ -281,4 +281,70 @@ private void WireIfAnyLoaded(Action wireAction, params PackageResolver[] package
             wireAction();
         }
     }
+
+    [LoggerMessage(Level = LogLevel.Information, Message = "Configured Config Server configuration provider.")]
+    private partial void LogConfigServerConfigured();
```
TimHess, line 286: I've seen examples of putting these methods into a separate file, do you think it would be worth the effort to do that here? Many of these files with just one or two log lines per class I'd say no, but this class has a fair number

src/Configuration/src/Encryption/DecryptionConfigurationSource.cs
```diff
@@ -35,5 +35,5 @@ public IConfigurationProvider Build(IConfigurationBuilder builder)
     }
 
     [LoggerMessage(Level = LogLevel.Trace, Message = "Build for {SourceCount} sources and {PropertyCount} properties.")]
-    private static partial void LogBuild(ILogger<DecryptionConfigurationSource> logger, int sourceCount, int propertyCount);
+    private static partial void LogBuild(ILogger logger, int sourceCount, int propertyCount);
```
TimHess, line 38: Why do some of these have an ILogger param and some don't?

bart-vmware: Because the source generator requires that. For example, if the type has multiple `ILogger` fields, or both an `ILogger` and `ILoggerFactory` field.

src/Bootstrap/src/AutoConfiguration/BootstrapScanner.cs
bart-vmware: That typically happens to keep the file from becoming very large. There isn't much code in this file, and, from what I recall, it's the only one with quite a few of these methods.

src/Discovery/test/Eureka.Test/Transport/EurekaClientTest.cs
```diff
@@ -128,6 +128,35 @@ public sealed class EurekaClientTest
         }
         """;
 
+    private static readonly string ExpectedJsonRequestBody = """
```
TimHess, line 131: Nice little optimization here for both readability and DRY

src/Management/src/Prometheus/PrometheusExtensions.cs
```diff
@@ -30,7 +30,7 @@ public static class PrometheusExtensions
     /// </returns>
     public static IServiceCollection AddPrometheusActuator(this IServiceCollection services)
     {
-        return AddPrometheusActuator(services, true, null);
+        return services.AddPrometheusActuator(true, null);
```
TimHess: The changes in this file look like they result in the loss of several null checks

TimHess (changes requested):

src/Management/src/Prometheus/PrometheusExtensions.cs
bart-vmware: I don't see changes in null checks, but the code switched to extension method syntax for private methods. I will revert that.

TimHess (approved):

PR #1656, opened 2026-03-06
Use json source generator

### LINKED ISSUE #970: Use System.Text.Json source generator

bart-vmware:
## Description

This PR migrates System.Text.Json serialization from reflection-based to compile-time source-generated `JsonSerializerContext` classes across the non-actuator parts of the codebase (actuator endpoints are excluded because their serializer options are configurable at runtime).

### Key changes:

1. Source-generated JSON serialization -- Bew `JsonSerializerContext` subclasses replace static `JsonSerializerOptions` instances and update (de)serialization calls throughout Eureka, HttpClients, Cloud Foundry middleware, Management port middleware, ~and SpringBootAdmin client~.

1. Explicit `[JsonPropertyName]` on all serialized properties that are exposed at actuator endpoints -- Guards against wire-level breaking changes when users configure an alternate casing policy, since the property names are dictated by Spring.

1. Cloud Foundry permissions reworked -- ~Switched from v2 to v3 cloud controller API~, replaced loose `Dictionary<string, JsonElement>` parsing with a strongly-typed `PermissionsResponse`, and improved error handling in parsing Cloud Controller responses (graceful failure instead of throwing `SecurityException`). Test mock scenarios were renamed for clarity ("success" -> "full-permissions", "no_sensitive_data" -> "restricted-permissions") and new scenarios added for "no-permissions" and "broken-response".

1. Loggers actuator bug fixes -- Now respects configured serializer options when reading request bodies, corrects nullability, ~and no longer assumes RESET when the request body can't be parsed~.

1. Test improvements -- Expanded serialization roundtrip tests for Eureka types, added `ToString()` debug output coverage, updated CloudFoundry mock URLs and scenarios to match the new permission logic.

Closes #970.

## Quality checklist


src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminJsonSerializerContext.cs
```diff
@@ -0,0 +1,29 @@
...
+
+using System.Text.Json;
+using System.Text.Json.Serialization;
+using System.Text.Json.Serialization.Metadata;
+
+namespace Steeltoe.Management.Endpoint.SpringBootAdminClient;
+
+[JsonSourceGenerationOptions]
+[JsonSerializable(typeof(Application))]
+[JsonSerializable(typeof(SpringBootAdminApiClient.RegistrationResult))]
```
TimHess: Probably not a big deal, but the PR description says it adds `[JsonPropertyName]` to _all_ serialized properties to guard against wire-level breaking changes. But `RegistrationResult.Id` doesn't have one.

src/Discovery/src/HttpClients/LoadBalancers/ServiceInstancesResolver.cs
TimHess, line 158: Should these properties have `[JsonPropertyName]` set?

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
```diff
@@ -33,7 +33,9 @@ public async Task EmptyTokenIsUnauthorized()
     }
 
     [Theory]
+    [InlineData(false, false, EndpointPermissions.None)]
     [InlineData(false, true, EndpointPermissions.Restricted)]
+    [InlineData(true, false, EndpointPermissions.None)]
```
TimHess, line 38: Were you able to find out if this scenario is possible? It does seem implausible, so probably doesn't need extensive research

src/Management/src/Endpoint/Actuators/Loggers/LoggersEndpointMiddleware.cs
```diff
@@ -31,35 +31,37 @@ internal sealed partial class LoggersEndpointMiddleware(
...
 
-                Dictionary<string, string> change = await DeserializeRequestAsync(httpContext.Request.Body, cancellationToken);
+                Dictionary<string, string?> changes = await DeserializeRequestAsync(httpContext.Request.Body, cancellationToken);
 
-                change.TryGetValue("configuredLevel", out string? level);
-
-                LogChangeRequest(loggerName, level ?? "RESET");
-
-                if (!string.IsNullOrEmpty(loggerName))
+                if (changes.TryGetValue("configuredLevel", out string? level))
```
TimHess: This seems like a good and safe change, but just wanted to confirm you validated that both SBA and Apps Manager send the expected request body and there aren't real cases of an actuator client sending an empty body here?

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
```diff
@@ -91,34 +93,31 @@ public async Task<SecurityResult> GetPermissionsAsync(string accessToken, Cancel
...
+                _ => EndpointPermissions.None
+            };
+
+            LogPermissions(permissions);
+            return permissions;
         }
         catch (Exception exception) when (!exception.IsCancellation())
         {
-            throw new SecurityException($"Exception extracting permissions from json: {SecurityUtilities.SanitizeInput(json)}", exception);
+            return null;
```
TimHess, line 119: If there's an exception during deserialization, it will be swallowed and the result will be `Bad Gateway`... While the scenario is pretty unlikely, and it will be pretty clear what happened, this could leave anybody trying to diagnose the situation with a more work to do trying to understand why it happened since they won't be able to see what that response body was

TimHess (dismissed):
Generally LGTM, some nice improvements! Comments can be resolved if you don't think changes are needed for any of them

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
bart-vmware: I haven't looked into it, just trying to be defensive by default.

src/Management/src/Endpoint/Actuators/CloudFoundry/PermissionsProvider.cs
bart-vmware: Yes, they need to turn on debug logging. This is a scenario that should never happen, unless the Cloud Controller API changes or misbehaves. An API change requires a major version bump, while a misbehaving CC is a bug that Broadcom needs to investigate. I wouldn't expect end-users to go that deep.

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminJsonSerializerContext.cs
bart-vmware: Hmm, Cursor initially generated a very detailed description for the PR, so I asked to simplify it. I'll update the description to indicate that this matters to users who pass their own casing convention via configuration, which is only possible for actuator endpoints.

src/Discovery/src/HttpClients/LoadBalancers/ServiceInstancesResolver.cs
bart-vmware: Same, this only matters for data exposed at actuator endpoints. Serializer options are not configurable here.

src/Management/src/Endpoint/Actuators/Loggers/LoggersEndpointMiddleware.cs
bart-vmware: I didn't verify; turns out it works in Apps Manager, but not in SBA. According to the [Spring Boot actuator docs](https://docs.spring.io/spring-boot/api/rest/actuator/loggers.html#loggers.clearing-level), sending an empty JSON object is the correct way to reset, so I've reverted this change.

src/Management/test/Endpoint.Test/Actuators/CloudFoundry/PermissionsProviderTest.cs
bart-vmware: I have reverted the change to use v3 because it breaks Apps Manager.

src/Management/src/Endpoint/Actuators/Loggers/LoggersEndpointMiddleware.cs
bart-vmware: I also reverted source generator usage for SBA. Turns out there's a bug in the runtime: serialization using a `Stream` only succeeds the first 10 times. After that, a fast-path kicks in that breaks it. So, after running for a few minutes with SBA, re-registration starts failing.

src/Management/test/Endpoint.Test/Actuators/Loggers/LoggersActuatorSerilogTest.cs
```diff
@@ -214,11 +214,8 @@ public async Task Can_change_minimum_levels_with_serilog()
             }
             """);
 
-        HttpResponseMessage resetResponse = await httpClient.PostAsync(new Uri("http://localhost/actuator/loggers/Fake.Category"), new StringContent("""
-            {
-                "configuredLevel": null
```
TimHess, line 219: One of the two Actuator clients we're concerned with is still sending this body, shouldn't we maintain at least one test that sends `"configuredLevel": null`?

src/Management/src/Endpoint/Actuators/Loggers/LoggersEndpointMiddleware.cs
TimHess: > I also reverted source generator usage for SBA. Turns out there's a bug in the runtime: serialization using a `Stream` only succeeds the first 10 times. After that, a fast-path kicks in that breaks it. So, after running for a few minutes with SBA, re-registration starts failing.

Is the runtime bug already logged? We could add a follow-up issue to revisit when a fix ships

src/Management/test/Endpoint.Test/Actuators/Loggers/LoggersActuatorSerilogTest.cs
bart-vmware: Turns out nobody actually sends that. Apps Manager sends an empty request body to reset.

```
[actuator-api-management-sample,b42d540512ba4ea74fbb1ebfbd2ecb33,fbfe1af59e3637be,4fbb1ebfbd2ecb33,false] --> REQUEST BODY: {"configuredLevel":"ERROR"}
...
[actuator-api-management-sample,e3681b0ab29c44d8732750740557549e,0f983143c692c19c,732750740557549e,false] --> REQUEST BODY:
```

src/Management/src/Endpoint/Actuators/Loggers/LoggersEndpointMiddleware.cs
bart-vmware: Created https://github.com/dotnet/runtime/issues/125375.

src/Management/test/Endpoint.Test/Actuators/Loggers/LoggersActuatorSerilogTest.cs
TimHess: Weird, that's not what the browser shows on an Apps Manager reset:
image

bart-vmware: You're right, my logging wasn't implemented correctly. I've pushed a commit that adds a test for Apps Manager.

TimHess (approved):

PR #1666, opened 2026-03-30
Fix Eureka Dynamic Port Assignment overriding user-configured ports

TimHess:
## Description

- Detect explicitly configured `eureka:instance:port` or `eureka:instance:securePort` to avoid overwriting.
- Add public `UseAspNetCoreUrls` property (default: `true`, to match Consul) to allow disabling dynamic port detection.

Fixes #1665

## Quality checklist

### DIFF
--- a/src/Discovery/src/Consul/Configuration/ConsulDiscoveryOptions.cs
+++ b/src/Discovery/src/Consul/Configuration/ConsulDiscoveryOptions.cs
@@ -176,5 +176,9 @@ public sealed class ConsulDiscoveryOptions
     /// <summary>
     /// Gets or sets a value indicating whether to register with the port number ASP.NET Core is listening on. Default value: true.
     /// </summary>
+    /// <remarks>
+    /// This property is ignored when <see cref="Port" /> or <see cref="Scheme" /> is explicitly configured, or when <see cref="UseNetworkInterfaces" /> is
+    /// <c>true</c>.
+    /// </remarks>
     public bool UseAspNetCoreUrls { get; set; } = true;
 }
--- a/src/Discovery/src/Consul/Configuration/ConsulDiscoveryOptions.cs
+++ b/src/Discovery/src/Consul/Configuration/ConsulDiscoveryOptions.cs
@@ -175,10 +175,9 @@ public sealed class ConsulDiscoveryOptions
 
     /// <summary>
     /// Gets or sets a value indicating whether to register with the port number ASP.NET Core is listening on. Default value: true.
-    /// </summary>
-    /// <remarks>
+    /// <para />
     /// This property is ignored when <see cref="Port" /> or <see cref="Scheme" /> is explicitly configured, or when <see cref="UseNetworkInterfaces" /> is
     /// <c>true</c>.
-    /// </remarks>
+    /// </summary>
     public bool UseAspNetCoreUrls { get; set; } = true;
 }
--- a/src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
+++ b/src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
@@ -20,14 +20,13 @@ public sealed partial class EurekaInstanceOptions
     internal const string DefaultStatusPageUrlPath = "/info";
     internal const string DefaultHealthCheckUrlPath = "/health";
 
-#pragma warning disable S1067 // Expressions should not be too complex
-    internal bool ShouldSetPortsFromListenAddresses =>
-        UseAspNetCoreUrls && (!Platform.IsCloudFoundry || IsContainerToContainerMethod() || IsForceHostNameMethod()) &&
-        (!IsNonSecurePortEnabled || NonSecurePort == null) && (!IsSecurePortEnabled || SecurePort == null);
-#pragma warning restore S1067 // Expressions should not be too complex
+    private bool IsSuitableRegistrationMethod => !Platform.IsCloudFoundry || IsContainerToContainerMethod() || IsForceHostNameMethod();
+    internal bool ShouldSetPortsFromListenAddresses => UseAspNetCoreUrls && IsSuitableRegistrationMethod && !IsPortConfigured;
 
     internal TimeSpan LeaseRenewalInterval => TimeSpan.FromSeconds(LeaseRenewalIntervalInSeconds);
     internal TimeSpan LeaseExpirationDuration => TimeSpan.FromSeconds(LeaseExpirationDurationInSeconds);
+
+    internal bool IsPortConfigured { get; set; }
 
     /// <summary>
     /// Gets or sets the unique ID (within the scope of the app name) of the instance to be registered with Eureka.
@@ -211,10 +210,9 @@ public sealed partial class EurekaInstanceOptions
 
     /// <summary>
     /// Gets or sets a value indicating whether to register with the port number(s) ASP.NET Core is listening on. Default value: true.
-    /// </summary>
-    /// <remarks>
+    /// <para />
     /// This property is ignored when <see cref="NonSecurePort" /> or <see cref="SecurePort" /> is explicitly configured.
-    /// </remarks>
+    /// </summary>
     public bool UseAspNetCoreUrls { get; set; } = true;
 
     /// <summary>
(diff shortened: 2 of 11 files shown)


src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -205,6 +209,11 @@ public sealed partial class EurekaInstanceOptions
         Name = DataCenterName.MyOwn
     };
 
+    /// <summary>
+    /// Gets or sets a value indicating whether to register with the port number ASP.NET Core is listening on. Default value: true.
```
bart-vmware:
```suggestion
    /// Gets or sets a value indicating whether to register with the port number(s) ASP.NET Core is listening on. Default value: true.
```

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -250,52 +259,59 @@ internal void SetPortsFromListenAddresses(IEnumerable<string> listenOnAddresses,
                 }
             }
 
-            int? nonSecurePort = IsNonSecurePortEnabled ? NonSecurePort : null;
-            int? securePort = IsSecurePortEnabled ? SecurePort : null;
-
-            if (nonSecurePort != listenHttpPort)
+            if (!IsNonSecurePortConfigured)
```
bart-vmware: Do we still want to pick up http ports from ASP.NET when only an https port is configured (and the other way around)? It would make more sense to me to bail out entirely if a http and/or https port is explicitly configured.

src/Discovery/src/Eureka/EurekaServiceCollectionExtensions.cs
```diff
@@ -81,7 +81,24 @@ private static void ConfigureEurekaClientOptions(IServiceCollection services)
...
     {
-        services.AddOptions<EurekaInstanceOptions>().BindConfiguration(EurekaInstanceOptions.ConfigurationPrefix);
+        OptionsBuilder<EurekaInstanceOptions> optionsBuilder = services.AddOptions<EurekaInstanceOptions>();
+        optionsBuilder.BindConfiguration(EurekaInstanceOptions.ConfigurationPrefix);
+
+        optionsBuilder.Configure<IConfiguration>((options, configuration) =>
+        {
+            IConfigurationSection instanceSection = configuration.GetSection(EurekaInstanceOptions.ConfigurationPrefix);
+
+            if (instanceSection.GetSection("Port").Exists())
```
bart-vmware: By reading directly from `IConfiguration`, all changes from user-defined options configurer are effectively ignored. This is why existing logic uses `IPostConfigureOptions`, which runs after user-defined configurers.

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
TimHess: Probably not, updated

src/Discovery/test/Eureka.Test/DynamicPortAssignmentTest.cs
```diff
@@ -38,4 +39,109 @@ public async Task Applies_dynamically_assigned_ports_after_startup()
...
+        builder.Configuration.AddInMemoryCollection(appSettings);
+        builder.Services.AddEurekaDiscoveryClient();
+
+        await using WebApplication app = builder.Build();
+        await app.StartAsync(TestContext.Current.CancellationToken);
+
+        var infoManager = app.Services.GetRequiredService<EurekaApplicationInfoManager>();
+
+        infoManager.Instance.IsSecurePortEnabled.Should().BeTrue();
+        infoManager.Instance.SecurePort.Should().Be(443);
```
bart-vmware, line 98: Should also assert that the non-secure wasn't turned on. Similar assertions for the other tests.

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -205,6 +207,11 @@ public sealed partial class EurekaInstanceOptions
         Name = DataCenterName.MyOwn
     };
 
+    /// <summary>
+    /// Gets or sets a value indicating whether to register with the port number(s) ASP.NET Core is listening on. Default value: true.
```
bart-vmware, line 213: We should document the conditions when this property is ignored (also for Consul)

src/Discovery/src/Eureka/PostConfigureEurekaInstanceOptions.cs
```diff
@@ -140,7 +140,12 @@ private void SetPorts(EurekaInstanceOptions options)
             options.IsSecurePortEnabled = true;
         }
 
-        if (options.NonSecurePort == null && options.SecurePort == null)
+        if (options.NonSecurePort != null || options.SecurePort != null)
```
bart-vmware: Should also check the related IsEnabled properties.

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -20,11 +20,13 @@ public sealed partial class EurekaInstanceOptions
     internal const string DefaultStatusPageUrlPath = "/info";
     internal const string DefaultHealthCheckUrlPath = "/health";
 
-    private bool UseAspNetCoreUrls => !Platform.IsCloudFoundry || IsContainerToContainerMethod() || IsForceHostNameMethod();
+    private bool ShouldUseAspNetCoreUrls => UseAspNetCoreUrls && (!Platform.IsCloudFoundry || IsContainerToContainerMethod() || IsForceHostNameMethod());
 
     internal TimeSpan LeaseRenewalInterval => TimeSpan.FromSeconds(LeaseRenewalIntervalInSeconds);
     internal TimeSpan LeaseExpirationDuration => TimeSpan.FromSeconds(LeaseExpirationDurationInSeconds);
 
+    internal bool IsPortConfigured { get; set; }
```
bart-vmware: It looks like we don't need this property anymore. It can be combined into `ShouldUseAspNetCoreUrls` (renamed to `ShouldSetPortsFromListenAddresses`), which can act as a condition to enter `SetPortsFromListenAddresses` from both call sites (similar to Consul).

src/Discovery/src/Consul/Configuration/ConsulDiscoveryOptions.cs
```diff
@@ -176,5 +176,9 @@ public sealed class ConsulDiscoveryOptions
     /// <summary>
     /// Gets or sets a value indicating whether to register with the port number ASP.NET Core is listening on. Default value: true.
     /// </summary>
+    /// <remarks>
+    /// This property is ignored when <see cref="Port" /> or <see cref="Scheme" /> is explicitly configured, or when <see cref="UseNetworkInterfaces" /> is
+    /// <c>true</c>.
+    /// </remarks>
```
bart-vmware: Unfortunately, remarks don't propagate to the JSON configuration schema (otherwise the build would be broken without a related update). Can you move this into the summary instead?

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -205,6 +209,14 @@ public sealed partial class EurekaInstanceOptions
         Name = DataCenterName.MyOwn
     };
 
+    /// <summary>
+    /// Gets or sets a value indicating whether to register with the port number(s) ASP.NET Core is listening on. Default value: true.
+    /// </summary>
+    /// <remarks>
+    /// This property is ignored when <see cref="NonSecurePort" /> or <see cref="SecurePort" /> is explicitly configured.
+    /// </remarks>
```
bart-vmware: Same concern about the JSON configuration schema.

src/Discovery/test/Eureka.Test/DynamicPortAssignmentTest.cs
```diff
@@ -38,4 +39,112 @@ public async Task Applies_dynamically_assigned_ports_after_startup()
...
+
+        await using WebApplication app = builder.Build();
+        await app.StartAsync(TestContext.Current.CancellationToken);
+
+        var infoManager = app.Services.GetRequiredService<EurekaApplicationInfoManager>();
+
+        infoManager.Instance.IsNonSecurePortEnabled.Should().BeFalse();
+        infoManager.Instance.NonSecurePort.Should().Be(0);
+        infoManager.Instance.IsSecurePortEnabled.Should().BeFalse();
+        infoManager.Instance.SecurePort.Should().Be(0);
```
bart-vmware, line 184: Why only assert on the disabled port number here? Isn't the port also 0 in the other tests?

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -231,7 +243,7 @@ internal bool IsForceHostNameMethod()
 
     internal void SetPortsFromListenAddresses(IEnumerable<string> listenOnAddresses, string source, ILogger<EurekaInstanceOptions> logger)
     {
-        if (UseAspNetCoreUrls)
+        if (ShouldSetPortsFromListenAddresses)
```
bart-vmware: This condition should be in both callers (and removed here), so that `GetListenAddresses` doesn't execute.

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
```diff
@@ -20,7 +20,11 @@ public sealed partial class EurekaInstanceOptions
     internal const string DefaultStatusPageUrlPath = "/info";
     internal const string DefaultHealthCheckUrlPath = "/health";
 
-    private bool UseAspNetCoreUrls => !Platform.IsCloudFoundry || IsContainerToContainerMethod() || IsForceHostNameMethod();
+#pragma warning disable S1067 // Expressions should not be too complex
```
bart-vmware: Could you please clean this up so we don't need a suppression? It _is_ a complex expression now.
For example, define a separate property `HasAnyPortsEnabled` that evaluates the secure/non-secure enabled flags and port numbers.

bart-vmware (changes requested):

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
TimHess: Found a scenario that does appear to necessitate using a property, demonstrated in `Applies_dynamically_assigned_ports_when_kestrel_overrides_urls_config` - doesn't matter whether the kestrel config uses `0` or a specific number, that test fails if we don't use a property

bart-vmware: Does that mean it can't be a calculated property?
ie:
```c#
internal bool IsAnyPortConfigured => this is 
    { IsNonSecurePortEnabled: true, NonSecurePort: not null } or 
    { IsSecurePortEnabled: true, SecurePort: not null };
```

src/Discovery/test/Eureka.Test/DynamicPortAssignmentTest.cs
```diff
@@ -38,4 +40,145 @@ public async Task Applies_dynamically_assigned_ports_after_startup()
...
+        builder.Services.AddEurekaDiscoveryClient();
+
+        await using WebApplication app = builder.Build();
+        await app.StartAsync(TestContext.Current.CancellationToken);
+
+        var infoManager = app.Services.GetRequiredService<EurekaApplicationInfoManager>();
+
+        infoManager.Instance.IsNonSecurePortEnabled.Should().BeTrue();
+        infoManager.Instance.NonSecurePort.Should().NotBe(5000);
+        infoManager.Instance.NonSecurePort.Should().BePositive();
```
bart-vmware, line 71: An assertion on secure port (value + enabled) is missing here.

src/Discovery/src/Eureka/Configuration/EurekaInstanceOptions.cs
TimHess: Correct, the test I mentioned fails if we use the calculated property

bart-vmware (approved):
LGTM

PR #1667, opened 2026-03-31
Config Server stability improvements

### LINKED ISSUE #1664: ConfigServerClientOptions-related Enhancements
`IHostApplicationBuilder.AddConfigServer` should have an overload that takes a `ConfigServerClientOptions` instance.

Furthermore, the following code in Steeltoe:
```c#
if (!string.IsNullOrEmpty(hostEnvironment.EnvironmentName) &&
    hostEnvironment.EnvironmentName != "Production")
{
    // Only take IHostEnvironment.EnvironmentName when it was explicitly set (it defaults to "Production").
    // In the default case, we want the various other ways of setting the environment name to kick in.
    options.Environment = hostEnvironment.EnvironmentName;
}
```
is problematic. `ConfigServerClientOptions.EnvironmentName` should default to an empty string, and then `ConfigServerConfigurationProvider` should assume `Production` if still empty.

bart-vmware:
## Description

- Fix various multi-threading issues, such as torn/stale reads (by swapping snapshots) and various race conditions
- Reduced test flakiness by replacing `Sandbox` usage with `MemoryFileProvider` (without sleeps)
- Fix detection of local configuration changes and atomically update internal state
- Fix local configuration reloads to discard stale settings from earlier updates when keys are removed
- Fix injected `IOptionsMonitor<ConfigServerClientSettings>` to return the same settings as the provider uses internally
- Fix the preservation of the certificate issuer chain during configuration reloads
- Recreate `HttpClientHandler` for each request because reusing with changing certificates is not thread-safe
- Fix unbounded timer allocation when using HashCorp Vault token lease renewal
- Graceful shutdown, signaling in-flight timer callbacks to terminate instead of throwing unobserved exceptions
- Rerun service discovery during polled reloads (cached in Eureka)
- Fix race condition where multiple temporary service containers were built for service discovery
- Fix leaking change callback accumulation on repeated configuration reloads
- Fix threadpool starvation during retries
- Fix duplicate log messages and improve exceptions (messages and stack traces)
- Add overloads to post-configure Config Server options from code
- Postpone polling reloads to fix duplicate refresh on startup
- Fix configuration change detection when combined with placeholder/decryption providers
- Improved test coverage

Fixes #1664.

## Quality checklist


src/Configuration/src/ConfigServer/ConfigServerConfigurationSource.cs
```diff
@@ -5,82 +5,75 @@
...
-    /// Gets the default settings the Config Server client uses to contact the Config Server.
+    /// Gets the initial options the client uses to contact Config Server.
     /// </summary>
-    internal ConfigServerClientOptions DefaultOptions { get; }
+    public ConfigServerClientOptions DefaultOptions { get; }
 
     /// <summary>
-    /// Gets the configuration the Config Server client uses to contact the Config Server. Values returned override the default values provided in
-    /// <see cref="DefaultOptions" />.
+    /// Gets the configuration the client uses to contact Config Server. Entries overrule <see cref="DefaultOptions" />.
```
TimHess: Why change to `overrule`? AFAIK that's more of a legal term. Not necessarily a show-stopper, but I'm not convinced it's an improvement

src/Configuration/src/ConfigServer/ConfigServerConfigurationSource.cs
```diff
@@ -5,82 +5,75 @@
...
     private readonly ILoggerFactory _loggerFactory;
 
-    internal List<IConfigurationSource> Sources { get; } = [];
-    internal Dictionary<string, object> Properties { get; } = [];
+    internal List<IConfigurationSource> Sources { get; }
+    internal Dictionary<string, object> Properties { get; }
 
     /// <summary>
-    /// Gets the default settings the Config Server client uses to contact the Config Server.
+    /// Gets the initial options the client uses to contact Config Server.
```
TimHess: This should probably stick with "default" rather than "initial" since the options could be used throughout the entire lifecycle of the application if no additional configuration options are used. Also applies in the constructor

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationProviderTest.Loading.cs
```diff
@@ -172,68 +152,344 @@ await TestFailureTracer.CaptureAsync(async tracer =>
...
+        provider.Load();
+
+        await Task.Delay(2.Seconds(), TestContext.Current.CancellationToken);
+        handler.Mock.GetMatchCount(request).Should().Be(0);
+    }
+
+    [Theory]
+    [InlineData(false, "00:00:01")]
+    [InlineData(true, "00:00:00")]
+    public void OnSettingsChanged_stops_reload_timer_when_polling_becomes_ineffective(bool enabled, string pollingInterval)
```
TimHess:
```suggestion
    public void OnSettingsChanged_stops_reload_timer_when_polling_no_longer_enabled(bool enabled, string pollingInterval)
```
On first reading this test name, I thought maybe we were doing something like disabling reloads or adding a back-off when polling stops working. New name is a suggestion, there could be better options

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationProviderTest.Loading.cs
```diff
@@ -172,68 +152,344 @@ await TestFailureTracer.CaptureAsync(async tracer =>
...
+        var fileProvider = new MemoryFileProvider();
+
+        fileProvider.IncludeAppSettingsJsonFile("""
+            {
+              "spring": {
+                "cloud": {
+                  "config": {
+                    "name": "myName",
+                    "token": "MyVaultToken",
+                    "pollingInterval": "00:00:00"
```
TimHess: Is the inclusion of `pollingInterval` in this test intentional? I was confused for a minute, thinking it was `TokenRenewRate` since this test is for Vault, and that something was wrong here since a 0 value disables the config polling refresh.

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationProviderTest.Loading.cs
```diff
@@ -908,7 +1020,8 @@ private static ConfigServerClientOptions GetCommonOptions()
     {
         return new ConfigServerClientOptions
         {
-            Name = "myName"
+            Name = "myName",
+            Environment = "Staging"
```
TimHess, line 1032: This value does not appear to be used for anything, is this intentional?

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationBuilderExtensionsTest.cs
```diff
@@ -217,37 +199,94 @@ public void AddConfigServer_VCAP_SERVICES_Override_Defaults(string vcapServices)
...
         provider.ClientOptions.Uri.Should().NotBe("https://uri-from-settings");
         provider.ClientOptions.Uri.Should().Be("https://uri-from-vcap-services");
+        provider.ClientOptions.ClientId.Should().Be("some-client-id");
+        provider.ClientOptions.ClientSecret.Should().Be("some-secret");
+        provider.ClientOptions.AccessTokenUri.Should().Be("https://uaa-uri-from-vcap-services/oauth/token");
     }
 
     [Fact]
-    public void AddConfigServer_PaysAttentionToSettings()
+    public void AddConfigServer_CallbackOverridesConfigurationOverridesInitialOptions()
```
TimHess:
```suggestion
    public void AddConfigServer_CallbackOverridesConfigurationAndDefaultOptions()
```

src/Configuration/test/ConfigServer.Test/ConfigServerHealthContributorTest.cs
```diff
@@ -143,16 +105,25 @@ public async Task GetPropertySources_ReturnsExpected()
...
             Cached = new ConfigEnvironment()
         };
 
+        var optionsSnapshot = new ConfigServerClientOptions
+        {
+            Health =
+            {
+                TimeToLive = 1
+            }
+        };
```
TimHess: Do we need this? Looks like we could just use `contributor.Provider!.ClientOptions` below instead. The duplicated TTL value is what initially caught my eye.

src/Configuration/test/ConfigServer.Test/ConfigServerClientOptionsTest.cs
```diff
@@ -114,4 +115,337 @@ public async Task ConfigureConfigServerClientOptions_WithValues()
...
+        optionsMonitor.CurrentValue.FailFast.Should().BeTrue();
+
+        configuration["example-server-key"].Should().Be("example-server-value");
+
+        fileProvider.ReplaceAppSettingsJsonFile("""
+            {
+              "custom": {
+              "profileName": "example-profile"
+            },
+            "spring": {
```
TimHess:
```suggestion
              "custom": {
                "profileName": "example-profile"
              },
              "spring": {
```

src/Configuration/src/ConfigServer/ConfigServerHealthContributor.cs
```diff
@@ -74,45 +76,46 @@ internal void UpdateHealth(HealthCheckResult health, IList<PropertySource> sourc
...
-        if (IsCacheStale(currentTime))
+        if (IsCacheStale(currentTime, optionsSnapshot))
         {
             LastAccess = currentTime;
             LogCacheStale();
-            Cached = await provider.LoadInternalAsync(false, cancellationToken);
+
+            try
+            {
+                Cached = await provider.LoadInternalAsync(optionsSnapshot, false, cancellationToken);
```
TimHess, line 98: I see this isn't new for this PR, but I guess I didn't realize that calling the config server health contributor would actually trigger a fetch... Isn't that a little odd to have enabled by default?

TimHess (changes requested):

TimHess:
Most of the feedback is fairly superficial, these changes appear to be significant improvements and I don't see any major issues at this time

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationProviderTest.Loading.cs
bart-vmware: I wondered the same, then was falsely tricked by AI into believing this was needed to disable polling reload, so it wouldn't interfere with this test. But that's nonsense because it defaults to zero, so I'll remove it.

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationProviderTest.Loading.cs
bart-vmware: Yes, various tests fail without it, because mock expectations aren't met. We need a deterministic value that doesn't get auto-replaced later with "Production" when initially empty.

src/Configuration/src/ConfigServer/ConfigServerHealthContributor.cs
bart-vmware: It triggers a fetch, but then discards the returned configuration, so the data in `IConfiguration` is unaffected. See the second parameter, which is `false` here:
```c#
internal async Task<ConfigEnvironment?> LoadInternalAsync(ConfigServerClientOptions optionsSnapshot,
    bool updateDictionary, CancellationToken cancellationToken)
```
This looks fine to me.

What I find odd is that the result is cached. A health check is supposed to return the _current_ situation, regardless of whether the app crashes as a result. Using stale data could be considered just as undesirable as crashing. If the health endpoint returned the current state instead, the calling system could determine whether to recycle the app instance after multiple failures. But now we're lying. I didn't change it because I'm not sure why it works this way, so there may be a good reason I'm unaware of.

I also find it questionable that retries are respected during the health check if configured. But I left it as-is for the same reason.

TimHess: I'm fairly sure the reason it's cached is to avoid overloading the config server. With this health check we're making the server do the same amount of work as when we actually care about the contents of the response... Given that this contributor is for a config provider (that is queried at app startup and perhaps never again), I'd argue that it could be considered perfectly acceptable to report based on the last request to the server and never make additional requests (unless the last request was a failure).

Probably best to leave it as-is until/unless it becomes a problem - I believe what we have was based on Spring code

src/Configuration/test/ConfigServer.Test/ConfigServerConfigurationProviderTest.Loading.cs
TimHess: Ah, after removing it I see the failures... 21 places expecting it to have _some_ value, but none actually caring what the value is

TimHess (approved):
LGTM, thanks!

src/Configuration/src/ConfigServer/ConfigServerHealthContributor.cs
bart-vmware: > I'm fairly sure the reason it's cached is to avoid overloading the config server.

Health checks are supposed to hit the server each time they execute. If the server can't handle that, don't use health checks.

> With this health check we're making the server do the same amount of work as when we actually care about the contents of the response...

I agree that a full fetch is not needed. A simple ping that verifies connectivity should be sufficient.

> Given that this contributor is for a config provider (that is queried at app startup and perhaps never again), I'd argue that it could be considered perfectly acceptable to report based on the last request to the server and never make additional requests (unless the last request was a failure).

That is only applicable if polling is disabled. Then it basically acts as a readiness probe.

> Probably best to leave it as-is until/unless it becomes a problem - I believe what we have was based on Spring code

I agree.

PR #1672, opened 2026-04-14
Add IDiscoveryClient.InstancesFetched event

bart-vmware:
## Description

Adds `IDiscoveryClient.InstancesFetched` event that enables responding to changes in remote services and their instances.
Implemented for Eureka and configuration-based clients.

Reverts the change from #1666 that added taking `UseNetworkInterfaces` into account to determine Consul ports.

## Quality checklist

### DIFF
--- a/src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
+++ b/src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
@@ -1,0 +1,23 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+namespace Steeltoe.Common.Discovery;
+
+/// <summary>
+/// Provides data for the <see cref="IDiscoveryClient.InstancesFetched" /> event.
+/// </summary>
+public sealed class DiscoveryInstancesFetchedEventArgs : EventArgs
+{
+    /// <summary>
+    /// Gets the updated list of service instances, grouped by service ID.
+    /// </summary>
+    public IReadOnlyDictionary<string, IReadOnlyList<IServiceInstance>> InstancesByServiceId { get; }
+
+    public DiscoveryInstancesFetchedEventArgs(IReadOnlyDictionary<string, IReadOnlyList<IServiceInstance>> instancesByServiceId)
+    {
+        ArgumentNullException.ThrowIfNull(instancesByServiceId);
+
+        InstancesByServiceId = instancesByServiceId;
+    }
+}
--- a/src/Common/src/Common/Discovery/IDiscoveryClient.cs
+++ b/src/Common/src/Common/Discovery/IDiscoveryClient.cs
@@ -15,6 +15,11 @@ public interface IDiscoveryClient
     string Description { get; }
 
     /// <summary>
+    /// Occurs when service instances have been fetched from the discovery server.
+    /// </summary>
+    public event EventHandler<DiscoveryInstancesFetchedEventArgs> InstancesFetched;
+
+    /// <summary>
     /// Gets information used to register the local service instance (this app) to the discovery server.
     /// </summary>
     /// <returns>
--- a/src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
+++ b/src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the Apache 2.0 License.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections.ObjectModel;
 using Microsoft.Extensions.Options;
 using Steeltoe.Common.Discovery;
 
@@ -10,17 +11,69 @@ namespace Steeltoe.Discovery.Configuration;
 /// <summary>
 /// A discovery client that reads service instances from app configuration.
 /// </summary>
-public sealed class ConfigurationDiscoveryClient : IDiscoveryClient
+public sealed class ConfigurationDiscoveryClient : IDiscoveryClient, IDisposable
 {
     private readonly IOptionsMonitor<ConfigurationDiscoveryOptions> _optionsMonitor;
+    private readonly IDisposable? _changeTokenRegistration;
 
     public string Description => "A discovery client that returns service instances from app configuration.";
+
+    /// <summary>
+    /// Occurs when the configuration of service instances has been reloaded.
+    /// </summary>
+    public event EventHandler<DiscoveryInstancesFetchedEventArgs>? InstancesFetched;
 
     public ConfigurationDiscoveryClient(IOptionsMonitor<ConfigurationDiscoveryOptions> optionsMonitor)
     {
         ArgumentNullException.ThrowIfNull(optionsMonitor);
 
         _optionsMonitor = optionsMonitor;
+        _changeTokenRegistration = optionsMonitor.OnChange(OnOptionsChanged);
+    }
+
+    private void OnOptionsChanged(ConfigurationDiscoveryOptions options)
+    {
+        if (InstancesFetched != null)
+        {
+            ReadOnlyDictionary<string, IReadOnlyList<IServiceInstance>> instancesByServiceId = ToServiceInstanceMap(options.Services);
+            var eventArgs = new DiscoveryInstancesFetchedEventArgs(instancesByServiceId);
+            RaiseFetchEvent(eventArgs);
+        }
+    }
+
+    private static ReadOnlyDictionary<string, IReadOnlyList<IServiceInstance>> ToServiceInstanceMap(IList<ConfigurationServiceInstance> services)
+    {
+        // @formatter:wrap_chained_method_calls chop_always
+        // @formatter:wrap_before_first_method_call true
+
+        return services
+            .Where(service => service.ServiceId != null)
+            .GroupBy(service => service.ServiceId!, StringComparer.OrdinalIgnoreCase)
+            .ToDictionary(grouping => grouping.Key, grouping => (IReadOnlyList<IServiceInstance>)grouping
+                .Select(instance => (IServiceInstance)instance)
+                .ToList()
+                .AsReadOnly())
+            .AsReadOnly();
+
+        // @formatter:wrap_before_first_method_call restore
+        // @formatter:wrap_chained_method_calls restore
+    }
+
+    private void RaiseFetchEvent(DiscoveryInstancesFetchedEventArgs eventArgs)
+    {
+        // Execute on separate thread, so we won't block the configuration system in case the handler logic is expensive.
+        ThreadPool.QueueUserWorkItem(_ =>
+        {
+            try
+            {
+                InstancesFetched?.Invoke(this, eventArgs);
+            }
+            catch (Exception)
+            {
+                // Intentionally left empty. Adding a logger to the constructor is a breaking change.
+                // Adding an extra constructor confuses the service container.
+            }
+        });
     }
 
     /// <inheritdoc />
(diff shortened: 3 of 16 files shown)


src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
```diff
@@ -10,17 +11,69 @@ namespace Steeltoe.Discovery.Configuration;
...
+    /// Occurs when the configuration of service instances has been reloaded.
+    /// </summary>
+    public event EventHandler<DiscoveryInstancesFetchedEventArgs>? InstancesFetched;
+
     public ConfigurationDiscoveryClient(IOptionsMonitor<ConfigurationDiscoveryOptions> optionsMonitor)
     {
         ArgumentNullException.ThrowIfNull(optionsMonitor);
 
         _optionsMonitor = optionsMonitor;
+        _changeTokenRegistration = optionsMonitor.OnChange(OnOptionsChanged);
```
TimHess, line 31: Should `InstancesFetched` fire here in the constructor to align with the initial fetch behavior in Eureka?

If it is not intended to fire here, should it be documented that it may never fire, or is it unlikely a consumer would want an event from the initial load?

src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
```diff
@@ -0,0 +1,23 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the Apache 2.0 License.
+// See the LICENSE file in the project root for more information.
+
+namespace Steeltoe.Common.Discovery;
+
+/// <summary>
+/// Provides data for the <see cref="IDiscoveryClient.InstancesFetched" /> event.
+/// </summary>
+public sealed class DiscoveryInstancesFetchedEventArgs : EventArgs
```
TimHess, line 10: Is there enough context in general usage of this class that the name could be shortened to `InstancesFetchedEventArgs`?

src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Gets the updated list of service instances, grouped by service ID.
```
TimHess, line 13: It could be helpful for this mention something about case sensitivity.

It is also (slightly) possible for this list to be out of date with `GetInstancesAsync()`... should it be reframed more like `Gets the service instances from the notification` so it isn't mistaken as "guaranteed to be the most current list"

src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
```diff
@@ -10,17 +11,69 @@ namespace Steeltoe.Discovery.Configuration;
...
+        ThreadPool.QueueUserWorkItem(_ =>
+        {
+            try
+            {
+                InstancesFetched?.Invoke(this, eventArgs);
+            }
+            catch (Exception)
+            {
+                // Intentionally left empty. Adding a logger to the constructor is a breaking change.
+                // Adding an extra constructor confuses the service container.
```
TimHess, line 74: Would it be excessive to a callback on `ConfigurationDiscoveryOptions` so these can be handled somehow?

src/Common/src/Common/Discovery/IDiscoveryClient.cs
```diff
@@ -14,6 +14,11 @@ public interface IDiscoveryClient
     /// </summary>
     string Description { get; }
 
+    /// <summary>
+    /// Occurs when service instances have been fetched from the discovery server.
```
TimHess, line 18: a server may not be involved
```suggestion
    /// Occurs when service instances have been updated from the discovery backend.
```

TimHess: Could also mention something about how not all implementations raise this event and/or that inconsistently blocking or slow handlers could result in out-of-order event completion

src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
bart-vmware: > Should `InstancesFetched` fire here in the constructor to align with the initial fetch behavior in Eureka?

No, because it would be pointless. Subscribing to the event is only possible _after_ an instance has been created. If a client raises the event from its constructor, that's an unobservable implementation detail.

> If it is not intended to fire here, should it be documented that it may never fire, or is it unlikely a consumer would want an event from the initial load?

The typical consumption pattern is:
- Create an instance (or pull from the service container)
- Subscribe to the event
- Fetch service instances, storing the result for later comparison to detect relevant changes
- In the event handler, compare the stored result with the instances in the event to determine if something changed

I'm not sure what you mean by "should it be documented that it may never fire." It fires when the underlying instances in `IConfiguration` are reloaded.

src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
bart-vmware: This follows the typical pattern of:
```c#
public class Order
{
    public OrderStatus Status { get; set; }
}

public enum OrderStatus
{
    Pending,
    Devivered
}
```
I think the current name properly reflects its scope: instances in service discovery, rather than object instances or any other kind of instance.

src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
bart-vmware: > It could be helpful for this mention something about case sensitivity.

Case sensitivity is an implementation choice, not a requirement. Configuration-based and Eureka are case-insensitive. I don't think Consul is, though I'm not sure.

> It is also (slightly) possible for this list to be out of date with `GetInstancesAsync()`... should it be reframed more like `Gets the service instances from the notification` so it isn't mistaken as "guaranteed to be the most current list"

Not sure what you mean by being out of date or by the need to document it. Take a `MouseClickEvent`, for example. The event info contains the X and Y coordinates of the pointer when the click occurred. By the time the handler executes, the coordinates have likely changed. That's how events work.

The only reason service instances are provided in the event info is so the caller doesn't need to invoke `GetInstancesAsync` to retrieve them. The information is already available, and calling the method requires async/await, which is typically unavailable in an event handler.

src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
bart-vmware: What would the advantage of that be? Consumers aren't supposed to crash in an event handler. If they do, we prevent the configuration notification system from breaking as a result. If diagnostics are needed when that happens, the easiest would be for the consumer to catch unhandled exceptions in their event handler and log them.

Aside from that, a new `ConfigurationDiscoveryOptions` instance is created during each option reload. So any callback subscription from the previous options instance would be lost.

src/Common/src/Common/Discovery/IDiscoveryClient.cs
bart-vmware: I'm trying to blend in with the existing interface documentation, which consistently refers to fetching instances from a discovery server.

It's true that not all implementations may raise this event. The same applies to the existing `GetLocalServiceInstance`, which isn't always implemented. If the need comes up, we can add `Supports*` properties or even a `QueryCapabilities` method for dynamic feature discovery, but I didn't want to go that far now. We have no need for that at the moment.

How event handlers are dispatched is an implementation detail. If coordination between callback invocations is needed, consumers can use locking primitives. But that's a pretty esoteric case. Similar to handling callbacks from `System.Threading.Timer`: they're supposed to complete quickly or enqueue something. I wouldn't want to impose a dispatcher strategy requirement in the interface definition.

src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
bart-vmware: Regarding case sensitivity:
- The configuration-based client is case-insensitive because its data source is `IConfiguration`, whose keys are by definition case-insensitive.
- The Eureka client is case-insensitive for Spring compatibility. Registering an instance as "foo" ends up as "FOO" on the server. So a consumer that queries for "foo" is interpreted as case-insensitive.
- A consumer of a discovery client can make its own choice of whether casing matters. If the consumer uses the instances for load balancing, it should always ignore case because the host names to look up are case-insensitive. On the other hand, if the consumer just wants to print the instances, it probably wants to preserve the casing as returned from the server.

TimHess: Regarding case sensitivity: it is expected that all the implementations do the right thing for their case, and any form of statement here should be redundant. I am OK with leaving it out, but (as demonstrated by your last minute catch) it is something that would be easy to miss for anyone else implementing this.

Regarding stale data, I do have a test to demonstrate it, but it does depend on the handler being fairly wildly inconsistent in completion time (eg: the handling of the first event takes more time to complete than a second change event firing _and_ completing). We can move on without changing this.

src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
TimHess: Maybe I'm misunderstanding something, but I was under the impression that the initial fetch in Eureka would trigger the event, whereas there's never any event with the config provider unless there's a change, so the event handlers would need to be built a bit differently per-discovery-implementation

src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
TimHess: Fair point on new instance, and agree about event handlers shouldn't crash. Just thinking about it from the perspective of that desire to log being noted in the comment here, and us not providing anything. Considering it's error handling of user code, it's also fair to just expect the user to figure that out on their own

src/Common/src/Common/Discovery/IDiscoveryClient.cs
TimHess: I suppose we don't want to rewrite it all now, and the config provider was fundamentally intended as a testing/extreme-minimum provider. We can leave it

TimHess (approved):

src/Discovery/src/Configuration/ConfigurationDiscoveryClient.cs
bart-vmware: > Maybe I'm misunderstanding something, but I was under the impression that the initial fetch in Eureka would trigger the event

I haven't checked, but that's likely the case. However, it happens during construction. Unless users explicitly create a `EurekaDiscoveryClient` instance, there's no way to observe it:
```c#
var builder = WebApplication.CreateBuilder();
builder.AddEurekaDiscoveryClient(); // invokes event during ctor
var app = builder.Build();

var client = app.Services.GetRequiredService<EurekaDiscoveryClient>();
client.InstancesFetches += .... // first opportunity to listen
```

So I don't think it really matters that Eureka does that.

> whereas there's never any event with the config provider unless there's a change, so the event handlers would need to be built a bit differently per-discovery-implementation

From an interface perspective, the event is raised when instances are fetched. Whether that fetching occurs periodically (Eureka), or for some other reason (`IConfiguration` reload) is an implementation detail. Why would handlers need to be built differently for that?

TimHess: Ah, right. I should have gone back to the code instead of just theorizing. This looks fine

src/Common/src/Common/Discovery/DiscoveryInstancesFetchedEventArgs.cs
bart-vmware: Regarding stale data, I still don't understand the concern. The event args are a snapshot of the instances at that point in time. Did you spot a bug where that's not the case?

PR #1677, opened 2026-04-17
Follow-up changes for new `IDiscoveryClient.InstancesFetched` event

bart-vmware:
## Description

- Adjusted data in `EurekaDiscoveryClient.InstancesFetched` event to match what's returned from `GetInstancesAsync` (follow-up for #1672)
  - Keyed by VIP addresses instead of app names
  - Take `options.ReturnUpInstancesOnly` into account
  - Don't return an empty collection when no instances are found
- Remove duplicate URLs (case-insensitive) in load balancers. This could happen when both client-side and platform-based load balancing are active.

I considered using the new event in load balancers to evict entries in the distributed cache (which is disabled by default). The load balancer's caching exists because Consul doesn't cache on its own. Another reason is to prevent all app instances to simultaniously query Eureka at startup (the cache in `EurekaDiscoveryClient` is not distributed). So subscribing to the event is useless for Consul, because it never raises it. For Eureka, it doesn't help with app startup. Aside from that, the load balancer in a single app instance can't draw conclusions about when to evict cache keys, as it may be unaware of other app instances that added cache keys for remote services not used by the first app.

## Quality checklist

### DIFF
--- a/src/Discovery/src/Eureka/EurekaDiscoveryClient.cs
+++ b/src/Discovery/src/Eureka/EurekaDiscoveryClient.cs
@@ -373,20 +373,19 @@ public sealed partial class EurekaDiscoveryClient : IDiscoveryClient
 
     private static ReadOnlyDictionary<string, IReadOnlyList<IServiceInstance>> ToServiceInstanceMap(ApplicationInfoCollection apps)
     {
-        // @formatter:wrap_chained_method_calls chop_always
-        // @formatter:wrap_before_first_method_call true
+        var dictionary = new Dictionary<string, IReadOnlyList<IServiceInstance>>(StringComparer.OrdinalIgnoreCase);
 
-        return apps
-            .SelectMany(app => app.Instances)
-            .GroupBy(instance => instance.AppName, StringComparer.OrdinalIgnoreCase)
-            .ToDictionary(grouping => grouping.Key, grouping => (IReadOnlyList<IServiceInstance>)grouping
-                .Select(instance => instance.ToServiceInstance())
-                .ToList()
-                .AsReadOnly(), StringComparer.OrdinalIgnoreCase)
-            .AsReadOnly();
+        foreach (string vipAddress in apps.VipInstanceMap.Keys.ToArray())
+        {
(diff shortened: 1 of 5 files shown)


TimHess (approved):
The change generally LGTM, I do wonder about propagating the overloaded usage of service id and whether the interface should have some commentary around the specific field being implementation-dependent. But that could be opening a can of worms that's not easily solved

bart-vmware:
For future reference, here's what Gemini has to say about it:

> is app name in eureka the same as vip and secure vip address?

No, the app name (`appname`) in Eureka is not always the same as the VIP address (`vipAddress`) or Secure VIP address (`secureVipAddress`), although they are often configured to be identical in simple setups. 

Here is the breakdown of the differences based on Eureka's architecture:

1. App Name (`eureka.instance.appname`)
   Purpose: This is the logical name of the application (e.g., `user-service`, `order-service`).
   Usage: It is used by Eureka to group all instances of the same service together.
   Default: In Spring Cloud, this typically defaults to the `${spring.application.name}`. 

2. VIP Address (`eureka.instance.vipAddress`)
   Purpose: This is the "Virtual Internet Protocol" address. It is a logical identifier that clients use to discover and communicate with a group of service instances.
   Usage: It is intended for load balancing. Clients often use this address (e.g., `sampleservice.mydomain.net`) rather than a specific server instance.
   Default: By default, Spring Cloud sets this to be the same as the spring.application.name. 

3. Secure VIP Address (`eureka.instance.secureVipAddress`) 
   Purpose: This is the virtual address used for secure communication (HTTPS) with the service.
   Usage: If a service supports both HTTP and HTTPS, it might have both a `vipAddress` and a `secureVipAddress`. 

Key Differences in Configuration

When they are the same: In many setups, the `appname` and `vipAddress` are identical (e.g., both are `orderservice`) because it simplifies service discovery.
When they are different: The `vipAddress` is often used to define a more granular or complex routing address, such as `userservice-v1` or `userservice.prod.net`, while the `appname` remains simply userservice.
Function: `appname` is for registration identification, while `vipAddress` is for lookup and load-balancing.

PR #1680, opened 2026-04-21
Connectors: add option to turn off the built-in post-processors

### LINKED ISSUE #1154: Provide a clear extensibility model for connectors
Steeltoe Connectors have been completely rebuilt for the 4.0 release and are not currently designed for extensibility.

This issue will be used to track progress and feature requests for that effort.

Please feel free to comment about your use cases and expectations.

### LINKED ISSUE #1676: credentials:hostname should be credentials:host according to cloudfoundry specs
## Describe the bug
The cloudfoundry specification of the endpoint (and credentials) object should define a 'host' property.
In the code (of the postgresql Connection at least) the property name used is 'hostname'. 
This makes the Connection fail on Stackit.

The same is true for the 'port' property, that might also be defined as 'ports' and then as an array.

[Steeltoe line of code](https://github.com/SteeltoeOSS/Steeltoe/blob/d95d9bfeed00d270ccd9f217b9cb8a184fe847d4/src/Configuration/src/CloudFoundry/ServiceBindings/PostProcessors/PostgreSqlCloudFoundryPostProcessor.cs#L45)
[Specification](https://github.com/cloudfoundry/servicebroker/blob/master/spec.md#endpoint-object)

## Steps to reproduce
1. Deploy an app using postgresql on stackit
2. failure

## Expected behavior

The hostname of the database server should be correctly extracted from the following entry in the VCAP_SERVICES. 

```
 "stackit-postgres-flex": [
{
"label": "stackit-postgres-flex",
"provider": null,
"plan": "stackit-postgres-flex",
"name": "myapp",
"tags": [
"sql",
"database",
"object-relational",
"consistent",
"flex"
],
"instance_guid": "27d0af97-a4b0-46e3-20eb-bec246074b00",
"instance_name": "gridbox",
"binding_guid": "363b9a13-22b4-483a-62c6-9d328c2d24cb",
"binding_name": null,
"credentials":
{ "database": "stackit", 
"host": "c3ee356c-7939-4159-bc08-6915f3a0b6a5.postgresql.eu01.onstackit.cloud", 
"id": "16693", 
"password": "secret", 
"port": 5432, 
"roles": [ "login", "createdb" ], 
"uri": "postgresql://cfc2d0286c:wQ0LwOrPB7xTri41qdc0oIhxe7Xi0GVEjnA48xcCZsqvbpqfb3TFSdSo2nX3yrgF@c3ee356c-3939-4159-bc08-6915f3a0b6a5.postgresql.eu01.onstackit.cloud:5432/stackit", 
"username": "randomname" },
"syslog_drain_url": null,
"volume_mounts": []
}
]
}
```

 - Steeltoe Version 4.1.0
 - Platform: Cloud Foundry
 - OS: Linux
 - .NET Version:10
 - Any other library versions to note

### LINKED ISSUE #1678: Invalid credentials in VCAP_SERVICES crash Host Startup
## Describe the bug

With the VCAP_SERVICE from issue #1676 but the tag 'postgresql' added to the list of tags, the startup of the Host fails spectacularly.

## Steps to reproduce

1. Configure a service on you application that uses the work 'host' instead of 'hostname' in the credentials 

1.these are the calles to Steeltoe done before running the app: 
```  builder.Services.AddAllActuators();
             builder.AddCloudFoundryConfiguration();
            builder.Configuration.AddCloudFoundry();
            builder.AddPostgreSql();
 call app.Run();
```
the result:

```
15:57:35.293: [APP/PROC/WEB.0]       Application startup exception
15:57:35.293: [APP/PROC/WEB.0]       System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
15:57:35.293: [APP/PROC/WEB.0]        ---> System.Reflection.TargetInvocationException: Exception has been thrown by the target of an invocation.
15:57:35.293: [APP/PROC/WEB.0]        ---> System.ArgumentNullException: Value cannot be null. (Parameter 'Host')
15:57:35.293: [APP/PROC/WEB.0]          at System.ArgumentNullException.Throw(String paramName)
15:57:35.293: [APP/PROC/WEB.0]          at System.ArgumentNullException.ThrowIfNull(Object argument, String paramName)
15:57:35.293: [APP/PROC/WEB.0]          at System.ArgumentException.ThrowNullOrWhiteSpaceException(String argument, String paramName)
15:57:35.293: [APP/PROC/WEB.0]          at System.ArgumentException.ThrowIfNullOrWhiteSpace(String argument, String paramName)
15:57:35.293: [APP/PROC/WEB.0]          at Npgsql.NpgsqlConnectionStringBuilder.PostProcessAndValidate()
15:57:35.293: [APP/PROC/WEB.0]          at Npgsql.NpgsqlConnection.SetupDataSource()
15:57:35.293: [APP/PROC/WEB.0]          at Npgsql.NpgsqlConnection.set_ConnectionString(String value)
15:57:35.293: [APP/PROC/WEB.0]          at Npgsql.NpgsqlConnection..ctor(String connectionString)
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Constructor(Object obj, IntPtr* args)
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
15:57:35.293: [APP/PROC/WEB.0]          --- End of inner exception stack trace ---
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InvokeWithOneArg(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.RuntimeConstructorInfo.Invoke(BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
15:57:35.293: [APP/PROC/WEB.0]          at System.RuntimeType.CreateInstanceImpl(BindingFlags bindingAttr, Binder binder, Object[] args, CultureInfo culture)
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Common.DynamicTypeAccess.TypeAccessor.CreateInstance(Object[] arguments) in /_/src/Common/src/Common/DynamicTypeAccess/TypeAccessor.cs:line 27
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.PostgreSql.DynamicTypeAccess.NpgsqlConnectionShim.CreateInstance(PostgreSqlPackageResolver packageResolver, String connectionString) in /_/src/Connectors/src/Connectors/PostgreSql/DynamicTypeAccess/NpgsqlConnectionShim.cs:line 23
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.PostgreSql.PostgreSqlServiceCollectionExtensions.CreateConnection(IServiceProvider serviceProvider, String serviceBindingName, PostgreSqlPackageResolver packageResolver) in /_/src/Connectors/src/Connectors/PostgreSql/PostgreSqlServiceCollectionExtensions.cs:line 117
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.PostgreSql.PostgreSqlServiceCollectionExtensions.<>c__DisplayClass2_0.<AddPostgreSql>b__0(IServiceProvider serviceProvider, String serviceBindingName) in /_/src/Connectors/src/Connectors/PostgreSql/PostgreSqlServiceCollectionExtensions.cs:line 68
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.Connector`2.<>c__DisplayClass9_0.<.ctor>b__0() in /_/src/Connectors/src/Connectors/Connector.cs:line 75
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.Connector`2.CreateConnection() in /_/src/Connectors/src/Connectors/Connector.cs:line 133
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.Connector`2.GetConnection() in /_/src/Connectors/src/Connectors/Connector.cs:line 95
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InterpretedInvoke_Method(Object obj, IntPtr* args)
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
15:57:35.293: [APP/PROC/WEB.0]          --- End of inner exception stack trace ---
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.MethodBaseInvoker.InvokeWithNoArgs(Object obj, BindingFlags invokeAttr)
15:57:35.293: [APP/PROC/WEB.0]          at System.Reflection.RuntimeMethodInfo.Invoke(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Common.DynamicTypeAccess.ReflectionAccessor.InvokeMethod(String name, Boolean isPublic, Object instance, Object[] arguments) in /_/src/Common/src/Common/DynamicTypeAccess/ReflectionAccessor.cs:line 114
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Common.DynamicTypeAccess.InstanceAccessor.InvokeMethod(String name, Boolean isPublic, Object[] arguments) in /_/src/Common/src/Common/DynamicTypeAccess/InstanceAccessor.cs:line 63
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.DynamicTypeAccess.ConnectorShim`1.GetConnection() in /_/src/Connectors/src/Connectors/DynamicTypeAccess/ConnectorShim.cs:line 28
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.PostgreSql.PostgreSqlServiceCollectionExtensions.CreateHealthContributor(IServiceProvider serviceProvider, String serviceBindingName, PostgreSqlPackageResolver packageResolver) in /_/src/Connectors/src/Connectors/PostgreSql/PostgreSqlServiceCollectionExtensions.cs:line 95
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.PostgreSql.PostgreSqlServiceCollectionExtensions.<>c__DisplayClass2_0.<AddPostgreSql>b__1(IServiceProvider serviceProvider, String serviceBindingName) in /_/src/Connectors/src/Connectors/PostgreSql/PostgreSqlServiceCollectionExtensions.cs:line 69
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Connectors.ConnectorOptionsBinder.<>c__DisplayClass2_0.<RegisterHealthContributor>b__0(IServiceProvider serviceProvider) in /_/src/Connectors/src/Connectors/ConnectorOptionsBinder.cs:line 78
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitIEnumerable(IEnumerableCallSite enumerableCallSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitIEnumerable(IEnumerableCallSite enumerableCallSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitConstructor(ConstructorCallSite constructorCallSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceLookup.CallSiteRuntimeResolver.VisitRootCache(ServiceCallSite callSite, RuntimeResolverContext context)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceProvider.CreateServiceAccessor(ServiceIdentifier serviceIdentifier)
15:57:35.293: [APP/PROC/WEB.0]          at System.Collections.Concurrent.ConcurrentDictionary`2.GetOrAdd(TKey key, Func`2 valueFactory)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceProvider.GetService(ServiceIdentifier serviceIdentifier, ServiceProviderEngineScope serviceProviderEngineScope)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService(IServiceProvider provider, Type serviceType)
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.Extensions.DependencyInjection.ServiceProviderServiceExtensions.GetRequiredService[T](IServiceProvider provider)
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Management.Endpoint.ActuatorRouteBuilderExtensions.MapActuators(IEndpointRouteBuilder builder) in /_/src/Management/src/Endpoint/ActuatorRouteBuilderExtensions.cs:line 29
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Management.Endpoint.ApplicationBuilderExtensions.<>c__DisplayClass1_0.<UseActuatorEndpoints>b__0(IEndpointRouteBuilder endpoints) in /_/src/Management/src/Endpoint/ApplicationBuilderExtensions.cs:line 58
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.AspNetCore.Builder.EndpointRoutingApplicationBuilderExtensions.UseEndpoints(IApplicationBuilder builder, Action`1 configure)
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Management.Endpoint.ApplicationBuilderExtensions.UseActuatorEndpoints(IApplicationBuilder builder, Action`1 configureEndpoints) in /_/src/Management/src/Endpoint/ApplicationBuilderExtensions.cs:line 56
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Management.Endpoint.ApplicationBuilderExtensions.UseActuatorEndpoints(IApplicationBuilder builder) in /_/src/Management/src/Endpoint/ApplicationBuilderExtensions.cs:line 27
15:57:35.293: [APP/PROC/WEB.0]          at Steeltoe.Management.Endpoint.ConfigureActuatorsMiddlewareStartupFilter.<>c__DisplayClass0_0.<Configure>b__0(IApplicationBuilder app) in /_/src/Management/src/Endpoint/ConfigureActuatorsMiddlewareStartupFilter.cs:line 38
15:57:35.293: [APP/PROC/WEB.0]          at Microsoft.AspNetCore.Hosting.GenericWebHostService.StartAsync(CancellationToken cancellationToken)
15:57:35.294: [APP/PROC/WEB.0] fail: Microsoft.Extensions.Hosting.Internal.Host[11]
15:57:35.294: [APP/PROC/WEB.0]       Hosting failed to start

```

## Expected behavior
A shorter more descriptive error message about an invalid connectionstring.

 - Steeltoe Version 4.1.0
 - Platform: Cloud Foundry (stackit)
 - OS: linux
 - .NET Version 10
 - Any other library versions to note

bart-vmware:
## Description

Steeltoe’s Cloud Foundry service-binding pipeline is more controllable and less intrusive for real-world platforms.

- **Targeted built-in binding support** — When you configure a connector (for example, PostgreSQL or Redis), Steeltoe only applies the Cloud Foundry post-processing that matches that kind of service instead of always running every built-in broker mapper. That reduces surprise transformations when you use multiple services or brokers whose credential shapes do not match the built-in brokers' assumptions.
- **Opt out of default binding translation** — You can turn off Steeltoe’s automatic Cloud Foundry and Kubernetes service-binding wiring for a connector and supply your own mapping from platform credentials (including non-standard or third-party brokers) into the configuration shape Steeltoe connectors already understand. Inline API documentation includes a concrete mapping example.
- **Manual configuration** — Hosts and libraries that build configuration themselves can register Cloud Foundry bindings with an explicit set of broker types so only the desired built-in mappers run.
- **Reference updates** — Comments that point to vendor documentation for binding credential fields use current documentation URLs (no runtime behavior change).

Default behavior for typical Steeltoe connector usage remains automatic based on the environment; the changes mainly help when built-in mappings conflict with your broker or when you need full control over how credentials become connection settings.

Fixes #1154, fixes #1678.
Closes #1676 (we'll reconsider if it comes up again).


src/Connectors/src/Connectors/ConnectorConfigureOptionsBuilder.cs
```diff
@@ -13,4 +13,83 @@ public sealed class ConnectorConfigureOptionsBuilder
...
+    /// <c>true</c> when using custom logic to convert platform-based credentials to driver-specific configuration keys.
+    /// <para>
+    /// <example>
+    /// For example, to use a third-party Cloud Foundry service broker, the following code can be used to map the PostgreSQL credentials to the format that
+    /// <see href="https://www.npgsql.org/doc/api/Npgsql.NpgsqlConnectionStringBuilder.html">
+    /// NpgsqlConnectionStringBuilder
+    /// </see>
+    /// expects:
+    /// <code><![CDATA[
+    /// Environment.SetEnvironmentVariable("VCAP_SERVICES", """
```
TimHess: Can we change this around so it's more like "When a VCAP_SERVICES entry follows this format..." instead of `Environment.SetEnvironmentVariable`?

src/Connectors/src/Connectors/ConnectorConfigurer.cs
```diff
@@ -20,8 +20,11 @@ public static void Configure<TPostProcessor>(IConfigurationBuilder builder, Acti
             var optionsBuilder = new ConnectorConfigureOptionsBuilder();
             configureAction?.Invoke(optionsBuilder);
 
-            builder.AddCloudFoundryServiceBindings();
-            builder.AddKubernetesServiceBindings();
+            if (!optionsBuilder.SkipDefaultServiceBindings)
```
TimHess, line 27: In addition to this hammer that turns off _all_ of the built-in bindings support, it could be nice to selectively disable what support is included. I think [ignoreKeyPredicate](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Configuration/src/CloudFoundry/ServiceBindings/ConfigurationBuilderExtensions.cs#L70) might be relevant here, but I'm not positive. Would making that more accessible or easier to use get us there?

bart-vmware: I looked into that, but then the key predicate needs to know about the internals of the specific post-processors. For example, the predicate needs to filter keys such as `vcap:services:csb-azure-postgresql:0:binding_guid`.

If I understand correctly, the issue is that mixing built-in and third-party connectors breaks it. For example:
```c#
builder.AddPostgreSql(configure => configure.SkipDefaultServiceBindings = true, null);
builder.AddRabbitMQ(); // built-in RabbitMQ broker
```

The `AddRabbitMQ` call adds all built-in post-processors, which makes the former `SkipDefaultServiceBindings` effectively useless.

The underlying problem is that multiple connectors share the same set of post-processors. I'm looking into refactoring so that multiple `CloudFoundryServiceBindingConfigurationSource` entries can coexist in configuration.

bart-vmware: Pushed commits to address this. In case it helps, here's a description of what's changed:

---

### Cloud Foundry service bindings (`Steeltoe.Configuration.CloudFoundry`)

**Singleton source removed.** `AddCloudFoundryServiceBindings` no longer guards on `!builder.EnumerateSources<CloudFoundryServiceBindingConfigurationSource>().Any()`. Every qualifying call appends another `CloudFoundryServiceBindingConfigurationSource` to `IConfigurationBuilder.Sources`, so the composite configuration can contain **multiple** CF binding providers.

**Post-processor registration is partitioned.** A new public `[Flags]` enum `CloudFoundryServiceBrokerTypes` identifies which built-in CF post-processors apply to **that** source instance. `RegisterPostProcessors` registers only processors whose flag is set (Eureka, Identity, MongoDB, MySQL, PostgreSQL, RabbitMQ, Redis, SQL Server). The previous behavior (all processors on one source) maps to `CloudFoundryServiceBrokerTypes.All`.

**API surface.** New overloads wire `brokerTypes` through; the lowest-level overload accepts `IServiceBindingsReader?` (null → `EnvironmentServiceBindingsReader`) and skips adding a source when `brokerTypes == None`. `ArgumentNullException` for `serviceBindingsReader` was dropped in favor of null-coalescing to the default reader.

**Source type.** `CloudFoundryServiceBindingConfigurationSource`’s constructor now takes `(IServiceBindingsReader, CloudFoundryServiceBrokerTypes)`. The broker set is stored on the source (currently used for `[DebuggerDisplay]`).

**Tests.** `ConfigurationBuilderExtensionsTest` gains coverage that `PostgreSql | MySql` yields exactly two registered post-processors on the single added source for that call.

---

### Connectors (`Steeltoe.Connectors`)

**`ConnectorConfigurer.Configure<TPostProcessor>`** now requires a non-null `Action<ConnectorConfigureOptionsBuilder>`, plus `IServiceBindingsReader?` and `ILoggerFactory`. It calls `AddCloudFoundryServiceBindings(DefaultIgnoreKeyPredicate, serviceBindingsReader, optionsBuilder.CloudFoundryBrokerTypes, loggerFactory)` instead of the parameterless CF registration, still honoring `SkipDefaultServiceBindings` (when true, neither CF nor Kubernetes default binding registration runs from this path—unchanged from prior commit’s semantics).

**`ConnectorConfigureOptionsBuilder`** gains an **internal** `CloudFoundryBrokerTypes` property (`CloudFoundryServiceBrokerTypes`) populated by each connector’s `Configure*` wrapper.
**Per-connector `Configure*` / `Add*` extensions** wrap the user delegate to set `CloudFoundryBrokerTypes` to:
- the connector’s **single** broker flag when defaults are on (e.g. PostgreSQL → `PostgreSql`), or  
- `None` when `SkipDefaultServiceBindings` is true, or  
- **`None` unconditionally for Cosmos DB** (so that connector’s registration does not add a CF binding source with Steeltoe’s relational/document-oriented CF post-processors).

**Testability.** Internal overloads were added on several connectors (e.g. `ConfigurePostgreSql(..., IServiceBindingsReader?)`, `AddPostgreSql(..., IServiceBindingsReader?)`) so tests can supply `StringServiceBindingsReader` **without** manually calling `AddCloudFoundryServiceBindings` on the builder first—important now that each connector may register its own CF source.

**Kubernetes bindings** are unchanged in this commit: `AddKubernetesServiceBindings` still deduplicates on `KubernetesServiceBindingConfigurationSource` (single K8s source per builder).

---

### Test / assertion fallout
Any test that assumed **exactly one** `CloudFoundryServiceBindingConfigurationProvider` in the built `IConfiguration` was relaxed (e.g. `ContainSingle` → `NotBeEmpty`) across Bootstrap, Discovery, Security, Connectors, and EF extension tests.

**`CustomBrokerTests.cs`** was added and the large “third-party service bindings” PostgreSQL scenario was **moved out of** `PostgreSqlConnectorTest` into that file; `PostgreSqlConnectorTest` was updated to use the new `AddPostgreSql(..., serviceBindingsReader)` overload for multi-binding scenarios.

---

### Reviewer takeaways
1. **Functional change:** CF VCAP processing can run through **N** configuration sources/providers, each with a **subset** of post-processors, instead of one global source with all processors.  
2. **Tradeoff:** Same `VCAP_SERVICES` payload may be read and post-processed multiple times (once per added CF source), in exchange for isolating which processors run per connector registration.  
3. **Public API:** New enum and `AddCloudFoundryServiceBindings` overloads are recorded in `PublicAPI.Unshipped.txt` for the Cloud Foundry configuration assembly.

bart-vmware: Also updated the PR description.

TimHess: "Kubernetes bindings are unchanged in this commit" -- This whole situation is duplicated there, isn't it? We should either add an issue to track it (which is probably fine since I don't know of any use case for the feature outside of TAP, which may not be running any .NET apps anymore), or just go ahead and apply the same fix there too (if it isn't significant work) so that it's all in sync.

src/Connectors/src/Connectors/MySql/MySqlHostApplicationBuilderExtensions.cs
```diff
@@ -43,16 +44,16 @@ public static IHostApplicationBuilder AddMySql(this IHostApplicationBuilder buil
     public static IHostApplicationBuilder AddMySql(this IHostApplicationBuilder builder, Action<ConnectorConfigureOptionsBuilder>? configureAction,
         Action<ConnectorAddOptionsBuilder>? addAction)
     {
-        return AddMySql(builder, MySqlPackageResolver.Default, configureAction, addAction);
+        return AddMySql(builder, MySqlPackageResolver.Default, configureAction, addAction, null);
     }
 
     internal static IHostApplicationBuilder AddMySql(this IHostApplicationBuilder builder, MySqlPackageResolver packageResolver,
-        Action<ConnectorConfigureOptionsBuilder>? configureAction = null, Action<ConnectorAddOptionsBuilder>? addAction = null)
+        Action<ConnectorConfigureOptionsBuilder>? configureAction, Action<ConnectorAddOptionsBuilder>? addAction, IServiceBindingsReader? serviceBindingsReader)
```
TimHess, line 51: The change to make these actions non-optional has created a bit of churn in the tests, couldn't we just make `serviceBindingsReader` also optional instead?

src/Configuration/src/CloudFoundry/ServiceBindings/ConfigurationBuilderExtensions.cs
```diff
@@ -69,15 +87,41 @@ public static IConfigurationBuilder AddCloudFoundryServiceBindings(this IConfigu
...
+    public static IConfigurationBuilder AddCloudFoundryServiceBindings(this IConfigurationBuilder builder, Predicate<string> ignoreKeyPredicate,
+        IServiceBindingsReader? serviceBindingsReader, CloudFoundryServiceBrokerTypes brokerTypes, ILoggerFactory loggerFactory)
     {
         ArgumentNullException.ThrowIfNull(builder);
         ArgumentNullException.ThrowIfNull(ignoreKeyPredicate);
-        ArgumentNullException.ThrowIfNull(serviceBindingsReader);
         ArgumentNullException.ThrowIfNull(loggerFactory);
 
-        if (!builder.EnumerateSources<CloudFoundryServiceBindingConfigurationSource>().Any())
+        if (brokerTypes != CloudFoundryServiceBrokerTypes.None)
```
TimHess: Can we eliminate the multi-post-processing problem by changing the `CloudFoundryServiceBrokerTypes` property on `CloudFoundryServiceBindingConfigurationSource` to `internal` and checking it on all instances before adding a new source?

```csharp
// Figure out which broker types haven't been registered yet
CloudFoundryServiceBrokerTypes missingBrokerTypes = brokerTypes;
foreach (var existingSource in builder.Sources.OfType<CloudFoundryServiceBindingConfigurationSource>())
{
    missingBrokerTypes &= ~existingSource.BrokerTypes;
}

// Only add a new source if there are actually missing post-processors
if (missingBrokerTypes != CloudFoundryServiceBrokerTypes.None)
{
    var source = new CloudFoundryServiceBindingConfigurationSource(serviceBindingsReader ?? DefaultReader, missingBrokerTypes)
    // ...
```

src/Connectors/src/Connectors/ConnectorConfigurer.cs
```diff
@@ -3,25 +3,32 @@
...
 namespace Steeltoe.Connectors;
 
 internal static class ConnectorConfigurer
 {
-    public static void Configure<TPostProcessor>(IConfigurationBuilder builder, Action<ConnectorConfigureOptionsBuilder>? configureAction,
-        TPostProcessor connectionStringPostProcessor)
+    private static readonly Predicate<string> DefaultIgnoreKeyPredicate = _ => false;
+
+    public static void Configure<TPostProcessor>(IConfigurationBuilder builder, Action<ConnectorConfigureOptionsBuilder> configureAction,
+        TPostProcessor connectionStringPostProcessor, IServiceBindingsReader? serviceBindingsReader, ILoggerFactory loggerFactory)
```
TimHess, line 19: Can/should we make the `LoggerFactory` param optional here with a default value of `NullLoggerFactory.Instance` so we don't have to repeat that at all of the non-test call sites? Or even remove the param entirely and always pass in `NullLoggerFactory.Instance` since that's what's happening now

src/Connectors/src/Connectors/MongoDb/MongoDbConnectionStringBuilder.cs
```diff
@@ -37,22 +37,16 @@ public object? this[string keyword]
                 return ConnectionString;
             }
 
-            // Allow getting unknown keyword, if it was set earlier. We don't pretend to know all valid query string parameters.
-            if (_settings.TryGetValue(keyword, out string? value))
-            {
-                return value;
-            }
-
-            AssertIsKnownKeyword(keyword);
```
TimHess, line 46: This behavior change should probably be in the PR description

TimHess (changes requested):
In addition to the other comments, I'm not sure this PR should fix #1154 unless we have a test or documentation that shows how to add your own connector (maybe Oracle would be a good example since we had it in the past?)

src/Connectors/src/Connectors/MySql/MySqlHostApplicationBuilderExtensions.cs
bart-vmware: I did so initially, but got warnings from Sonar that the overload wouldn't be callable unless named parameters are used. This is the simplest solution, and it's internal-only.

src/Configuration/src/CloudFoundry/ServiceBindings/ConfigurationBuilderExtensions.cs
bart-vmware: I don't see what problem that solves. Whether a single provider enumerates multiple times (postgres + mysql), or multiple providers (postgres + mysel) enumerate once, doesn't make a difference, right?

src/Connectors/src/Connectors/ConnectorConfigurer.cs
bart-vmware: It's all internal code, so it doesn't matter much. This isn't even called from tests. The refactoring makes it a bit easier to see what's going on.

src/Connectors/src/Connectors/MongoDb/MongoDbConnectionStringBuilder.cs
bart-vmware: It's a logical consequence of opening up for external usage. I initially implemented it very strictly because there was no need to read anything except the properties we knew about. The change aligns with the broader intent to open up for extensibility, which is what the PR description already mentions. I'm not sure how it fits in there, as the description is targeted at usage, not so much about internal implementation details.

src/Connectors/src/Connectors/ConnectorConfigurer.cs
bart-vmware: Kubernetes bindings have not changed. The changes affect Cloud Foundry only. As you said, TAP has been discontinued, so I don't see the point of opening an issue to add more flexibility to an abandoned feature. It is significant work to implement similar changes for TAP, if we can even test it. I'd rather ship working code than focus on TAP.

bart-vmware:
> In addition to the other comments, I'm not sure this PR should fix #1154 unless we have a test or documentation that shows how to add your own connector (maybe Oracle would be a good example since we had it in the past?)

We've discussed this offline. The connectors in 3x were very flexible but also very hard to understand, use or extend. Connectors in 4x are much simpler and no longer "bring your own technology". They are constrained to a small set of data sources: PostgreSQL, MySQL, SQL Server, MongoDB, CosmosDB, RabbitMQ and Redis/Valkey.

With the changes in this PR, it's easy to:
1. Parse `VCAP_SERVICES` into a configuration structure of choice
2. Obtain the connection string from configuration and feed that to a downstream driver, depending on its API
3. Implement `IHealthContributor` to add health checks

The "add-Oracle" scenario isn't a goal right now, so I'd like to close the "extend-anything" issue until we see multiple similar requests and have a specific plan and clear understanding of what needs to be opened up. Until then, the implementation is better left internal (I wouldn't have been able to implement this PR if everything were public already).

src/Connectors/src/Connectors/ConnectorConfigurer.cs
TimHess: No, I didn't mean to E2E test it just a low-effort duplication of the code & test changes. The spec is open and supposed to be for Kubernetes generically, but I'm not sure what else outside of TAP actually uses it. I suppose we can leave it until/unless somebody actually runs into it and complains.

src/Configuration/src/CloudFoundry/ServiceBindings/ConfigurationBuilderExtensions.cs
TimHess: This test will demo the problem, but I don't think there's any other way to hit it besides `AddCloudFoundryServiceBindings()` + `AddSomeOtherThing()`, so it's probably not that significant. 

```csharp
[Fact]
public void Adding_Connector_After_Global_CloudFoundry_Adds_Duplicate_Configuration_Source()
{
    var builder = new ConfigurationBuilder();

    // 1. User explicitly adds all CF service bindings (registers CloudFoundryServiceBrokerTypes.All)
    builder.AddCloudFoundryServiceBindings();
    // 2. User adds a specific connector.
    builder.ConfigureRabbitMQ();

    // Under the current PR's code, there are now two CloudFoundry configuration sources in the builder.
    var cfSources = builder.Sources.OfType<CloudFoundryServiceBindingConfigurationSource>().ToList();
    
    // This assertion passes
    cfSources.Should().HaveCount(2);
    // Consequence: When builder.Build() runs, VCAP_SERVICES is read from the environment and 
    // parsed from JSON into memory twice. Furthermore, RabbitMQCloudFoundryPostProcessor executes twice.
    //
    // If you use the bitwise flags to check existing sources before adding a new one:
    // missingBrokerTypes &= ~existingSource.BrokerTypes;
    // ...this count drops back down to 1, completely eliminating the performance tradeoff.
}
```

Unless this triggers some other concern for you, I'm fine with you resolving this conversation.

TimHess (dismissed):

src/Configuration/src/CloudFoundry/ServiceBindings/ConfigurationBuilderExtensions.cs
bart-vmware: I understand now. That's a nice optimization, pushed.

TimHess (approved):
LGTM

PR #1688, opened 2026-05-01
Bump 3x vulnerable dependencies; drop net60

bart-vmware:
## Summary

This PR for Steeltoe 3.x:
- Addresses vulnerable dependency chains, notably OpenTelemetry and KubernetesClient.
- Drops support for .NET 6 and adds .NET 8 support for Integration/Messaging/Stream.
- Consolidates Microsoft/System package versions on the 8.x line to reduce restore conflicts and improve the security posture of transitive dependencies.
- Libraries that previously targeted `netstandard2.0` only now multi-target `net8.0`, providing the richer APIs of dependencies to consumers.

### Target frameworks

Applications on .NET 6 should move to .NET 8 or remain on an earlier Steeltoe release.

### Kubernetes packages

`Steeltoe.Common.Kubernetes`, `Steeltoe.Extensions.Configuration.KubernetesBase`, and `Steeltoe.Discovery.Kubernetes` previously shipped for `netstandard2.0` and `netstandard2.1`. They now target `net8.0` only.

Rationale: No non-vulnerable version of `KubernetesClient` exists that is compatible with netstandard.

Effect: Apps that rely on netstandard-only Kubernetes packages from .NET Framework, older .NET Core, or similar need to run on .NET 8, or continue using an older Steeltoe release.

### Messaging, integration, and stream

`Steeltoe.Integration`, `Steeltoe.Messaging`, and `Steeltoe.Stream` are now supported on .NET 8.

### OpenTelemetry and tracing

Management tracing packages move to the OpenTelemetry 1.15.x line, with matching instrumentation package versions where applicable.

Rationale: remain on supported, patched OpenTelemetry releases.

### Supply chain and NuGet audit

Where audits flagged issues, selected transitive packages are pinned to patched versions (for example `System.Net.Http`, `System.Text.RegularExpressions`, and `System.Security.Cryptography.Xml` in the affected projects).

---

## Suggested upgrade checklist

1. Retarget applications from `net6.0` to `net8.0`, or stay on a prior Steeltoe release.
2. If Kubernetes configuration or discovery relied on netstandard-only packages on older runtimes, plan a move to .NET 8 for those workloads.
3. Align direct OpenTelemetry references with the 1.15.x line if your application manages those packages explicitly.
4. Regression-test Kubernetes configuration, secrets, and discovery after the `KubernetesClient` major version upgrade.
5. Run `dotnet list package --vulnerable` or the organization’s equivalent after upgrading Steeltoe.

---

## Quality checklist


.github/workflows/package.yml
```diff
@@ -31,17 +31,19 @@ jobs:
 
     steps:
     - name: Setup .NET
-      uses: actions/setup-dotnet@v4
+      uses: actions/setup-dotnet@v5
       with:
         dotnet-version: |
-          6.0.*
           8.0.*
+          10.0.*
```
TimHess, line 38: Why do we need .net 10 in these workflows?

versions.props
```diff
@@ -12,17 +12,18 @@
     <JwtTokensVersion>7.7.1</JwtTokensVersion>
     <SymReaderVersion>1.2.0</SymReaderVersion>
     <SymReaderPortableVersion>1.4.0</SymReaderPortableVersion>
-    <SystemDiagnosticsVersion>8.0.0</SystemDiagnosticsVersion>
-    <SystemJsonVersion>8.0.5</SystemJsonVersion>
-    <SystemNetHttpJsonVersion>3.2.1</SystemNetHttpJsonVersion>
-    <SystemReflectionVersion>4.6.0</SystemReflectionVersion>
+    <SystemVersion>8.0.0</SystemVersion>
+    <SystemJsonVersion>
+      <!-- Earliest non-vulnerable version in 8.x release line -->
```
TimHess, line 17: I don't think we should add `<!-- Earliest non-vulnerable version in 8.x release line -->` on any of these, especially when they weren't changed from the prior release

versions.props
```diff
@@ -12,17 +12,18 @@
     <JwtTokensVersion>7.7.1</JwtTokensVersion>
```
TimHess, line 12: `VersionPrefix` needs to bump to `3.4.0`

build/templates/component-build.yaml
```diff
@@ -52,13 +52,6 @@ jobs:
       projects: $(SolutionFile)
       arguments: -f net8.0 ${{parameters.skipFilter}} $(CommonTestArgs)
```
TimHess, line 53: Should we add a suppression for `Check failure on line 53 in build/templates/component-build.yaml`, or maybe use an enum for the input (like `testFilterPreset: none | integration | skipOnLinux | skipOnMacOS`) and expand that out into the skip filters during the run?

versions.props
bart-vmware: I did so to indicate these are intentionally different from `SystemVersion`, which I introduced. The comment is meant to warn against ever unifying them (which I initially did). Do you still believe they should be removed?

versions.props
TimHess: We can do this in a separate PR

.github/workflows/package.yml
bart-vmware: I'm not sure. We're already running on NET10 SDK because it's included in the runners anyway. I don't know if our workflows rely on SDK features that require v10 or higher, such as the vulnerability checks. So I thought to include them explicitly, but I can remove them to speed up the builds a little. Let me know what you prefer.

versions.props
bart-vmware: Let's handle that separately, so it's easier to track when versions bump.

versions.props
TimHess: Given the life expectancy on the 3.x, I don't foresee that as a likely condition... If you want to keep them that's fine, I just don't see the value

build/templates/component-build.yaml
bart-vmware: Could it be a stale comment? I've already reviewed the security hotspots in SonarCloud, where this was reported as well.

versions.props
bart-vmware: Ok, let's keep them then.

build/templates/component-build.yaml
TimHess: Oh, now I see that you marked it False Positive in Sonar now... they could really make that clearer in their interface. 

I agree that it's not a _real_ issue since it's not true user-controlled data. It's annoying that it still shows in the github UI, but we can just move on

.github/workflows/package.yml
TimHess: It's probably only seconds of build time, not worth the effort - but what do you make of some of these checks being duplicated?
image

bart-vmware: I'm not sure what you mean. What duplication?

TimHess: It looks like it's running the checks `Scan vulnerable dependencies` and `Package` twice - there are unique runs under each of those that look like they ran at the same time

bart-vmware: Oh, now I see it. For Package, one run happened 40 minutes later. Perhaps because I did a force-push a bit later? It could be a GHA glitch. I saw a banner saying there were issues and that they were reindexing. Let's keep an eye on it for subsequent PRs, but I don't think there's much we can do now.

TimHess (approved):

PR #1691, opened 2026-05-08
Address flaky tests

TimHess:
## Description

One flaky test is fixed with a small code change to prevent one last SBA refresh request after refresh has been disabled.

The rest are pure test changes:
- Address race condition in thread dump test
- Catch network-related SQLServer connection error
- Remove retry from eureka server timeout
- More buffer for infra around health aggregator parallelization test (allow +1s for +3s work)

## Quality checklist

### DIFF
--- a/src/Connectors/test/Connectors.Test/RelationalDatabaseHealthContributorTest.cs
+++ b/src/Connectors/test/Connectors.Test/RelationalDatabaseHealthContributorTest.cs
@@ -111,7 +111,10 @@ public sealed class RelationalDatabaseHealthContributorTest
         result.Description.Should().Be("SQL Server health check failed");
         result.Details.Should().Contain("host", "localhost");
         result.Details.Should().Contain("service", "Example");
-        result.Details.Should().ContainKey("error").WhoseValue.As<string>().Should().StartWith("SqlException: Connection Timeout Expired.");
+
+        result.Details.Should().ContainKey("error").WhoseValue.As<string>().Should().Match(exception =>
+            exception.StartsWith("SqlException: Connection Timeout Expired.", StringComparison.InvariantCulture) ||
+            exception.StartsWith("SqlException: A network-related or instance-specific error", StringComparison.InvariantCulture));
     }
 
     [Fact(Skip = "Integration test - Requires local SQL Server instance")]
--- a/src/Connectors/test/Connectors.Test/RelationalDatabaseHealthContributorTest.cs
+++ b/src/Connectors/test/Connectors.Test/RelationalDatabaseHealthContributorTest.cs
@@ -113,8 +113,8 @@ public sealed class RelationalDatabaseHealthContributorTest
         result.Details.Should().Contain("service", "Example");
 
         result.Details.Should().ContainKey("error").WhoseValue.As<string>().Should().Match(exception =>
-            exception.StartsWith("SqlException: Connection Timeout Expired.", StringComparison.InvariantCulture) ||
-            exception.StartsWith("SqlException: A network-related or instance-specific error", StringComparison.InvariantCulture));
+            exception.StartsWith("SqlException: Connection Timeout Expired.", StringComparison.Ordinal) ||
+            exception.StartsWith("SqlException: A network-related or instance-specific error", StringComparison.Ordinal));
     }
 
     [Fact(Skip = "Integration test - Requires local SQL Server instance")]
--- a/src/Discovery/test/Eureka.Test/EurekaServiceCollectionExtensionsTest.cs
+++ b/src/Discovery/test/Eureka.Test/EurekaServiceCollectionExtensionsTest.cs
@@ -62,7 +62,7 @@ public sealed class EurekaServiceCollectionExtensionsTest
         var appSettings = new Dictionary<string, string?>
         {
             ["Eureka:Client:EurekaServer:ConnectTimeoutSeconds"] = "1",
-            ["Eureka:Client:EurekaServer:RetryCount"] = "1"
+            ["Eureka:Client:EurekaServer:RetryCount"] = "0"
         };
 
         IConfiguration configuration = new ConfigurationBuilder().AddInMemoryCollection(appSettings).Build();
--- a/src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
+++ b/src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
@@ -44,15 +44,20 @@ internal sealed partial class SpringBootAdminPeriodicRefresh : IAsyncDisposable
 
             do
             {
-                LogStartingRefreshCycle();
+                // A tick queued just before periodic refresh was disabled would still be delivered here.
+                // Checking the period prevents executing a stale tick when refresh has been turned off.
+                if (isFirstTime || _periodicTimer.Period != Timeout.InfiniteTimeSpan)
+                {
+                    LogStartingRefreshCycle();
 
-                try
-                {
-                    await _runner.RunAsync(isFirstTime, _timerTokenSource.Token);
-                }
-                catch (Exception exception) when (!exception.IsCancellation())
-                {
-                    LogRefreshCycleFailed(exception);
+                    try
+                    {
+                        await _runner.RunAsync(isFirstTime, _timerTokenSource.Token);
+                    }
+                    catch (Exception exception) when (!exception.IsCancellation())
+                    {
+                        LogRefreshCycleFailed(exception);
+                    }
                 }
 
                 isFirstTime = false;
--- a/src/Management/test/Endpoint.Test/Actuators/Health/HealthAggregationTest.cs
+++ b/src/Management/test/Endpoint.Test/Actuators/Health/HealthAggregationTest.cs
@@ -408,7 +408,8 @@ public sealed class HealthAggregationTest
             }
             """);
 
-        stopwatch.Elapsed.Should().BeGreaterThan(500.Milliseconds()).And.BeLessThan(6.Seconds());
+        // Upper bound must be less than 2+3+4=9s if contributors ran sequentially.
+        stopwatch.Elapsed.Should().BeGreaterThan(500.Milliseconds()).And.BeLessThan(9.Seconds());
     }
 
     [Fact]
--- a/src/Management/test/Endpoint.Test/Actuators/Health/HealthAggregationTest.cs
+++ b/src/Management/test/Endpoint.Test/Actuators/Health/HealthAggregationTest.cs
@@ -366,9 +366,9 @@ public sealed class HealthAggregationTest
     {
         List<IHealthContributor> contributors =
         [
-            new SlowContributor(1.Seconds()),
             new SlowContributor(2.Seconds()),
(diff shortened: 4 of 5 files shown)


src/Management/test/Endpoint.Test/Actuators/Health/HealthAggregationTest.cs
```diff
@@ -408,7 +408,7 @@ public async Task Aggregates_contributors_in_parallel()
             }
             """);
 
-        stopwatch.Elapsed.Should().BeGreaterThan(500.Milliseconds()).And.BeLessThan(5.Seconds());
+        stopwatch.Elapsed.Should().BeGreaterThan(500.Milliseconds()).And.BeLessThan(6.Seconds());
```
bart-vmware:
```suggestion
        stopwatch.Elapsed.Should().BeGreaterThan(500.Milliseconds()).And.BeLessThan(9.Seconds());
```

Adding up the durations when slow contributors would run sequentially.

src/Management/test/Endpoint.Test/Actuators/ThreadDump/EventPipeThreadDumperTest.cs
```diff
@@ -87,11 +94,16 @@ private static class NestedType
     {
         public static void BackgroundThreadCallback(object? argument)
         {
-            var cancellationToken = (CancellationToken)argument!;
+            (CancellationToken cancellationToken, ManualResetEventSlim threadReady) = ((CancellationToken, ManualResetEventSlim))argument!;
```
bart-vmware: ~threadReady~ threadStarted

src/Management/test/Endpoint.Test/Actuators/ThreadDump/EventPipeThreadDumperTest.cs
```diff
@@ -87,11 +94,16 @@ private static class NestedType
...
+
+            threadReady.Set();
 
             while (!cancellationToken.IsCancellationRequested)
             {
-                Thread.Sleep(TimeSpan.FromMilliseconds(50));
+                // Only actively-running threads are shown in the thread dump, so we need to make sure the CPU is in use.
+                // Thread.Sleep(0) yields to allow the EventPipe rundown thread to make progress on .NET 8.
```
bart-vmware: Can this be put in a `#if` if only .NET 8 requires the yield? It looks contradictory to spin to keep the thread busy, but then yield anyway.

src/Management/test/Endpoint.Test/Actuators/ThreadDump/EventPipeThreadDumperTest.cs
```diff
@@ -16,19 +16,26 @@ public sealed class EventPipeThreadDumperTest
...
 
         using var loggerProvider = new CapturingLoggerProvider();
         using var loggerFactory = new LoggerFactory([loggerProvider]);
         ILogger<EventPipeThreadDumper> logger = loggerFactory.CreateLogger<EventPipeThreadDumper>();
 
-        var optionsMonitor = new TestOptionsMonitor<ThreadDumpEndpointOptions>();
+        // Use a longer collection window to provide enough sample opportunities on a loaded CI runner.
+        var optionsMonitor = TestOptionsMonitor.Create(new ThreadDumpEndpointOptions
+        {
+            Duration = 100
```
bart-vmware, line 37: Is this required for all TFMs?

src/Connectors/test/Connectors.Test/RelationalDatabaseHealthContributorTest.cs
```diff
@@ -111,7 +111,10 @@ public async Task SQLServer_Not_Connected_Returns_Down_Status()
         result.Description.Should().Be("SQL Server health check failed");
         result.Details.Should().Contain("host", "localhost");
         result.Details.Should().Contain("service", "Example");
-        result.Details.Should().ContainKey("error").WhoseValue.As<string>().Should().StartWith("SqlException: Connection Timeout Expired.");
+
+        result.Details.Should().ContainKey("error").WhoseValue.As<string>().Should().Match(exception =>
+            exception.StartsWith("SqlException: Connection Timeout Expired.", StringComparison.InvariantCulture) ||
+            exception.StartsWith("SqlException: A network-related or instance-specific error", StringComparison.InvariantCulture));
```
bart-vmware: Could these just use `StringComparison.Ordinal`?

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
```diff
@@ -44,15 +44,20 @@ private async Task TimerLoopAsync(TimeSpan interval)
...
-                LogStartingRefreshCycle();
-
-                try
-                {
-                    await _runner.RunAsync(isFirstTime, _timerTokenSource.Token);
-                }
-                catch (Exception exception) when (!exception.IsCancellation())
+                // A tick queued just before periodic refresh was disabled would still be delivered here.
+                // Checking the period prevents executing a stale tick when refresh has been turned off.
+                if (isFirstTime || _periodicTimer.Period != Timeout.InfiniteTimeSpan)
```
bart-vmware, line 49: I don't understand what problem this solves. It seems to reduce a race condition rather than fix it? I'm assuming the interval change happens due to a configuration reload. The docs on `PeriodicTimer.Period`:
```
/// <remarks>
/// All prior ticks of the timer, including any that may be waiting to be consumed by <see cref="WaitForNextTickAsync"/>,
/// are unaffected by changes to <see cref="Period"/>. Setting <see cref="Period"/> affects only and all subsequent times
/// at which the timer will tick.
/// </remarks>
```

What is the execution flow that benefits from this change?

src/Discovery/test/Eureka.Test/EurekaServiceCollectionExtensionsTest.cs
```diff
@@ -62,7 +62,7 @@ public async Task AddEurekaDiscoveryClient_UsesServerTimeout()
         var appSettings = new Dictionary<string, string?>
         {
             ["Eureka:Client:EurekaServer:ConnectTimeoutSeconds"] = "1",
-            ["Eureka:Client:EurekaServer:RetryCount"] = "1"
+            ["Eureka:Client:EurekaServer:RetryCount"] = "0"
```
bart-vmware, line 65: I suspect this change is fine, because the meaning of `RetryCount` was fixed in v4 to mean: "the number of retries", as opposed to the meaning in 3x: "the number of attempts". Traced this test down to https://github.com/SteeltoeOSS/Steeltoe/blob/3.x/src/Discovery/test/Eureka.Test/EurekaDiscoveryClientBuilderExtensionsTest.cs#L58 in 3.x.

Can you confirm my analysis is correct?

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
TimHess: From what I've been seeing in CI, we've been experiencing an extra call to the refresh the app's listing in SBA after we change the configuration.

This scenario spells it out: 

- timer fires a tick at T=100ms
- at T=99ms the config change sets `Period = InfiniteTimeSpan`
- `WaitForNextTickAsync` still delivers the already-queued tick
- without the guard, `_runner.RunAsync` fires one more time against the now-disabled URL. 

The guard detects that `Period` is now `InfiniteTimeSpan` and skips that stale tick. The test assertion is `BeLessOrEqualTo(1)` (not 0) because the race can also fire in the other direction, where the tick arrives at exactly the right nanosecond before the `Period` write completes.

src/Management/test/Endpoint.Test/Actuators/Health/HealthAggregationTest.cs
TimHess: Technically correct, but it is higher than needed in practice. My work laptop did actually fail this test several times with the original timings, but passed over 150 times in a row with 6 seconds on these slower contributors

src/Discovery/test/Eureka.Test/EurekaServiceCollectionExtensionsTest.cs
TimHess: Correct.

v3 (`EurekaHttpClient.cs`):

`for (var retry = 0; retry < GetRetryCount(Config); retry++)`
RetryCount = total iterations = total attempts. RetryCount=1 → loop runs once (retry=0) → 1 attempt.

v4 (`EurekaClient.cs`):

```csharp
int tryCount = clientOptions.EurekaServer.RetryCount + 1;
for (int attempt = 1; attempt <= tryCount; attempt++)
```
RetryCount = additional retries, +1 added to get total attempts. RetryCount=0 → tryCount=1 → 1 attempt.

src/Management/test/Endpoint.Test/Actuators/ThreadDump/EventPipeThreadDumperTest.cs
TimHess: Not strictly required for all TFMs... it was added to increase sample size across the board, but we can try scoping this to net8 since that's probably the only place it has mattered so far

src/Management/src/Endpoint/SpringBootAdminClient/SpringBootAdminPeriodicRefresh.cs
bart-vmware: Ok, so it's to compensate for the documented behavior of `WaitForNextTickAsync` not respecting the changed period yet if a tick was already queued?

So behavior would still break if the period changes _right after_ entering the added `if` statement? Why is it a problem that an additional refresh happens around the same time as the internal changes in the first place? Which test is being fixed by this, and are its assumptions correct at all?

TimHess: > Ok, so it's to compensate for the documented behavior of `WaitForNextTickAsync` not respecting the changed period yet if a tick was already queued?

Yes.

> So behavior would still break if the period changes _right after_ entering the added `if` statement? 

Yes, but this fix reduces the race window to the duration of the `Period` check itself vs. without the fix where the window is the full timer interval.

>Why is it a problem that an additional refresh happens around the same time as the internal changes in the first place? 

In production, it's arguably not a real problem.

> Which test is being fixed by this, and are its assumptions correct at all?

The test is [`PeriodicRefreshCanBeTurnedOffAfterStart`](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/test/Endpoint.Test/SpringBootAdminClient/HostBuilderTest.cs#L212). As an alternative to the code change, we could make [this assertion](https://github.com/SteeltoeOSS/Steeltoe/blob/main/src/Management/test/Endpoint.Test/SpringBootAdminClient/HostBuilderTest.cs#L268) use `BeLessOrEqualTo(1)`

bart-vmware: Ok, let's keep this then. Changing the test is worse, as it hides the actual problem.

bart-vmware (approved):

PR #1695, opened 2026-05-12
Fix CA1873: Potentially expensive logging

bart-vmware:
## Description

Follow-up of #1652.

Additionally, masks the username/password in URLs returned from the HTTP exchanges actuator.

## Quality checklist

### DIFF
--- a/src/Configuration/src/Abstractions/CompositeConfigurationProvider.cs
+++ b/src/Configuration/src/Abstractions/CompositeConfigurationProvider.cs
@@ -56,12 +56,7 @@ internal abstract partial class CompositeConfigurationProvider : IConfigurationP
         ArgumentNullException.ThrowIfNull(earlierKeys);
 
         string[] earlierKeysArray = earlierKeys as string[] ?? earlierKeys.ToArray();
-
-        if (_logger.IsEnabled(LogLevel.Trace))
-        {
-            string earlierKeyNames = string.Join(", ", earlierKeysArray.Select(key => $"'{key}'"));
-            LogGetChildKeys(GetType().Name, earlierKeyNames, parentPath);
-        }
+        ExpensiveLogGetChildKeys(earlierKeysArray, parentPath);
 
         IConfiguration? section = parentPath == null ? ConfigurationRoot : ConfigurationRoot?.GetSection(parentPath);
 
@@ -75,6 +70,15 @@ internal abstract partial class CompositeConfigurationProvider : IConfigurationP
         keys.AddRange(earlierKeysArray);
         keys.Sort(ConfigurationKeyComparer.Instance);
         return keys;
+    }
+
+    private void ExpensiveLogGetChildKeys(string[] earlierKeysArray, string? parentPath)
+    {
+        if (_logger.IsEnabled(LogLevel.Trace))
+        {
+            string earlierKeyNames = string.Join(", ", earlierKeysArray.Select(key => $"'{key}'"));
+            LogGetChildKeys(GetType().Name, earlierKeyNames, parentPath);
+        }
     }
 
     public virtual bool TryGet(string key, out string? value)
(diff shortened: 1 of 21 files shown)


src/Configuration/src/Abstractions/CompositeConfigurationProvider.cs
```diff
@@ -77,6 +72,15 @@ public IEnumerable<string> GetChildKeys(IEnumerable<string> earlierKeys, string?
         return keys;
     }
 
+    private void ExpensiveLogGetChildKeys(string[] earlierKeysArray, string? parentPath)
```
TimHess, line 75: This new `ExpensiveLog*` convention only makes sense in the context of CA1873, do we really need to include Expensive in the name of all of them? Additionally, this specific change doesn't seem to add much value (other than aligning on a somewhat cosmetic pattern) since we already had the logic in place to check if the logger is enabled.

What about inlining all the `.IsEnabled` checks and dropping the wrapper methods? That way we could also avoid the risk of reusing the log methods that have `SkipEnabledCheck = true,` and forgetting to include the manual check.

For masked URIs we could also use a struct that overrides ToString to contain the expense: 

```csharp
private readonly record struct MaskedUri(Uri Value)
{
    public override string ToString() => Value.ToMaskedString();
}

[LoggerMessage(Level = LogLevel.Debug, Message = "Adding credentials from '{RequestUri}' to Authorization header.")]
private partial void LogAddingCredentials(MaskedUri requestUri);
// Call site:
LogAddingCredentials(new MaskedUri(requestUri));
```

src/Discovery/src/Eureka/EurekaClient.cs
TimHess: Here, line 295 and ConfigServerConfigurationProvider.cs line 690 all call `.ToMaskedString()` and should probably be guarded as well. Apparently the analyzer doesn't catch them because they're at the `Warning` level, but if we're trying to fix this across the board we should probably get them too.

bart-vmware: The analyzer used to flag these, but it got relaxed after customer feedback. Logging at warning level is likely to actually happen, so it doesn't trigger.

src/Configuration/src/Abstractions/CompositeConfigurationProvider.cs
bart-vmware: That's a way better solution, love it!

I'm not thrilled with the `Expensive` prefix either, but I had to come up with something. We're down to 6 occurrences now.

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/HttpExchangesActuatorTest.cs
```diff
@@ -366,7 +366,7 @@ public async Task Respects_maximum_queue_capacity()
                   "timestamp": "2024-09-19T00:00:25",
                   "request": {
                     "method": "GET",
-                    "uri": "http://localhost:80/id/25"
+                    "uri": "http://localhost/id/25"
```
TimHess, line 369: I don't know that it's necessarily a _problem_, but I don't understand why `:80` disappeared from these exchanges, do you know why?

I suppose it's a consequence of `new UriBuilder(source)` being in the flow?

src/Discovery/src/Eureka/EurekaClient.cs
TimHess: I suppose we're focused on the analyzer warning, which technically doesn't apply to those cases and wouldn't be enforced by the tooling in the future anyway... Additionally, while there may be some optimization to be had, it's unlikely to matter since these warnings _should_ always be logged anyway

TimHess (approved):
LGTM

src/Management/test/Endpoint.Test/Actuators/HttpExchanges/HttpExchangesActuatorTest.cs
bart-vmware: The reason is that the JSON serializer treats properties of type `Uri` differently from `string`. See https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Serialization/Converters/Value/UriConverter.cs.

I didn't like that it contained `:80` in the first place (it's not in the data we feed it), but didn't care enough to worry about or follow up on it.

PR #1711, opened 2026-05-29
Update Steeltoe prerequisites in README.md

bart-vmware:
Also added the new releases to https://github.com/SteeltoeOSS/Steeltoe/wiki/Steeltoe-Support-Versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,12 +29,12 @@ In addition to the [feature documentation](https://steeltoe.io/api), we have bui
 
 ### Prerequisites
 
-| Steeltoe Version | .NET Version |
-| --- | --- |
-| 4.x | .NET 8 - 9 |
-| 3.x | .NET Core 3.1 - .NET 8 |
-| 3.x (Integration, Messaging and Stream) | .NET Core 3.1 - .NET 6 |
-| 2.x | .NET Framework 4.6.1+ |
+| Steeltoe Branch | .NET Version | Status |
+| --- | --- | --- |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,12 +29,12 @@ In addition to the [feature documentation](https://steeltoe.io/api), we have bui
...
 
-| Steeltoe Version | .NET Version |
-| --- | --- |
-| 4.x | .NET 8 - 9 |
-| 3.x | .NET Core 3.1 - .NET 8 |
-| 3.x (Integration, Messaging and Stream) | .NET Core 3.1 - .NET 6 |
-| 2.x | .NET Framework 4.6.1+ |
+| Steeltoe Branch | .NET Version | Status |
+| --- | --- | --- |
+| main | .NET 8 - 10 | Unstable |
```
TimHess, line 34: `Unstable` could be read as buggy, glitchy, flawed, etc where this is supposed to be more about how the apis may change... should we say `development` instead?

bart-vmware: This aligns with https://start.steeltoe.io/. It _is_ unstable, in the sense that it is not a thoroughly tested release, but just a snapshot whose overall quality (from a consumption perspective) is unknown. Unstable also indicates it's changing over time and may break existing users.

TimHess: Yeah, I would have a general preference for avoiding the negative connotations. But it _is_ already live with initializr, so here/now isn't the place to set a new standard

TimHess (approved):

PR #1713, opened 2026-05-29
Try running GHA workflows on macos-26, which is going to become macos-latest

bart-vmware:
Experimental, do not merge.


bart-vmware:
Closing, all builds are green.

PR #1717, opened 2026-07-13
Add NuGet package to write git.properties file

bart-vmware:
## Description

Adds a new build-time-only package, **`Steeltoe.Management.GitProperties.Build`**, that generates a `git.properties` file compatible with the [Spring Boot Actuator format](https://docs.spring.io/spring-boot/reference/actuator/endpoints.html#actuator.endpoints.info.git-commit-information), and wires it into the existing `Info` actuator endpoint (`GitInfoContributor`) so the values are automatically exposed at runtime.

### How it works

- New MSBuild tasks project (`src/Management/src/GitProperties.Build`) hooks into the build pipeline, auto-detected when a project references `Steeltoe.Management.Endpoint` (configurable via `GenerateGitProperties`/`GitPropertiesConsumingPackageIds`).
- Repository-wide info (commit id, branch, tags, remote URL, dirty state, etc.) is computed **at most once per build** and cached in a shared `obj/git.properties.cache` file at the repo root, reused by every project/TFM in a solution. The cache refreshes automatically via MSBuild incremental-build Inputs tracking (HEAD, config, packed-refs, refs/**) whenever something changes in the repo.
- Git worktrees and submodules are each treated as independent repositories with their own shared cache.
- For deployments that don't carry `.git` along (e.g. `cf push`), `dotnet build -t:WriteGitPropertiesFallbackFile` writes a durable fallback copy that's used automatically whenever no live `.git` is available.
- `GitInfoContributor` (runtime side) now checks next to the app's own assembly first, falling back to the current working directory, so it finds `git.properties` regardless of how the app is launched.

### Diagnostics

Seven `GITPROPS00x` codes cover every place generation can be skipped or left partial (repository not found, invalid, git executable missing, incompatible version, no commits, shallow clone, dirty-state unknown) — see the package's [Diagnostics table](https://github.com/SteeltoeOSS/Steeltoe/blob/git-properties-msbuild/src/Management/src/GitProperties.Build/PackageReadme.md#diagnostics). All are forgiving by default: generation is skipped with a suppressible warning instead of failing the build, so it's safe to add this package to projects that aren't always built inside a git checkout (e.g. a Docker build stage).

### Known, documented limitations

Called out explicitly in the package readme's "Good to know" section:
- File *deletions* inside `.git` (e.g. `git tag -d`, `git fetch --unshallow`) don't invalidate the shared cache by themselves — it catches up the next time something else changes.
- Changing your *global* git username/email isn't tracked by the shared cache either, for the same reason.

### CI changes

- The test suite spawns real nested `dotnet build`/`publish` subprocesses per test, which fragments coverage data (ephemeral temp paths, duplicate `<class>` entries per subprocess); `merge-subprocess-coverage.ps1` repairs that before upload.
- A `Category=GitProperties` xUnit trait (on the test project's shared base class) is used to run the tests isolated in cibuilds, which is needed because coverage is collected differently: Coverlet is unable to instrument subprocesses.
- `COMMON_TEST_ARGS`/`SONAR_TEST_ARGS` now only contain flags genuinely shared by every `dotnet test` invocation in each workflow; VSTest-coverage-collector-specific flags moved to their own `COVERAGE_COLLECT_ARGS` var, since this suite collects coverage its own way and can't use them.

## Test plan

- 43 new tests in `GitProperties.Build.Test`, covering auto-detection, all 7 diagnostics (including both trigger paths for the 3 diagnostics that can fire for two different reasons), the shared cache (multi-project, multi-TFM, incremental build, build-time-vs-commit-time), the fallback file (including `.gitignore` interaction), worktrees, submodules, publish/push scenarios, and property-content correctness (branch env var fallback, non-ASCII data, multiple remotes, shallow clones).
- 1 new unit test (`GitInfoContributorTest`) for the runtime path-resolution change.

## Quality checklist


.github/workflows/sonarcube.yml
```diff
@@ -82,13 +83,17 @@ jobs:
...
+      run: dotnet tool restore --verbosity minimal
+
     - name: Begin Sonar .NET scanner
       id: sonar_begin
       env:
         SONAR_TOKEN: ${{ secrets.SONAR_TOKEN }}
       run: >-
         dotnet sonarscanner begin /k:"SteeltoeOSS_steeltoe" /o:"steeltoeoss" /d:sonar.token="${{ secrets.SONAR_TOKEN }}"
         /d:sonar.host.url="https://sonarcloud.io" /d:sonar.cs.opencover.reportsPaths=**/coverage.opencover.xml
+        /d:sonar.cs.cobertura.reportsPaths=**/*.cobertura.xml
```
TimHess, line 96: Is there any value in keeping opencover around?

.github/workflows/Steeltoe.All.yml
```diff
@@ -22,8 +22,9 @@ env:
   DOTNET_NOLOGO: true
   SOLUTION_FILE: 'src/Steeltoe.All.slnx'
   COMMON_TEST_ARGS: >-
-    --no-build --configuration Release --collect "XPlat Code Coverage" --logger trx --results-directory ${{ github.workspace }}/TestOutput
-    --settings coverlet.runsettings --blame-crash --blame-hang-timeout 1m
+    --no-build --configuration Release --logger trx --results-directory ${{ github.workspace }}/TestOutput
+    --blame-crash --blame-hang-timeout 5m
```
TimHess, line 26: Should this change to blame-hang-timeout be scoped down to just the project that needs it? Does it still even need to be 5m after the test optimization effort you went through?

I'm not sure that the benefits of varying the timeout would be worth the complexity, or that it's even that much of a real concern.

src/Management/src/GitProperties.Build/AtomicFile.cs
```diff
@@ -0,0 +1,150 @@
...
+    public static void Write(string path, List<string> lines)
+    {
+        string? directory = Path.GetDirectoryName(path);
+
+        if (!string.IsNullOrEmpty(directory))
+        {
+            Directory.CreateDirectory(directory);
+        }
+
+        string tempPath = Path.Combine(directory ?? string.Empty, $"{Path.GetRandomFileName()}~");
```
TimHess: Should this file name be slightly less random so that it's easier to identify where it came from?

src/Management/src/GitProperties.Build/build/Steeltoe.Management.GitProperties.Build.targets
```diff
@@ -0,0 +1,187 @@
...
+    <GitCommitIdAbbrevLength Condition="'$(GitCommitIdAbbrevLength)' == ''">7</GitCommitIdAbbrevLength>
+  </PropertyGroup>
+
+  <ItemGroup Condition="Exists('$(GitPropertiesFallbackFile)')">
+    <!-- Outside any Target, so this is evaluated early enough to hide the fallback file from Solution Explorer. -->
+    <None Remove="$(GitPropertiesFallbackFile)" />
+  </ItemGroup>
+
+  <!--
+    Tasks intentionally don't run when DesignTimeBuild is true, to prevent shelling out to git on every keystroke.
```
TimHess: Should this comment move closer to line 49, where it's applicable for the first time? I was initially a bit confused what this was relating to since it's kind of floating here

src/Management/src/GitProperties.Build/GitPropertiesFormat.cs
```diff
@@ -0,0 +1,21 @@
...
+
+/// <summary>
+/// Rules specific to the content/shape of a git.properties file.
+/// </summary>
+internal static class GitPropertiesFormat
+{
+    public const string CommitIdDescribeKey = "git.commit.id.describe";
+
+    /// <summary>
+    /// Collapses line breaks to a literal "\n" so a value can never span multiple physical lines. GitInfoContributor can't handle multiline values.
```
TimHess, line 15: Is a limitation in our `GitInfoContributor`, or does it also apply to what Spring is producing/reading? 

We should probably align pretty closely with Spring on this one... I'd be ok with logging an issue and treating it as a follow-up if/when it actually matters

.github/workflows/sonarcube.yml
bart-vmware: It is still needed for all other Steeltoe code. Switching everything to Cobertura is a much larger change.

src/Management/src/GitProperties.Build/GitPropertiesFormat.cs
bart-vmware: Both are true. The Steeltoe limitation is intentional and matches what real Spring Boot `git.properties` files look like. This is standard `java.util.Properties` escaping behavior.

src/Management/src/GitProperties.Build/AtomicFile.cs
bart-vmware: Updated to generate a file named `git.properties.cache.<random>~`.

src/Management/src/GitProperties.Build/build/Steeltoe.Management.GitProperties.Build.targets
bart-vmware: I've moved the comment, but still as a separate section because it applies to _all_ entry-level tasks (both existing and future ones), not just the first one. Where the `$(DesignTimeBuild)` condition actually appears depends on the task dependency tree and may change over time, so it remains a general concern.

.github/workflows/Steeltoe.All.yml
bart-vmware: I'm not sure. Just googling for it:

image

The GitProperties tests highly benefit from parallelism. On my (fast) machine, the longest test takes 50 seconds, so 5 minutes should give sufficient headroom for slow build servers.

I should also note this setting is specific to _hangs_, which is quite uncommon. So bumping it doesn't affect failed assertions, which sometimes happen due to flakiness.

bart-vmware: If we do the split, then `component-shared-workflow.yml` also needs to run GitProperties tests separately. Now it doesn't, because coverage is not being collected.

.github/workflows/sonarcube.yml
TimHess: Is it worth logging a follow-up for it, or do you think it's likely to happen as part of moving to MTPv2?

.github/workflows/Steeltoe.All.yml
TimHess: Let's keep it like this and adjust only if needed

TimHess (approved):
LGTM

PR #1718, opened 2026-07-15
Fix broken build on .NET SDK v10.0.302

bart-vmware:
Drop the dependencies on `Microsoft.IdentityModel.Protocols.OpenIdConnect` and `Microsoft.IdentityModel.Tokens` from libraries and tests. Their explicit versions conflict, while they are already pulled in by `Microsoft.AspNetCore.Authentication.JwtBearer`.

From what I can tell, these dependencies were introduced in #1311 (Steeltoe 4.0). My assumption is that they were added at the time to be able to use newer APIs or use newer versions that contained fixes we needed.


TimHess:
> From what I can tell, these dependencies were introduced in #1311 (Steeltoe 4.0). My assumption is that they were added at the time to be able to use newer APIs or use newer versions that contained fixes we needed.

The 10.0.10 release of JwtBearer bumped the OpenIdConnect ref from 8.0.1 to 8.19.2... I don't recall the specific reason for our manual bump, but your assumption seems likely to be accurate

TimHess (approved):

PR #1719, opened 2026-07-20
Log instead of throwing when localhost is used for registry server

TimHess:
## Description

I've recently tried running both Eureka and apps inside a devcontainer, and that's prompted me to revisit #829.

Since this is a containerized environment where apps can run as native executables, we now have a use case where 'localhost' is a valid hostname for the service registry.

## Quality checklist

### DIFF
--- a/src/Discovery/src/Consul/PostConfigureConsulOptions.cs
+++ b/src/Discovery/src/Consul/PostConfigureConsulOptions.cs
@@ -11,15 +11,15 @@ namespace Steeltoe.Discovery.Consul;
 
 internal sealed partial class PostConfigureConsulOptions : IPostConfigureOptions<ConsulOptions>
 {
-    private readonly IOptionsMonitor<ConsulDiscoveryOptions> _discoveryOptionsMonitor;
+    private readonly IOptionsMonitor<ConsulDiscoveryOptions> _optionsMonitor;
     private readonly ILogger<PostConfigureConsulOptions> _logger;
 
-    public PostConfigureConsulOptions(IOptionsMonitor<ConsulDiscoveryOptions> discoveryOptionsMonitor, ILogger<PostConfigureConsulOptions> logger)
+    public PostConfigureConsulOptions(IOptionsMonitor<ConsulDiscoveryOptions> optionsMonitor, ILogger<PostConfigureConsulOptions> logger)
     {
-        ArgumentNullException.ThrowIfNull(discoveryOptionsMonitor);
+        ArgumentNullException.ThrowIfNull(optionsMonitor);
         ArgumentNullException.ThrowIfNull(logger);
 
-        _discoveryOptionsMonitor = discoveryOptionsMonitor;
+        _optionsMonitor = optionsMonitor;
         _logger = logger;
     }
 
@@ -27,14 +27,14 @@ internal sealed partial class PostConfigureConsulOptions : IPostConfigureOptions
     {
         ArgumentNullException.ThrowIfNull(options);
 
-        if (_discoveryOptionsMonitor.CurrentValue.Enabled && (Platform.IsContainerized || Platform.IsCloudHosted) && options.Host == "localhost")
+        if (_optionsMonitor.CurrentValue.Enabled && (Platform.IsContainerized || Platform.IsCloudHosted) && options.Host == "localhost")
         {
-            LogLocalhostConsulUrl(_logger, $"{options.Scheme}://{options.Host}:{options.Port}");
+            LogLocalhostConsulUrl($"{options.Scheme}://{options.Host}:{options.Port}");
         }
     }
 
     [LoggerMessage(EventId = 0, Level = LogLevel.Warning,
         Message = "Consul URL '{Url}' is unlikely to be valid in containerized or cloud environments. " +
             "Please configure Consul:Host with a non-localhost server.")]
-    private static partial void LogLocalhostConsulUrl(ILogger logger, string url);
+    private partial void LogLocalhostConsulUrl(string url);
 }
--- a/src/Discovery/src/Consul/PostConfigureConsulOptions.cs
+++ b/src/Discovery/src/Consul/PostConfigureConsulOptions.cs
@@ -9,20 +9,21 @@ using Steeltoe.Discovery.Consul.Configuration;
 
 namespace Steeltoe.Discovery.Consul;
 
-internal sealed partial class ValidateConsulOptions : IValidateOptions<ConsulOptions>
+internal sealed partial class PostConfigureConsulOptions : IPostConfigureOptions<ConsulOptions>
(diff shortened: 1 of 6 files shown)


src/Discovery/src/Consul/ValidateConsulOptions.cs
```diff
@@ -2,21 +2,24 @@
...
 
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Steeltoe.Common;
 using Steeltoe.Discovery.Consul.Configuration;
 
 namespace Steeltoe.Discovery.Consul;
 
-internal sealed class ValidateConsulOptions : IValidateOptions<ConsulOptions>
+internal sealed partial class ValidateConsulOptions : IValidateOptions<ConsulOptions>
```
bart-vmware: It looks like there is nothing left to validate anymore, so the logging may as well move directly into `PostConfigureConsulDiscoveryOptions`, and then this type can be removed.

src/Discovery/test/Consul.Test/ConsulServiceCollectionExtensionsTest.cs
```diff
@@ -142,22 +143,28 @@ public async Task RegistersHostedService()
     }
 
     [Fact]
-    public async Task ConsulOptionsValidation_FailsWhenRunningInCloudWithLocalhost()
+    public async Task ConsulOptionsValidation_WarnsWhenRunningInCloudWithLocalhost()
```
bart-vmware:
```suggestion
    public async Task PostConfigureConsulOptions_WarnsWhenRunningInCloudWithLocalhost()
```

src/Discovery/test/Consul.Test/ConsulServiceCollectionExtensionsTest.cs
```diff
@@ -142,22 +143,28 @@ public async Task RegistersHostedService()
...
+
+        string? logEntry = capturingLoggerProvider.GetAll().Should().ContainSingle().Which;
 
-        action.Should().ThrowExactly<OptionsValidationException>().WithMessage(
-            "Consul URL 'http://localhost:8500' is not valid in containerized or cloud environments. " +
-            "Please configure Consul:Host with a non-localhost server.");
+        logEntry.Should()
+            .Be(
+                $"WARN {typeof(PostConfigureConsulOptions).FullName}: Consul URL 'http://localhost:8500' is unlikely to be valid in containerized or cloud environments. " +
+                "Please configure Consul:Host with a non-localhost server.");
```
bart-vmware:
```suggestion
        logEntry.Should().Be(
            $"WARN {typeof(PostConfigureConsulOptions).FullName}: Consul URL 'http://localhost:8500' is unlikely to be valid in containerized or cloud environments. " +
            "Please configure Consul:Host with a non-localhost server.");
```

src/Discovery/src/Consul/PostConfigureConsulOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+        if (_discoveryOptionsMonitor.CurrentValue.Enabled && (Platform.IsContainerized || Platform.IsCloudHosted) && options.Host == "localhost")
+        {
+            LogLocalhostConsulUrl(_logger, $"{options.Scheme}://{options.Host}:{options.Port}");
+        }
+    }
+
+    [LoggerMessage(EventId = 0, Level = LogLevel.Warning,
+        Message = "Consul URL '{Url}' is unlikely to be valid in containerized or cloud environments. " +
+            "Please configure Consul:Host with a non-localhost server.")]
+    private static partial void LogLocalhostConsulUrl(ILogger logger, string url);
```
bart-vmware: Please use a non-static method without the `ILogger` parameter:
```c#
private partial void LogLocalhostConsulUrl(string url);
```
The source generator will pick the `_logger` field automatically.

src/Discovery/src/Consul/PostConfigureConsulOptions.cs
```diff
@@ -0,0 +1,40 @@
...
+using Microsoft.Extensions.Logging;
+using Microsoft.Extensions.Options;
+using Steeltoe.Common;
+using Steeltoe.Discovery.Consul.Configuration;
+
+namespace Steeltoe.Discovery.Consul;
+
+internal sealed partial class PostConfigureConsulOptions : IPostConfigureOptions<ConsulOptions>
+{
+    private readonly IOptionsMonitor<ConsulDiscoveryOptions> _discoveryOptionsMonitor;
```
bart-vmware: This could just be named `_optionsMonitor`, we don't have multiple. Same for constructor parameter.

src/Discovery/src/Eureka/ValidateEurekaClientOptions.cs
```diff
@@ -2,13 +2,14 @@
...
 
+using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using Steeltoe.Common;
 using Steeltoe.Discovery.Eureka.Configuration;
 
 namespace Steeltoe.Discovery.Eureka;
 
-internal sealed class ValidateEurekaClientOptions : IValidateOptions<EurekaClientOptions>
+internal sealed partial class ValidateEurekaClientOptions(ILogger<ValidateEurekaClientOptions> logger) : IValidateOptions<EurekaClientOptions>
```
bart-vmware: Should use an explicit constructor that null-checks its inputs.

src/Discovery/test/Eureka.Test/EurekaClientOptionsTest.cs
```diff
@@ -166,28 +167,6 @@ public async Task ClientDisabledBySpringCloudDiscovery_EnabledFalse()
         clientOptions.Value.Enabled.Should().BeFalse();
     }
 
-    [Fact]
-    public async Task ClientOptionsValidation_SucceedsWhenEurekaIsTurnedOff()
```
bart-vmware, line 170: I wonder if we should keep this test, asserting the warning is _not_ logged?

bart-vmware (changes requested):
Some of the feedback for Consul also applies to Eureka.
And could you fix the vulnerability job failure by adding an explicit ref to non-vulnerable https://www.nuget.org/packages/System.Security.Cryptography.Xml/8.0.4 with a `net8.0` `Condition` on it?

TimHess:
> Some of the feedback for Consul also applies to Eureka. And could you fix the vulnerability job failure by adding an explicit ref to non-vulnerable https://www.nuget.org/packages/System.Security.Cryptography.Xml/8.0.4 with a `net8.0` `Condition` on it?

The vulnerability job failure was a false positive, it was resolved upstream

bart-vmware (approved):
LGTM, thanks!

Issue #1359, opened 2024-09-19
Escaped dots in Spring configuration keys are not properly translated


bart-vmware:
## Describe the bug
As described at https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.yaml.mapping-to-properties, Spring Boot supports escaping dots in configuration keys:

> When binding to scalar values, keys with . in them do not need to be surrounded by []. Scalar values include enums and all types in the java.lang package except for Object. Binding a.b=c to Map<String, String> will preserve the . in the key and return a Map with the entry {"a.b"="c"}. For any other types you need to use the bracket notation if your key contains a .. For example, binding a.b=c to Map<String, Object> will return a Map with the entry {"a"={"b"="c"}} whereas [a.b]=c will return a Map with the entry {"a.b"="c"}.

Based on that, Steeltoe should translate the Spring configuration key `a.b.c` to `a:b:c` (which it already does). But also translate `[a.b].c` to `a.b:c`, which is currently not the case. And translate arrays, such as `a[0].b` to `a:0:b` (which it already does).

Because [there's no way to escape colons in .NET configuration keys](https://github.com/dotnet/runtime/issues/67616), there's nothing Steeltoe can do to properly translate when the Spring key contains a colon.

This affects the various Steeltoe configuration providers, such as Placeholder, SpringBoot, and Config Server. Key translation isn't correct everywhere and there's limited test coverage.

PR #1722, opened 2026-07-29
Add a CredHub PostProcessor

### LINKED ISSUE #28: Work with CredHub Service Broker
As a Cloud Foundry user, I want to be able to use the CredHub Service Broker https://docs.pivotal.io/credhub-service-broker/index.html in .NET apps. How do I do that?

TimHess:
## Description

Adds a PostProcessor for reading instances of CredHub service bindings

Fixes #28

## Quality checklist


src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
```diff
@@ -124,4 +124,62 @@ public void AddCloudFoundryServiceBindings_CanIgnoreKey()
...
+            }
+            """;
+
+        var reader = new StringServiceBindingsReader(credHubJson);
+        var builder = new ConfigurationBuilder();
+        builder.AddCloudFoundryServiceBindings(reader);
+        IConfigurationRoot configurationRoot = builder.Build();
+
+        const string keyPrefix = "steeltoe:service-bindings:credhub:my-credhub-service:";
+        configurationRoot.GetValue<string>($"{keyPrefix}Encrypt__Key").Should().Be("secret-value");
```
bart-vmware: Why shouldn't substitution happen here?

src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
```diff
@@ -124,4 +124,62 @@ public void AddCloudFoundryServiceBindings_CanIgnoreKey()
...
+
+        configurationRoot.GetValue<string>($"{keyPrefix}app:database:connection_string").Should().Be("Server=tcp:sql.domain;Database=db;");
+
+        configurationRoot.GetValue<string>($"{keyPrefix}app:database:max_pool_size").Should().Be("20");
+        configurationRoot.GetValue<string>($"{keyPrefix}app:feature__flag").Should().Be("enabled");
+        configurationRoot.GetValue<string>($"{keyPrefix}endpoints:0").Should().Be("https://api1.example.com");
+        configurationRoot.GetValue<string>($"{keyPrefix}endpoints:1").Should().Be("https://api2.example.com");
+
+        configurationRoot.GetValue<string>("Encrypt:Key").Should().Be("secret-value");
+        configurationRoot.GetValue<string>("some.setting").Should().Be("setting-value");
```
bart-vmware: Should this substitute `.` with `:`, so that users can share keys between Spring and .NET?

src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
TimHess: I'm on the fence about even mapping the keys to the `steeltoe:service-bindings` location at all... with the way it is right now, each of these keys will be present in three locations (whether or not any substitutions happen) - should we drop this additional mapping entirely?

src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
TimHess: I don't think we should do that, at least not right now. I believe credhub service instances are effectively cheap/free, and while it might be convenient to share them, we don't currently have a use-case for that situation. We also have no way of distinguishing between keys that should be shared and intentional `.` usage (something like this made up example shouldn't get translated: `maps.google.com:apikey`)

bart-vmware: Spring supports preserving literal dots in a key [by escaping them](https://stackoverflow.com/questions/34070987/escaping-a-dot-in-a-map-key-in-yaml-in-spring-boot), but they are unfolded into subkeys by default. A similar request for .NET was [rejected](https://github.com/dotnet/runtime/issues/67616). So by default, every dot/colon unfolds into a subkey; therefore, both Spring and .NET suffer from a key like: `https://maps.google.com`. 

Because Steeltoe being unable to handle the Spring escape format is a [known limitation](https://github.com/SteeltoeOSS/Steeltoe/issues/1359), we don't need to worry about that here. But I do think we should do the dot-to-colon substitution, because Steeltoe is designed to add .NET support to a mostly-Java-based architecture, and the translation is easy to implement. I think that sharing secrets enterprise-wide in a shared vault is a common scenario.

I've tried Config Server by injecting the following files into the Docker container:

test.properties:
```text
some.[props.key]=Example
```

test.yaml:
```yaml
some:
  "[yaml.key]":
    Example
```

And here is what Steeltoe stores in `IConfiguration`:
image

As shown above, Steeltoe can't handle the brackets (the known limitation), but it _does_ convert dots to colons.

src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
bart-vmware: I agree that adding them to `steeltoe:service-bindings` is a bit too much. If users appear to require it to disambiguate duplicate keys, we can add it on request in a future version. I also don't think we should translate `__` to `:` here, because it's a .NET-only convention for reading _environment variables_ into `IConfiguration`, which does not apply here (similarly, we don't strip the `DOTNET_` and `ASPNETCORE_` prefixes).

TimHess: Fair enough. I'll have to make an adjustment somewhere else, but we can discuss that there

src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
TimHess: Rather than just accept that condition as a known limitation, I'd like to take a crack at solving it here since this is a brand new feature... if we like it, we can centralize it later

src/Configuration/test/CloudFoundry.Test/ServiceBindings/ConfigurationBuilderExtensionsTest.cs
```diff
@@ -124,4 +124,58 @@ public void AddCloudFoundryServiceBindings_CanIgnoreKey()
...
+                  "plan": "default",
+                  "credentials": {
+                    "Encrypt__Key": "secret-value",
+                    "some.setting": "setting-value",
+                    "app": {
+                      "database": {
+                        "connection_string": "Server=tcp:sql.domain;Database=db;",
+                        "max_pool_size": "20"
+                      },
+                      "feature__flag": "enabled"
```
bart-vmware: This entry is redundant now, I think. Can we replace it with an array? I suspect the bracket-unescape conflicts with that, but I'm not sure. See https://stackoverflow.com/questions/69648470/how-to-define-and-read-array-of-objects-from-springs-application-properties.

bart-vmware: This is not what I meant; the dots are now in the values. So the added `servers` is basically the same as the existing `endpoints`.

My point is that a _key_ may contain both array elements and escaped dots. For example, a key like `path.[with.a.[0].in[dex]`, resulting from:

```json
{
  "credentials": {
    "path": {
      "with.a.[0].in]dex": "value"
    }
  }
}
```

How would that show up in `IConfiguration`, if at all? Should it warn on keys that cannot be represented?

I still don't think we _need_ to support escaping, but if we do, it should be implemented correctly. The parser looks way too simplistic. It all depends on what can actually appear in the CredHub data, and how it gets there. Needs a full end-to-end experiment to find out.

TimHess: That particular example doesn't appear well-formed, and I'd like to think that an array just doesn't make any sense at all in this scenario... But it's mostly hypothetical and I can't say for sure either way whether arrays might be used here, so I guess I'll just remove the escaping until/unless we get some feedback with more specific use cases

src/Configuration/src/CloudFoundry/ServiceBindings/PostProcessors/CredHubCloudFoundryPostProcessor.cs
```diff
@@ -0,0 +1,56 @@
...
+
+            foreach ((string fullKey, string? value) in configurationData.ToArray())
+            {
+                if (!fullKey.StartsWith(credentialsPrefix, StringComparison.OrdinalIgnoreCase))
+                {
+                    continue;
+                }
+
+                string keyWithoutPrefix = fullKey[credentialsPrefix.Length..];
+                string normalizedKey = keyWithoutPrefix.Replace(".", ConfigurationPath.KeyDelimiter, StringComparison.Ordinal);
```
bart-vmware, line 41:
```suggestion
                // Dots in a credential key are converted to colons so secrets can be shared between Spring and .NET apps.
                string keyWithoutPrefix = fullKey[credentialsPrefix.Length..];
                string normalizedKey = keyWithoutPrefix.Replace(".", ConfigurationPath.KeyDelimiter, StringComparison.Ordinal);
```

src/Configuration/src/CloudFoundry/ServiceBindings/PostProcessors/CredHubCloudFoundryPostProcessor.cs
```diff
@@ -0,0 +1,56 @@
...
+// See the LICENSE file in the project root for more information.
+
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Logging;
+
+namespace Steeltoe.Configuration.CloudFoundry.ServiceBindings.PostProcessors;
+
+// CredHub credentials have no fixed schema: the keys are arbitrary secret names chosen by whoever created the CredHub credential.
+// Each credential is written to the root of the configuration, so it can be read the same way a literal environment variable would be.
+// Dots in a credential key are converted to colons so secrets can be shared between Spring and .NET apps.
```
bart-vmware:
```suggestion
// CredHub credentials have no fixed schema: the keys are arbitrary secret names chosen by whoever created the CredHub credential.
// Each credential is written to the root of the configuration, so it can be read the same way a literal environment variable would be.
```

src/Configuration/test/CloudFoundry.Test/ServiceBindings/PostProcessorsTest.cs
```diff
@@ -336,6 +338,64 @@ public void Processes_Identity_configuration()
...
+
+        postProcessor.PostProcessConfiguration(provider, configurationData);
+
+        string keyPrefix = GetOutputKeyPrefix(TestBindingName, CredHubCloudFoundryPostProcessor.BindingType);
+        configurationData.Should().NotContainKey($"{keyPrefix}:simple");
+        configurationData.Should().ContainKey("Encrypt__Key").WhoseValue.Should().Be("test-encrypted-key-value");
+        configurationData.Should().ContainKey("simple").WhoseValue.Should().Be("test-simple-value");
+
+        // Dots are converted to colons, so secrets can be shared between Spring and .NET apps.
+        configurationData.Should().ContainKey("some:setting").WhoseValue.Should().Be("test-setting-value");
```
bart-vmware:
```suggestion
        configurationData.Should().ContainKey("some:setting").WhoseValue.Should().Be("test-setting-value");
```

bart-vmware (approved):

PR #1724, opened 2026-08-02
`git.properties`: support transitive references from shared projects

bart-vmware:
## Summary

`Steeltoe.Management.GitProperties.Build` can now be referenced from a shared project (like an Aspire `ServiceDefaults`) and made transitive, so every host app referencing that shared project gets its own `git.properties` automatically — without adding the package to each host app individually. The test suite was also overhauled to consume the package the same way a real application does, most notably by introducing a fake `Steeltoe.Management.Endpoint` package so tests exercise the real auto-detection path instead of bypassing it.

## Production change

- The package's `.targets` file is now packed only into `buildTransitive\` (previously `build\`).
- `PackageReadme.md`:
  - States explicitly that the package adds no runtime dependency to the application.
  - Documents that the "dirty" check is never cached: each project re-runs `git status` on every build, even though commit/branch/tag info comes from the shared cache.
  - Adds a new "Using this package in a shared project" section explaining how to install the package once in a shared project (e.g. `ServiceDefaults`) with `PrivateAssets="none"` and `GenerateGitProperties=false`, so host apps referencing that shared project pick up the transitive reference and generate their own `git.properties` via the default `auto` setting, including a `ServiceDefaults.csproj` example and a caution block about the added per-project `git` cost.
  - Removes the old "Build-time only" and "Performance impact" bullets from "Good to know" (superseded by the points above).

### Fake `Steeltoe.Management.Endpoint` package

A new `FakeSteeltoeManagementEndpointPackager` builds and packs an empty, fake `Steeltoe.Management.Endpoint` package into the same isolated local NuGet feed used for `Steeltoe.Management.GitProperties.Build`. It's packed once per test session.

The default `TestApp` now references both the real packed `Steeltoe.Management.GitProperties.Build` package and this fake `Steeltoe.Management.Endpoint` package via genuine `PackageReference`s, and no longer forces `GenerateGitProperties=true`. This makes it match the shape of a real Steeltoe host app and exercises the actual `auto`-detection code path (which looks for a `Steeltoe.Management.Endpoint` reference in the dependency graph) instead of sidestepping it. Several `AutoDetection` and `SharedCache` tests were simplified to use `repository.TestApp` directly, or to reference `Workspace.FakeEndpointPackageReference`, instead of building a bespoke dependency/consumer project pair per test.

### Packing and NuGet isolation

- `GitPropertiesSourcePackager` now `dotnet build`s the real `Steeltoe.Management.GitProperties.Build` project directly from its source location in the repository (instead of first copying the source tree and shared build-infrastructure files like `shared.props`, `stylecop.json`, `PackageReadme.md`, etc. into a temporary directory) and moves the resulting `.nupkg` into a session-scoped NuGet feed directory.
- `PackGitPropertiesSourceOnceFixture` (renamed from `CompileGitPropertiesBuildOnceFixture`) creates one session directory per test run, containing `feed` and `packages` subdirectories, and packs both the real and fake packages into it once via `[assembly: AssemblyFixture]`.
- Every test project's `nuget.config` now also pins `globalPackagesFolder` to the session's `packages` directory, so package extraction is isolated per test run instead of relying on the machine-wide NuGet cache.
- Every test now consumes `Steeltoe.Management.GitProperties.Build` via a real `PackageReference` restored from that feed. The previous dedicated `PublishPush/NuGetPackageConsumedViaPackageReferenceGeneratesGitPropertiesTest` (which built and restored an isolated one-off package/consumer to prove this) was removed since it's now implicit in every other test.

### Temp-directory handling

- `ProcessRunner` now redirects the `TMP`/`TEMP`/`TMPDIR` environment variables of every spawned `dotnet` process to a session-scoped temp directory. This is a workaround for https://github.com/dotnet/msbuild/issues/6219, where MSBuild/NuGet processes otherwise leave behind stray directories directly under the machine's real temp folder.
- Each test project now lives under `tests\<TestClassName>` inside the single session directory, rather than each test creating its own top-level directory directly under the machine temp folder.
- Cleanup of read-only attributes on disposal now only recurses into `.git` directories, instead of walking every file in the test directory (including `bin`/`obj` output), since only files under `.git` are ever marked read-only.

### `GitRepository` API

Collapsed from several narrowly-named, single-purpose methods (`AddProjectAsync`, `AddDependencyProjectAsync`, `AddTestAppReferencingAsync`, `AddPackageConsumerProjectAsync`) into two general, composable ones:

- `AddTestAppAsync(name, packageReferences, projectReferences, targetFrameworks)`
- `AddTestLibraryAsync(name, generateGitProperties, packageReferences, projectReferences)`

Both build their project file via the new `ProjectFileBuilder`, `PackageReference` (`PackageId`, `Version`, `PrivateAssets`), and `ProjectReference` (`ProjectName`) types, replacing hand-rolled, duplicated string interpolation previously spread across `TestProjectWriter`.

`CloneAsShallowAsync`, `AddWorktreeAsync`, `AddSubmoduleAsync`, and `SimulatePush` now reference the `TestApp` project that Git itself already copied into the clone/worktree/submodule/push destination (`GetExistingTestApp`), instead of regenerating a brand-new test app project there.

### Other changes

- `TestProjectWriter.cs` reduced from ~300 to ~90 lines, now only responsible for writing files for an already-built `ProjectFileBuilder` (`WriteAppProjectAsync`, `WriteLibraryProjectAsync`, `WriteFakeGitProjectAsync`, `WriteNonZeroExitCodeGitProjectAsync`, `WriteNuGetConfigAsync`).
- Renamed `GitPropertiesBuildTestBase` → `GitPropertiesTestBase`, `GitPropertiesTestWorkspace` → `TestWorkspace`, and `PublishPush\` → `Publishing\`.
- New `Transitivity\PrivateAssetsControlsPackageReferenceTransitivityTest` verifies the `ServiceDefaults` scenario end-to-end for all three `PrivateAssets` values (unset/`none`/`all`): a host app generates `git.properties` only when the library it references has `PrivateAssets` unset or `"none"`, and the library itself never generates its own.
- Removed two `AutoDetection` tests (`AutoDetectionGeneratesGitPropertiesWhenConsumingPackageReferencedTest`, `AutoDetectionExplicitFalseWinsOverDetectedConsumingPackageReferenceTest`) that became redundant once `TestApp` itself started referencing the fake Endpoint package and using the `auto` default — their behavior is now covered incidentally by `PropertyContent\GitPropertiesMatchGitOutputTest` and by the `Diagnostics\*WarnsByDefaultTest` tests, respectively.

## Quality checklist


src/Management/src/GitProperties.Build/PackageReadme.md
```diff
@@ -100,16 +102,39 @@ This command writes an extra copy of `git.properties` directly next to your proj
...
 
+## Using this package in a shared project
+
+Installing this package as shown under [Getting started](#getting-started), whether with the `dotnet` CLI or Visual Studio's Add Package dialog, writes `PrivateAssets="all"` into the `<PackageReference>` line automatically. This stops the reference from becoming transitive, so `git.properties` generation stays local to the project you installed it in. That's the right choice for most solutions: usually only a few host apps need `git.properties`, so keeping the reference non-transitive avoids running Git commands anywhere else.
+
+Some solutions centralize Steeltoe registration in one shared project instead, for example a .NET Aspire `ServiceDefaults` project used by several host apps. In that case, install this package once, in the shared project, and make the reference transitive so every host app referencing it generates its own `git.properties` file.
```
TimHess: ".NET Aspire" usage should be adjusted, here's a zero thought application of that idea (I'm not sure if you still want to mention .NET anywhere else in the sentence):
```suggestion
Some solutions centralize Steeltoe registration in one shared project instead, for example an Aspire `ServiceDefaults` project used by several host apps. In that case, install this package once, in the shared project, and make the reference transitive so every host app referencing it generates its own `git.properties` file.
```

TimHess (commented):
Nice addition and refactor, generally LGTM, just one nit

TimHess (approved):

PR #1725, opened 2026-08-05
git.properties: stop fallback-file refresh from overriding ProjectReference opt-out

bart-vmware:
## Description

`WriteGitPropertiesFallbackFile` forced `GenerateGitProperties=true` as a global MSBuild property, which propagates to every `ProjectReference` by default and silently overrode a referenced project's own opt-out (e.g. a shared ServiceDefaults project with `GenerateGitProperties=false`), generating a fallback file it explicitly didn't want. Removing the forced property lets each project's own `auto`/`true`/`false` decision stand for itself and for anything referencing it.

Also documents and tests the `Directory.Build.targets` wrapper needed to safely refresh `git.properties` across a whole solution, since calling the target directly fails with MSB4057 on any project that doesn't reference this package at all.

## Quality checklist

### DIFF
--- a/src/Management/src/GitProperties.Build/buildTransitive/Steeltoe.Management.GitProperties.Build.targets
+++ b/src/Management/src/GitProperties.Build/buildTransitive/Steeltoe.Management.GitProperties.Build.targets
@@ -72,7 +72,7 @@
 
     <PropertyGroup Condition="'$(_GitPropertiesSuppressGitRepositoryNotFound)' == ''">
       <_GitPropertiesSuppressGitRepositoryNotFound>false</_GitPropertiesSuppressGitRepositoryNotFound>
-      <_GitPropertiesSuppressGitRepositoryNotFound Condition="Exists('$(GitPropertiesFallbackFile)')">true</_GitPropertiesSuppressGitRepositoryNotFound>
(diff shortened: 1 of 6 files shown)


src/Management/src/GitProperties.Build/buildTransitive/Steeltoe.Management.GitProperties.Build.targets
```diff
@@ -181,7 +181,7 @@
       rule noted on ResolveGitPropertiesPaths above). The nested build reaches the SDK's own injection points instead, without ever reaching CoreCompile.
     -->
     <MSBuild Projects="$(MSBuildProjectFullPath)" Targets="AssignTargetPaths;PrepareForPublish"
-      Properties="Configuration=$(Configuration);GenerateGitProperties=true;GitPropertiesWriteToProjectDirectory=true;_GitPropertiesSuppressGitRepositoryNotFound=false" />
+      Properties="Configuration=$(Configuration);GitPropertiesWriteToProjectDirectory=true;_GitPropertiesSuppressGitRepositoryNotFound=false" />
```
TimHess: Don't we have the same case with `GitPropertiesWriteToProjectDirectory=true`, also propagating as a global property to child project references where it can't be overridden by a project's local setting?

bart-vmware: Yes, thanks for spotting that. Pushed to address this.

TimHess (dismissed):
LGTM

TimHess (approved):

PR #1730, opened 2026-08-12
Add support for Microting.EntityFrameworkCore.MySql EF Core driver

### LINKED ISSUE #1729: Steeltoe.Connectors.EntityFrameworkCore - Add support to Microting.EntityFrameworkCore.MySql
We can't upgrade to EF Core 10 because Pomelo is no longer [maintained](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql/issues/2007), and we have no way to to use another provider (like [Microting.EntityFrameworkCore.MySql](https://www.nuget.org/packages/Microting.EntityFrameworkCore.MySql)) since everything is internal.

Add [Microting.EntityFrameworkCore.MySql](https://www.nuget.org/packages/Microting.EntityFrameworkCore.MySql) to the supported list of MySql providers, or expose publicly how to get a connection string, so we can use any type of provider we would like.

I've instantiated a custom MySqlEntityFrameworkCorePackageResolver with the Microting.EntityFrameworkCore.MySql assembly and called the internal UseMySql using reflection. Which is not ideal.

bart-vmware:
## Description

[Microting.EntityFrameworkCore.MySql](https://www.nuget.org/packages/Microting.EntityFrameworkCore.MySql) is a new (maintained) fork of the abandoned [Pomelo.EntityFrameworkCore.MySql](https://github.com/PomeloFoundation/Pomelo.EntityFrameworkCore.MySql) driver. This PR adds support for using Microting in the EF Core Connector for MySQL.

At the time of writing, the latest Pomelo release targets EF Core 9, while Microting includes EF Core 10 support (and EF Core 11 support is on its way).

~Fixed: Package `Steeltoe.Connectors.EntityFrameworkCore` still referenced the 8.* version of `Microsoft.EntityFrameworkCore.Relational` in its `net10.0` target.~
**UPDATE:** Reverted because this is too much of a breaking change (see conversation below).

Closes #1729.

## Quality checklist

   - Samples PR: https://github.com/SteeltoeOSS/Samples/pull/476
   - Documentation PR: https://github.com/SteeltoeOSS/Documentation/pull/445

### DIFF
--- a/src/Connectors/src/EntityFrameworkCore/MySql/DynamicTypeAccess/MySqlEntityFrameworkCorePackageResolver.cs
+++ b/src/Connectors/src/EntityFrameworkCore/MySql/DynamicTypeAccess/MySqlEntityFrameworkCorePackageResolver.cs
@@ -11,8 +11,15 @@ namespace Steeltoe.Connectors.EntityFrameworkCore.MySql.DynamicTypeAccess;
 /// </summary>
 internal sealed class MySqlEntityFrameworkCorePackageResolver : PackageResolver
 {
+    private const string MicrotingPackageName = "Microting.EntityFrameworkCore.MySql ";
     private const string PomeloPackageName = "Pomelo.EntityFrameworkCore.MySql";
     private const string OraclePackageName = "MySql.EntityFrameworkCore";
+
+    private static readonly List<string> MicrotingAssemblyNames =
+    [
+        "Microting.EntityFrameworkCore.MySql",
+        "MySqlConnector"
+    ];
 
     private static readonly List<string> PomeloAssemblyNames =
     [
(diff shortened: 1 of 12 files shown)


src/Connectors/src/EntityFrameworkCore/MySql/DynamicTypeAccess/MySqlEntityFrameworkCorePackageResolver.cs
```diff
@@ -11,9 +11,16 @@ namespace Steeltoe.Connectors.EntityFrameworkCore.MySql.DynamicTypeAccess;
 /// </summary>
 internal sealed class MySqlEntityFrameworkCorePackageResolver : PackageResolver
 {
+    private const string MicrotingPackageName = "Microting.EntityFrameworkCore.MySql ";
```
TimHess:
```suggestion
    private const string MicrotingPackageName = "Microting.EntityFrameworkCore.MySql";
```

TimHess (changes requested):

TimHess:
> Fixed: Package `Steeltoe.Connectors.EntityFrameworkCore` still referenced the 8.* version of `Microsoft.EntityFrameworkCore.Relational` in its `net10.0` target.

It is standard to align Runtime and EF Core versions, so this change shouldn't be controversial, but I do wonder if it could cause some difficulty for Steeltoe consumers that aren't as strict about that alignment? 

I'm a little hesitant to potentially force a major version bump for a significant dependency inside a minor Steeltoe release.

bart-vmware:
> > Fixed: Package `Steeltoe.Connectors.EntityFrameworkCore` still referenced the 8.* version of `Microsoft.EntityFrameworkCore.Relational` in its `net10.0` target.
> 
> It is standard to align Runtime and EF Core versions, so this change shouldn't be controversial, but I do wonder if it could cause some difficulty for Steeltoe consumers that aren't as strict about that alignment?
> 
> I'm a little hesitant to potentially force a major version bump for a significant dependency inside a minor Steeltoe release.

I suppose your concern is justified. An app explicitly referencing EF Core 10 with Steeltoe `net10.0` today works _accidentally_ because Steeltoe doesn't call any EF Core APIs with binary-breaking changes between EF Core 8 and 10. As long as Steeltoe stays that way, we should be safe, but it's a slippery slope.

If Steeltoe updated to EF Core 10 on `net10.0`, all consumers are forced to as well. This means they can no longer use Pomelo 9.x with EF Core 9 on .NET 10. Aside from that, there are functional breaking changes between EF Core 8, 9 and 10, that we would be forcing upon users.

I'll revert the break.

TimHess (approved):
LGTM

PR #1744, opened 2026-09-15
Update memory dump dependencies

bart-vmware:
## Description

This PR adds skipped test coverage for verifying that the intertwined dependencies needed for taking memory and thread dumps are compatible. But running these tests takes several minutes, which is why they are skipped by default. The tests compile and run a console app that takes various dumps on all supported target frameworks, while referencing packaged Steeltoe libraries from an emphemeral NuGet feed. This verifies real end-to-end usage that mirrors a Steeltoe release pushed to a package feed.

> [!CAUTION]
> Each time dump-related Steeltoe dependencies are bumped, these tests should be run manually to verify compatibility.

This PR also updates dump-related dependencies to the latest compatible versions, so Steeltoe users benefit from minor downstream fixes accross the board.

### Why do we need such tests?
`Steeltoe.Management.Endpoint` references the following NuGet packages:
1. `Microsoft.Diagnostics.NETCore.Client` (used to obtain heap dumps)
2. `Microsoft.Diagnostics.Tracing.TraceEvent` (used to obtain stack traces)
   - Embeds a private copy of `Microsoft.Diagnostics.FastSerialization.dll` **
   - References package `Microsoft.Diagnostics.NETCore.Client` from above
3. `dotnet-gcdump` (embedded, because global tools can't be referenced)
   - Embeds a private copy of `Microsoft.Diagnostics.FastSerialization.dll` **

** It's fine when `dotnet-gcdump` is compiled against a different version of `Microsoft.Diagnostics.FastSerialization.dll`, as long as it is binary compatible. If not, a `TypeLoadException` or `MissingMethodException` is thrown at runtime (so when taking the dump).

## Quality checklist


src/Management/test/Endpoint.Test/Actuators/DumpPackagingTest.cs
```diff
@@ -0,0 +1,540 @@
...
+            string sourcePath = Path.Combine(libraryDirectory, "DumpProvider.cs");
+            await File.WriteAllTextAsync(sourcePath, LibrarySource);
+        }
+
+        private static async Task WriteAppProjectAsync(string appDirectory, string testAppName, string testLibraryName, string targetFramework,
+            string appSource)
+        {
+            string projectFilePath = Path.Combine(appDirectory, $"{testAppName}.csproj");
+            string projectPath = GetAppProjectFile(testLibraryName, targetFramework);
+            await File.WriteAllTextAsync(projectFilePath, projectPath);
```
TimHess, line 389:
```suggestion
            string projectFileContents = GetAppProjectFile(testLibraryName, targetFramework);
            await File.WriteAllTextAsync(projectFilePath, projectFileContents);
```

src/Management/test/Endpoint.Test/Actuators/DumpPackagingTest.cs
```diff
@@ -0,0 +1,540 @@
...
+
+    // To reproduce test failures, here's a combination that compiles fine, while crashing at runtime due to binary breaking changes:
+    // - dotnet-gcdump v9.0.621003 (embeds Microsoft.Diagnostics.FastSerialization v3.1.16.0)
+    // - Microsoft.Diagnostics.Tracing.TraceEvent v3.2.6 (embeds Microsoft.Diagnostics.FastSerialization v3.2.6.0)
+
+    [Theory(Skip = SkipReason)]
+    [MemberData(nameof(TestTargetFrameworks))]
+    public async Task Can_take_gcdump_from_packaged_Steeltoe_library(string targetFramework)
+    {
+        Assert.SkipWhen(!IsHighestTestHostFramework(), "Running this test only on the latest .NET SDK is sufficient.");
```
TimHess, line 35:
```suggestion
        Assert.SkipWhen(!IsHighestTestHostFramework(),
            $"This test only needs to run once, under the highest target framework ({ResolveTestTargetFrameworks().Max(...)}). " +
            $"Switch your run configuration's target framework if this is unexpectedly skipped.");
```
Not a huge deal, but perhaps we could use a more dynamic and precise message, for cases when the IDE auto-selects an unexpected framework? For example, I ran the tests in Rider and it selected .net8 for me the first time and it took a couple minutes to identify exactly what went wrong

bart-vmware: Updated the message to include the expected framework version.

src/Management/src/Endpoint/Actuators/HeapDump/HeapDumper.cs
```diff
@@ -100,8 +109,19 @@ private void CreateGCDump(int processId, string outputPath, string dumpDescripti
             var heapInfo = new DotNetHeapInfo();
             var memoryGraph = new MemoryGraph(50_000);
 
+            ResetEventPipeDotNetHeapDumperStaticState();
```
TimHess: Does `DumpFromEventPipe` fully join/await its reader task and session-stop task before returning true/false? The comment on `ConcurrentTextWriter` a few lines down says those tasks "can be active at the same time during session shutdown," which made me wonder whether one could still be unwinding after `DumpFromEventPipe` returns. If so, could a straggler thread from a finished dump write `dumpComplete = true` after the next dump's `Reset` call, right as its wait loop starts polling?

src/Management/src/Endpoint/ProcessDumpLock.cs
```diff
@@ -0,0 +1,30 @@
...
+
+namespace Steeltoe.Management.Endpoint;
+
+internal static class ProcessDumpLock
+{
+    private static readonly SemaphoreSlim Semaphore = new(1, 1);
+
+    public static async Task<IDisposable> EnterAsync(CancellationToken cancellationToken)
+    {
+        await Semaphore.WaitAsync(cancellationToken).ConfigureAwait(false);
```
TimHess: Would it be worth logging when a caller is waiting on the lock vs. actively dumping? Right now a second request blocking behind a slow dump looks identical to a hang from the outside.

TimHess (commented):
The testing addition is nice

src/Management/src/Endpoint/Actuators/HeapDump/HeapDumper.cs
```diff
@@ -131,31 +165,29 @@ private static void CreateHeapDump(HeapDumpType? heapDumpType, int processId, st
...
         }
 
-        logStream.Seek(0, SeekOrigin.Begin);
-        using var logReader = new StreamReader(logStream);
-        string logOutput = logReader.ReadToEnd();
+        string logOutput = logWriter.ToString();
 
         if (error != null || !succeeded)
         {
-            cancellationToken.ThrowIfCancellationRequested();
```
TimHess, line 158: Is this removal (also in `EventPipeThreadDumper.CaptureLogOutputAsync`) intentional?

src/Management/src/Endpoint/Actuators/ThreadDump/EventPipeThreadDumper.cs
```diff
@@ -83,55 +83,34 @@ public async Task<IList<ThreadInfo>> DumpThreadsAsync(CancellationToken cancella
...
                 : "Failed to create a thread dump.";
 
-            throw new InvalidOperationException(message, error);
+            throw new InvalidOperationException(message, exception);
         }
 
         if (isTraceLogEnabled)
         {
-            LogDumpLogCaptured(System.Environment.NewLine, logOutput);
+            LogDumpLogCaptured(System.Environment.NewLine, logWriter);
```
TimHess, line 110: Should this call `.ToString()` before passing to the logger method, like `HeapDumper.LogDumpLogCaptured` does?

src/Management/src/Endpoint/Middleware/EndpointMiddleware.cs
```diff
@@ -162,4 +171,7 @@ protected virtual async Task WriteResponseAsync(TResponse response, HttpContext
 
     [LoggerMessage(Level = LogLevel.Trace, Message = "CanInvoke returned false for {Method} request at path {Path}.")]
     private partial void LogInvokeDenied(string method, string? path);
+
+    [LoggerMessage(Level = LogLevel.Debug, Message = "Endpoint at path {Path} is busy handling another request.")]
```
TimHess, line 175:
```suggestion
    [LoggerMessage(Level = LogLevel.Information, Message = "Endpoint at path {Path} is busy handling another request.")]
```
Not a _necessary_ change, but I think it makes sense for this log entry to show up under the default configuration

src/Management/src/Endpoint/Actuators/ThreadDump/EventPipeThreadDumper.cs
bart-vmware: No, that triggers a warning that `ToString()` is potentially expensive. The logger automatically calls it for any object (otherwise tests would fail). I'll push a commit to align both methods.

src/Management/src/Endpoint/Actuators/HeapDump/HeapDumper.cs
bart-vmware: Restored.

bart-vmware: It moved to callers.

TimHess (approved):
LGTM

